PR #2, opened 2020-03-12
change base to nvidia-cuda

Cookt2:


subdavis:
@AlmightyYakob @mattdawkins Is this something that needs to be merged?

AlmightyYakob:
I still need to test this, apologies.

subdavis:
Oh, this is the thing you need to test.  I can help with that, LMK.

docker/viame_girder_worker_gpu.Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM kitware/viame:girder-worker-gpu-latest
```
AlmightyYakob, line 1: @Cookt2 Is this supposed to be built from the other included Dockerfile, `docker/girder_worker_gpu.docker`? This works correctly if I instead specify that image here, but leaving this as is and pulling down the image doesn't contain everything needed.

subdavis: @AlmightyYakob I imagine the published image just hasn't been updated since this is a PR and there's also no CI.

AlmightyYakob:
So it looks like this is mostly good, but I had to add some changes. I can make a new branch on this repo with this new stuff, so that we can change/update it without issue.

subdavis:
Closing in favor of #70

PR #3, opened 2020-03-16
Replace hardcoded pipeline list with api call

AlmightyYakob:
Updates the ui to support exposed pipelines list on the backend


client/src/views/Home.vue
```diff
@@ -50,13 +48,18 @@ export default {
...
     this.location_ = getLocationFromRoute(to);
     next();
   },
   methods: {
-    ...mapMutations(["setLocation"]),
+    ...mapMutations(["setLocation", "setPipelines"]),
+    async fetchPipelines() {
+      const { data } = await this.girderRest.get("viame/pipelines");
+      this.setPipelines(data);
+    },
```
subdavis: If pipelines are global state, I think it would make sense for a `fetchPipelines` action to be in the store.

client/src/views/Home.vue
```diff
@@ -250,7 +253,9 @@ export default {
                     v-on="on"
                     text
                     small
-                    :disabled="selectedEligibleClips.length < 1"
+                    :disabled="
+                      selectedEligibleClips.length < 1 || !pipelines.length
+                    "
```
subdavis: if this line is so long it has to wrap, probably best to refactor as a computed property.

AlmightyYakob:
> Probably best to move new api calls into the store. I recognize that the current existing code doesn't do this, but I think the existing style is wrong and we ought not follow it.
> 
> LMK what you think.

Agree with both of your suggestions, I'll change it.

AlmightyYakob:
@subdavis I commited your suggestions, as well as updated the `girder` module import to be more consistent, as we're importing it several places now. Lmk what you think

PR #18, opened 2020-03-27
Allow for annotating image-sequences when annotations not already present

AlmightyYakob:
This addresses the major concern raised by @mattdawkins. The diff is fairly large, as there needed to be several changes to get this to work, including fixing usability.

### DIFF
--- a/client/src/views/Viewer.vue
+++ b/client/src/views/Viewer.vue
@@ -93,9 +93,9 @@ export default {
         return null;
       }
 
-      const items = this.items
+      return this.items
         .filter(item => {
-          var name = item.name.toLowerCase();
+          let name = item.name.toLowerCase();
           return (
             name.endsWith("png") ||
             name.endsWith("jpeg") ||
@@ -105,15 +105,6 @@ export default {
         .map(item => {
           return `api/v1/item/${item._id}/download`;
         });
-
-      if (!items.length) {
-        this.$snackbar({
-          text: "No images found",
-          timeout: 4500
-        });
-      }
-
-      return items;
     },
     frameRate() {
       if (!this.dataset) {
@@ -394,6 +385,14 @@ export default {
     }
   },
   watch: {
+    imageUrls(val) {
+      if (!val.length) {
+        this.$snackbar({
+          text: "No images found",
+          timeout: 4500
+        });
+      }
+    },
     detections() {
       this.updatecheckedTracksAndTypes();
     }
--- a/client/src/views/Viewer.vue
+++ b/client/src/views/Viewer.vue
@@ -406,13 +406,24 @@ export default {
       this.dataset = dataset || null;
     },
     async loadDetections() {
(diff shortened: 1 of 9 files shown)


subdavis:
This will mega-conflict with #17 

@BryonLewis was first, so the git gods command ye to rebase (after he merges).  :(

AlmightyYakob:
This should be good regarding conflicts with #17. Since I've rebased off that branch, this PR should wait until that one is merged.

client/src/views/Viewer.vue
```diff
@@ -411,18 +423,27 @@ export default {
...
-        return Object.freeze(detection);
-      });
+      let detections = [];
+      try {
+        const { data } = await this.girderRest.get("viame_detection", {
+          params: { folderId: this.dataset._id }
+        });
+
+        detections = data
+          ? data.map(detection => Object.freeze(detection))
```
subdavis: I know it was never commented before, but using object.freeze to protect objects from reactivity like this really deserves some explanation.  Some developer will come along later and wonder what was going on.

server/viame_server/event.py
```diff
@@ -0,0 +1,29 @@
...
+        Item().save(item)
+
+        folder = Folder().findOne({"_id": item["folderId"]})
+
+        # FPS is hardcoded for now
+        folder["meta"].update({"type": "image-sequence", "viame": True, "fps": 30})
+        Folder().save(folder)
+
+
+def mark_folder_for_annotation(event):
```
subdavis: This function runs for all uploads, but only has an effect for folders.  You could rename this function `maybe_mark_folder_for_annotation(event)` to indicate that it doesn't always do that.

I think this function is straightforward, so you don't have to, just wanted to point out a pattern I've seen elsewhere in the girder ecosystem (particularly in worker)

client/src/views/Viewer.vue
AlmightyYakob: I'm not sure why it's being done either, it was there before I came onto the project. I think it's deserving of it's own PR though.

AlmightyYakob:
@subdavis I think I addressed your concerns. I also added a small change in 
f421a26 to match the client's [list](https://github.com/VIAME/VIAME-Web/blob/435fd702e322b7aefa0763b3b31464474449002e/client/src/views/Viewer.vue#L97-L99) of valid image formats, tested locally no problem.

client/src/views/Viewer.vue
```diff
@@ -422,6 +422,11 @@ export default {
     /* END TODO */
     async loadDataset(datasetId) {
       var { data: dataset } = await this.girderRest.get(`folder/${datasetId}`);
+
+      if (!dataset) {
+        throw "Could not fetch dataset!";
```
subdavis: I've actually never seen anyone do this.  I found an old stackoverflow thread that says this behaves unexpectedly in IE and firefox, but that was in 2012.  My gut says `throw new Error()` should always be used, but maybe you know something I don't.

PR #25, opened 2020-04-03
[refactor] annotator structure reorganization

subdavis:
* moves layers to subfolder
* moves controls to subfolder
* moves annotators to subfolder
* extracts common annotator logic to mixin
* documents annotation app design choices
* documents annotator app API
* Adds `yarn.lock` because somehow that was never committed. 
* Removes some chainWebpack config that 1) didn't make sense and 2) seemed to be breaking my build.  We really need to document when we do non-standard things....
* Uses a proper release version of GWC otherwise the dependency will change every time master updates.

This is just a first-pass reorganization.  `Viewer.vue` will need more consideration.  These were just some of the most obvious areas for improvement.


client/vue.config.js
```diff
@@ -12,14 +12,6 @@ module.exports = {
...
   chainWebpack: config => {
     config.resolve.symlinks(false);
-    config.module
-      .rule("additional")
-      .test(/\.m?jsx?$/)
-      .include.add(/@girder\/components/)
-      .add(/vue-utilities/)
-      .end()
-      .use()
-      .loader("babel-loader");
```
subdavis, line 22: This broke my build.  I have no idea what it was here for.  I definitely don't know why JSX is being referenced.

subdavis:
Screenshot from 2020-04-03 16-08-16

This is what happens, btw.

client/src/components/annotators/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+## state
+
+The state available on the injected Vue annotator instance.
+
+```js
+viewer // geojs viewer
+playing // boolean
+frame // number
+maxFrame // number
```
BryonLewis, line 13: viewer - is there a better name we should be calling this object?  imageViewer, frameViewer, imageRenderer, frameRenderer or something like that.  Looking at the code my mind immediately goes 'there's a Viewer.vue, must have some relationship to that' and that turns out to be wrong.  Could be better to rename the Viewer.vue to something more fitting.

subdavis: It's an instance of `geo.map()` so something like `geoViewer` or `geoMap` I think would be appropriate.

client/src/components/annotators/VideoAnnotator.vue
```diff
@@ -65,38 +33,10 @@ export default {
...
-      ];
-      interactorOpts.zoomAnimation = {
-        enabled: false
-      };
-      interactorOpts.momentum = {
-        enabled: false
-      };
-      interactorOpts.wheelScaleY = 0.2;
-      this.viewer.interactor().options(interactorOpts);
+      this.baseInit();
```
AlmightyYakob, line 39: Can you add a comment here referencing the mixin?

client/src/components/annotators/annotator.js
```diff
@@ -0,0 +1,95 @@
...
+    this.provided.$on("prev-frame", this.prevFrame);
+    this.provided.$on("next-frame", this.nextFrame);
+    this.provided.$on("play", this.play);
+    this.provided.$on("pause", this.pause);
+    this.provided.$on("seek", this.seek);
+    this.emitFrame();
+    this.emitFrame = throttle(this.emitFrame, 200);
+  },
+  methods: {
+    baseInit() {
```
AlmightyYakob, line 46: Might be good to add a comment to this function as well explaining what it's doing, if possible.

subdavis: this is explained in `src/components/annotators/README.md`

PR #29, opened 2020-04-06
[housekeeping] major refactor of Viewer using composition api

subdavis:
This is an exploration of how to resolve some of the complexity in `Viewer.vue` with `@vue/composition-api`.  I'm actually really happy with how this is feeling.  It relegates individual features to files and builds an explicit dependency graph of which feature requires which other feature.

This would mean that state continues to exist in `Viewer.vue` rather than in Vuex, but I'm actually OK with this right now because none of the state currently being tracked is needed by other routes, so vuex may not be the best choice.

This pattern would make building other viewers with different features straightforward, as well as multi-view apps where there is more than 1 video window.

Obviously this code doesn't work, it's mostly an example of how such a thing could be done.  I'd endorse writing these composition functions in TypeScript as a place to start with the TS rewrite :)

cc @BryonLewis and @AlmightyYakob


BryonLewis:
I like separating out the the computed functions into some groups.  Although I question if we should have some top level groups just feeding processed data to lower level components.  Is there a reason we shouldn't feed the raw data and have the child component process it?
Can we instead of having AnnotationStyle and TextStyle functions in the viewer, pass the needed data (selectedTrackId, filteredDetections, and colormapping) to the AnnotationLayer/TextLayer and have them calculated it instead.
If you look at the AnnotationLayer it takes the `AnnotationData `which is a format that is only used internally and used to create the `frameMap`, in the process it references this.data.polygon, which is created in a computed property of the viewer.  Is there a benefit in which everything that consumes root data (`filteredDetections`) has the same format of input data.  I think a similar thing could be done for the eventViewerData  and the lineChartData as well.  So each component within Viewer starts with the same formatted data ('filteredDetections' ) and performs their own manipulation. I don't know if this is of benefit for readability between components and maybe the current format is necessary for performance reasons.

subdavis:
Thanks for taking the time to look at this and respond -- those are good points.  The following are just some gut feelings from having written similar apps, and not really recommendations.  

>  Is there a reason we shouldn't feed the raw data and have the child component process it? Can we instead of having AnnotationStyle and TextStyle functions in the viewer, pass the needed data (selectedTrackId, filteredDetections, and colormapping) to the AnnotationLayer/TextLayer and have them calculated it instead.

We definitely could.  The drawback I see with this is it takes complexity from `Viewer` and multiplies it across similarly complex components.  It also expands the scope of those layers by forcing them to accept and deal with concepts they are currently isolated from by Viewer.  `AnnotationLayer` currently has no concept of `selection`, `track`, or even what a `detection` is.  It only deals with polygons and frames.  I think there's real value in separating those concepts and isolating them into their own modules.

> So each component within Viewer starts with the same formatted data ('filteredDetections' ) and performs their own manipulation.

I'm trying to look at this refactor as a way to improve modularity and make all the parts more testable and flexible.  I'd paraphrase your suggestion as "What if we just make all the layers aware of all the state they _might_ need, and let them use it how they want" (let me know if this isn't accurate).  That makes the testing burden higher by multiplying the possible input conditions.  Down the line, refactoring changes to the `filteredDetections` API surface become more expensive as it touches more files.  Each of the data manipulation functions, like `markerData` and `textData` and `annotationData` protect their respective layers from unnecessary data inside `filteredDetections` as well as from state like `confidence` which to me is valuable.

> current format is necessary for performance reasons.

I don't think there would be performance differences.

subdavis:
Had a meeting with @BryonLewis and @AlmightyYakob, we decided that this was a good direction to start with and I'm going to get a minimal working example with just playback working.

If we need to refactor composition functions as vuex modules later, that should be much less painful.  After my MVP, we will try and come up a with way to split up the rest of the refactor.  We will likely work off of some kind of staging branch and merge to master once the refactor has feature parity with master.

client/src/components/controls/LineChart.vue
```diff
@@ -85,95 +84,92 @@ export default {
...
-        .attr("d", d => line(d.values))
-        .style("stroke", d => (d.color ? d.color : "#4c9ac2"))
-        .on("mouseenter", function(d) {
-          var [x, y] = d3.mouse(this);
+        .append('path')
+        .attr('class', 'line')
+        .attr('d', (d) => line(d.values))
+        .style('stroke', (d) => (d.color ? d.color : '#4c9ac2'))
+        .on('mouseenter', (d) => {
+          const [_x, _y] = d3.mouse(this);
```
BryonLewis, line 144: Using the arrow function here rebinds `this` to the Vue Component instead of the current function.  Causing some errors to be logged because it is trying to find the position using a VueComponent as and argument for `d3.mouse(this)` instead of an svg element:
image
We could change it back to utilizing a `function (d) {` or instead utilize `d3.event.currentTarget` although that has some warnings with Babel and Webpack.  It has the possibility of changing the d3.event during the event (source: [d3.event](https://github.com/d3/d3-selection#event))

client/src/views/Viewer/Viewer.vue
```diff
@@ -0,0 +1,488 @@
...
+              <v-list-item-title>
+                Add feature points, starting with tail (t key)
+              </v-list-item-title>
+            </v-list-item>
+            <v-list-item @click="deleteDetection">
+              <v-list-item-title>
+                Delete both feauture points for current frame (q key)
+              </v-list-item-title>
+            </v-list-item>
+          </v-list>
```
BryonLewis, line 479: feauture -> feature

The last `@click`, should it be `deleteFeaturePoints` instead of `deleteDetection` ?

client/src/views/Viewer/Viewer.vue
```diff
@@ -0,0 +1,488 @@
...
+          v-if="imageUrls.length || videoUrl"
+          ref="playbackComponent"
+          v-mousetrap="[
+            { bind: 'g', handler: () => toggleFeaturePointing('head') },
+            { bind: 'h', handler: () => toggleFeaturePointing('head') },
+            { bind: 't', handler: () => toggleFeaturePointing('tail') },
+            { bind: 'y', handler: () => toggleFeaturePointing('tail') },
+            { bind: 'f', handler: () => nextFrame() },
+            { bind: 'd', handler: () => prevFrame() },
+            { bind: 'q', handler: deleteDetection }
```
BryonLewis, line 374: `deleteDetection` should it be `deleteFeaturePoints` ?

client/src/components/NavigationBar.vue
```diff
@@ -42,34 +42,56 @@ export default {
             this.runningJobIds.splice(this.runningJobIds.indexOf(jobId), 1);
           }
           break;
+        default:
```
BryonLewis, line 45: I know this isn't related to changes that you made, but can we insert a comment after
`case jobStatus.SUCCESS.value:`
`// fall through`
Just so we indicate that this is an intentional fall through in the switch statement.
Might be a discussion if we should turn on the linting rule for this.

client/src/components/NavigationTitle.vue
```diff
@@ -1,17 +1,20 @@
...
 };
 </script>
 
 <template>
   <v-toolbar-title class="my-title mx-5">
-    <v-tooltip open-delay="1000" bottom>
+    <v-tooltip
+      open-delay="1000"
+      bottom
+    >
```
BryonLewis, line 17: 1000ms seems a bit excessive for a tool tip unless it is expected that it isn't used most of the time.  It is only a version number.

Additionally have we considered linking the Logo to the current root of the site.  Most apps utilize the logo as a de facto home button.

client/src/components/controls/LineChart.vue
```diff
@@ -85,95 +84,92 @@ export default {
...
 
-      var axis = d3.axisRight(y).tickSize(width);
+      const axis = d3.axisRight(y).tickSize(width);
       this.axis = axis;
       svg
-        .append("g")
-        .attr("class", "axis-y")
+        .append('g')
+        .attr('class', 'axis-y')
         .call(axis)
```
BryonLewis, line 129: `this.axis` axis within the current filed is only referenced a few times and I don't think `this.axis` is referenced at all.  Not entirely sure what it is being used for?
Timeline.vue has a `this.axis` assignment in the mounted function, but it's` this.axis` is used across multiple methods.

client/src/components/controls/Timeline.vue
```diff
@@ -60,52 +60,51 @@ export default {
...
-    var axis = d3
+    const axis = d3
       .axisTop()
       .scale(scale)
       .tickSize(height - 30)
       .tickSizeOuter(0);
     this.axis = axis;
     this.g = d3
       .select(this.$refs.workarea)
-      .append("svg")
```
BryonLewis, line 92: `this.axis` - private non-reactive variable.  Should it be initialized in `created()` instead of first being introduced here?  I understand keeping it here because there isn't a really good initialization value other than `null`, just wondering opinions on that.

client/src/components/layers/EditAnnotationLayer.vue
```diff
@@ -96,33 +96,33 @@ export default {
...
         }
       } else if (this.editing) {
         this.changed = true;
-        if (typeof this.editing !== "string") {
+        if (typeof this.editing !== 'string') {
           throw new Error(
             `editing props needs to be a string of value ${geo
               .listAnnotations()
-              .join(", ")} when geojson prop is not set`
+              .join(', ')} when geojson prop is not set`,
```
BryonLewis, line 108: Is this more readable?
image

client/src/use/useEditingLayer.js
```diff
@@ -0,0 +1,108 @@
...
+  setTrackEditMode,
+  deleteDetection,
+  setDetection,
+}) {
+  async function deleteTrack({ trackId }) {
+    const _detections = detections.value;
+    const result = await prompt({
+      title: 'Confirm',
+      text: `Please confirm to delete track ${trackId}`,
+      confirm: true,
```
BryonLewis, line 21: Adding a catch to either return or throw an error when an object is passed that doesn't contain trackId.  So if trackId is null.  Right now it will still ask for confirmation, iterate through the detections and then exit.

client/src/use/useEventChart.js
```diff
@@ -0,0 +1,30 @@
+
+import { computed } from '@vue/composition-api';
+import { groupBy, minBy, maxBy } from 'lodash';
+
+export default function useEventChart({ filteredDetections, selectedTrackId, typeColorMap }) {
+  const eventChartData = computed(() => {
+    if (!filteredDetections.value.length === 0) {
+      return [];
```
BryonLewis, line 8: `if (!filteredDetections.value.length === 0)`  is this doing something I'm not understanding, if the length is zero it is still false and I believe it will fail if filteredDetections is not an array..

client/src/use/useTextLayer.js
```diff
@@ -0,0 +1,52 @@
...
+  typeColorMap,
+}) {
+  // TODO: what's the proper way to consume vuetify in a composition function?
+  const vuetify = inject('vuetify');
+
+  const textData = computed(() => {
+    if (!filteredDetections.value.length) {
+      return null;
+    }
+    const data = [];
```
BryonLewis, line 15: other use{Layers} will do a check for `(filteredDetections.value.length === 0)` I think we may want to make them the same.

- useEventChart.js
- useMarkerLayer.js
We may want to synchronize the return values for these.  Some return an empty array `[]` if the detection length is 0, others will return null instead.

subdavis:
I've resolved the bugs I'm aware of, and this now has the same functionality (or better) of master.

Once @BryonLewis is ready to merge #57, we can take one last look and I'll be ready to merge.  Thanks, everyone!

AlmightyYakob:
Thanks for all of your work on this @subdavis and @BryonLewis! This will make the rest of development much easier.

client/src/components/controls/Timeline.vue
subdavis: Out of scope of this PR maybe.  Let's go through and clean up individual files in the future.

PR #42, opened 2020-04-09
Add user guide button

subdavis:
Screenshot from 2020-04-09 10-39-53

Screenshot from 2020-04-09 10-45-07

Also adds linting fixes to README.md.  No changes to README appearance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,46 +1,47 @@
-
 <img src="http://www.viametoolkit.org/wp-content/uploads/2016/08/viami_logo.png" alt="VIAME Logo" width="200" height="78">
 <br>
-VIAME-Web is a web interface for performing data management, video annotation, and running the algorithms stored within
-the VIAME (https://github.com/VIAME/VIAME) repository. When compiled, docker instances for VIAME-Web can be run either as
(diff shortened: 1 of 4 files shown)


BryonLewis:
> Screenshot from 2020-04-09 10-39-53
> 
I know this is a default btn but is this styling used anywhere else?  It does make it more prominent but it's another consideration for other buttons in the Navbar, like logout (I think that uses the 'text' attribute).

I don't want to block the adding of a user guide button (it's definitely needed) but at some point we should re-evaluate what is going on here.  We have a User Guide button with an icon, followed by a filtering tool for displaying items,  followed by a Save icon which has no tool tip to indicate exactly what it is.  

Just some concepts:
image

image

Second image icons would obviously have hover state stating what they should do inside of the annotation.  The spacing on the actual toolbar for annotation stuff I didn't mess with, that could be changed.  Open the second image to larger, there is a clear vertical divider between Annotation Stuff, and system level stuff.

subdavis:
I was going for prominence.  We can let the design team make recommendations later.  I like how your concepts work, but I don't know if having it blend in like that is an issue.

PR #63, opened 2020-04-17
Fix video annotation/detection

AlmightyYakob:
This PR does two main things:

1. Updates the event listener to add the annotate metadata when uploading a video, as well as an image
2. Changes the metadata on folders/items, to work in a more sensible way. 

Now, the metadata structure is as follows:

* Folders have a metadata structure of:
```
annotate: true | false
fps: Int
type: "video" | "image-sequence"
viame: true
```

* Result detection items have a metadata structure of:
```
detection: <folderId>
```

The value of the detection field represents the folder to which the detections belong. Ideally, I think to keep track of videos more appropriately, we should find a way to set a "video" field on the folder metadata, with the item ID of the video for that folder. However, that's not possible with our current event listeners, as the event we receive is triggered before the item is created.

### DIFF
--- a/client/src/use/useGirderDataset.js
+++ b/client/src/use/useGirderDataset.js
@@ -4,6 +4,7 @@ import {
 
 const ImageSequenceType = 'image-sequence';
 const VideoType = 'video';
+const defaultFrameRate = 30;
 
 export default function useGirderDataset() {
   const girderRest = inject('girderRest');
@@ -12,7 +13,7 @@ export default function useGirderDataset() {
   const imageUrls = ref([]);
   const videoUrl = ref(null);
 
-  const frameRate = computed(() => dataset.value && dataset.value.meta.fps);
+  const frameRate = computed(() => (dataset.value && dataset.value.meta.fps) || defaultFrameRate);
 
   const annotatorType = computed(() => {
     if (!dataset.value) {
--- a/client/src/views/Home.vue
+++ b/client/src/views/Home.vue
@@ -170,8 +170,7 @@ export default {
           this.girderRest.put(
             `/item/${csvFile.itemId}/metadata?allowNull=true`,
             {
-              folderId: folder._id,
-              pipeline: null,
+              detection: folder._id,
             },
           );
         });
--- a/server/viame_server/__init__.py
+++ b/server/viame_server/__init__.py
@@ -53,4 +53,6 @@ class GirderPlugin(plugin.GirderPlugin):
             "check_annotations",
             check_existing_annotations,
         )
-        events.bind("model.upload.finalize", "fileUpload", maybe_mark_folder_for_annotation)
+        events.bind(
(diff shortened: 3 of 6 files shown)


subdavis:
The only part I don't understand is why detection items (The CSV file, I'm assuming?) need a reference back to their parent folder.

AlmightyYakob:
> The only part I don't understand is why detection items (The CSV file, I'm assuming?) need a reference back to their parent folder.

It's not necessary for the item itself to have this reference (the folder could have it), but in general it's used to associate the csv file with the video folder, as a detection. As far as why I did it this specific way, it was just a smaller change than having that live in the folder instead. 

I'd actually say it makes more sense to have `detection` live on the folder, as I think it conveys the semantics better, however we'd have to make sure we don't lose functionality. Currently, if there's several csv files with this `detection` field set to the folderId, the most recently created one will be used. We could replicate this on the folder itself by just changing the value of `detection` every time a save was done (and a new csv file was uploaded). If we did it this way, any time a new save is done, the old csv files wouldn't have a link to the folder. I'm not sure if this is a problem or not, just laying out the scenario. Curious to hear what you think @subdavis.

AlmightyYakob:
> 1. If you upload a video without a subfolder, you still don't get an annotate button.

This is because currently, when creating a folder, we're using the a girder web component, and the `viame` metadata field is not set. So when uploading the video, it doesn't mark the folder for annotation, as the `viame` field is not set to `true`. I'm not sure if we're able to pass metadata to the call being made in the GWC.

> 2. This caused the annotations I already had to no longer get discovered.  I think some kind of migration will be necessary, to check if annotations exist in the old format and add whatever new metadata is required for them to get picked up.  What do you think?

Yes, I agree. I think merging should be held off until I have a working script in place to do this.

BryonLewis:
> This is because currently, when creating a folder, we're using the a girder web component, and the `viame` metadata field is not set. So when uploading the video, it doesn't mark the folder for annotation, as the `viame` field is not set to `true`. I'm not sure if we're able to pass metadata to the call being made in the GWC.

Not entirely sure, but is this allowing you to set metadata on upload?
image
(in Upload.vue?)

AlmightyYakob:
> Not entirely sure, but is this allowing you to set metadata on upload?
> image
> (in Upload.vue?)

That allows you to set metadata on the folder created when also uploading a video, but the scenario I'm referring to is just using the `Create Folder` button to create a folder, and then uploading a video into that, without creating a folder (so in your example, `createFolder = false`).

BryonLewis:
understood

AlmightyYakob:
I've got a working script that will make the necessary migrations. If everything else is approved, I can merge this, take the server down, run the migration, pull in master with these changes, and start it up again.

AlmightyYakob:
Data migration went successfully :+1:

Issue #38, opened 2020-04-08
Fix running the pipelines


aashish24:

subdavis:
I'm stuck here.  I don't know where the docker images for kwiver came from, but it doesn't appear to work.  Moreover, the separation of concerns here isn't good, and it's causing a lot of confusion.

Right now, I'm working off the branch https://github.com/VIAME/VIAME-Web/pull/77, which has dockerfiles written by someone in Vision.  These do not appear to work.

In order to get the pipelines working, as well as clearly divide responsibilities, I propose the following:

1. Vision will provide us with a docker image for running kwiver pipelines.  Along with this image, Vision will provide example data known to produce valid results and an example `docker run` command explaining how to execute a pipeline using that example data and the image.  This image will not container girder worker.
1. D&A will take that base image and create our own derived girder worker image for integration with VIAME web.

This will clearly separate kwiver concerns from girder worker concerns, and follow best practices like verifying that all composite parts work before attempting integration, rather than doing everything at once.

@aashish24 and @mattdawkins please let me know what you think.  

If this makes sense, this issue will be considered blocked until the base kwiver image is available.

mattdawkins:
I don't care who does what, as long as we come to a short-term resolution.

The issue with the current container is I can't /bin/bash into it due to issues with celery and other non-kwiver software to test kwiver. The current container, as far as I know, is auto-built on Proteus via user click (not nightly) and builds this docker file:

https://github.com/VIAME/VIAME/blob/master/docker/viame_girder_worker_gpu.docker

If you want me to modify this to be based off an nvidia ubuntu image, it builds kwiver inside of it, and contains nothing else that's fine and is trivial.

subdavis:
Yes, that would be ideal, I think.  

The simplest ubuntu Nvidia base image that builds kwiver and installs pipelines to a known directory.

We can handle the addition of celery, worker, and everything else.

Thanks.

subdavis:
Update:

* Got a basic kwiver image from the official VIAME repo: https://hub.docker.com/r/kitware/viame
* Built girder worker into that base image, integrated into VIAME web.

#77 has been updated with my latest changes.  Currently run pipelines **is working** on that branch, but the output file is not named `annotations.csv`, so you can't view the results in the web UI.  This is the last thing to be done.

Thanks everyone for making these resources availabl.e

PR #70, opened 2020-04-21
Add Nvidia Support

AlmightyYakob:


AlmightyYakob:
After discussion with @subdavis, we're gonna wait to merge this until #38 is addressed, as otherwise we really have no way of testing this.

docker/viame_girder_worker_gpu.Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+WORKDIR /home
+
+RUN cd /viame/cmake \
+	&& chmod +x build_server_ubuntu.sh \
+        && sleep 1 \
+ 	&& ./build_server_ubuntu.sh
+
+RUN git clone https://github.com/VIAME/VIAME-Web.git /viame-web \
+	&& mv /viame-web/docker/provision provision \
+	&& mv /viame-web/server viame_girder
```
subdavis, line 14: This is definitely not how this should work.

docker/viame_girder_worker_gpu.Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM kitware/viame:girder-worker-gpu-latest
+
+USER root
+
+WORKDIR /home
+
+RUN cd /viame/cmake \
+	&& chmod +x build_server_ubuntu.sh \
+        && sleep 1 \
+ 	&& ./build_server_ubuntu.sh
```
subdavis, line 10: I don't see why this is being done in this dockerfile rather than in the base image.  It has no dependencies.

docker/girder_worker_gpu.docker
```diff
@@ -0,0 +1,70 @@
...
+COPY --from=build /girder_worker/docker-entrypoint.sh /docker-entrypoint.sh
+RUN pip3 install /*.tar.gz
+
+RUN useradd -D --shell=/bin/bash && useradd -m worker
+
+RUN chown -R worker:worker /usr/local/lib/python3.6/dist-packages/girder_worker
+
+USER worker
+
+RUN girder-worker-config set celery broker "amqp://%(RABBITMQ_USER)s:%(RABBITMQ_PASS)s@%(RABBITMQ_HOST)s/"
```
subdavis, line 63: Likewise, this should definitely not happen here.  You never bake runtime details into an image like this.

docker/viame_girder_worker_gpu.Dockerfile
AlmightyYakob: I agree, should this go in `girder.Dockerfile`?

subdavis: I was referring to the fact that this clones master rather than doing a COPY.

subdavis:
Closing in favor of #77

PR #74, opened 2020-04-24
Fixes for subfolders

subdavis:
There were a fair number of bugs I found while testing your branch. 

* Uploading images and video at the same time silently omits either video or images depending on what appears first in the list.  Now, it throws an error
* Ability to choose to create subfolders removed.  Now, it's totally deterministic and only happens when multiple videos are uploaded.  Taking this choice away from the user I think is good.
* Errors used to be duplicated since there is only 1 instance of `errorMessage`.  I've dealt with this by moving error reporting out of the `v-for` and having it tied exclusively to the first item in the pendingUpload list (see below)
Screenshot from 2020-04-24 09-35-18
* Errors during upload could get you stuck in an infinite loop (the while loop) since nothing gets dequeueud when an error is thrown.  This was fixed so that any upload error causes the whole upload process to end, and the user has to resolve the error by aborting the pendingUpload before restarting.  Now, errors can only be thrown by `pendingUploads[0]` because only the first item in the list will ever be actively running.  
* Data Browser didnt' refresh when errors occur, so if an upload partially fails, you couldn't see what parts succeeded.  This was fixed.

### DIFF
--- a/client/src/components/Upload.vue
+++ b/client/src/components/Upload.vue
@@ -41,7 +41,7 @@ function prepareFiles(files) {
       csv: csvFiles,
     };
   }
-  throw new Error('Function should not be called on empty array');
+  throw new Error('No supported data types found.  Please choose video or image frames.');
 }
 
 function entryToFile(entry) {
--- a/client/src/components/Upload.vue
+++ b/client/src/components/Upload.vue
@@ -106,7 +106,6 @@ export default {
     preUploadErrorMessage: null,
     pendingUploads: [],
     defaultFPS: '10', // requires string for the input item
-    subfolderDialog: false, // used to warn user about creating sub folders
   }),
   computed: {
     uploadEnabled() {
@@ -114,11 +113,6 @@ export default {
     },
   },
   methods: {
-    subfolderWarning(event) {
-      if (event) {
-        this.subfolderDialog = true;
-      }
-    },
     // Filter to show how many images are left to upload
     filesNotUploaded(item) {
       return item.files.filter(
@@ -375,7 +369,6 @@ export default {
                   hint="Enabled when many videos are being uploaded"
                   persistent-hint
                   class="pl-2"
-                  @change="subfolderWarning($event)"
                 />
               </v-col>
               <v-col>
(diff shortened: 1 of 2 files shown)


subdavis:
FWIW, I think this file has gotten too complex in its abuse of GWC's uploader.

Each "pendingUpload" should probably have its own component state and its own instance of the `fileUploader` mixin, I think.  But that's a problem for another day.

client/src/components/Upload.vue
```diff
@@ -16,16 +16,30 @@ function prepareFiles(files) {
...
+  const imageFiles = files.filter(imageFilter);
+  const csvFiles = files.filter(csvFilter);
+
+  if (videoFiles.length > 0 && imageFiles.length > 0) {
+    throw new Error('Do not upload images and videos in the same batch.');
+  } else if (csvFiles.length > 1) {
+    throw new Error('Can only upload a single CSV Annotation per import');
+  } else if (videoFiles.length > 1 && csvFiles.length > 0) {
+    throw new Error('Annotation upload is not supported when multiple videos are uploaded');
+  } else if (videoFiles.length) {
```
BryonLewis, line 29: Fails silently without user feedback if you try to upload a single CSV file.

client/src/components/Upload.vue
```diff
@@ -16,16 +16,30 @@ function prepareFiles(files) {
...
+      type: 'image-sequence',
+      media: imageFiles,
+      csv: csvFiles,
+    };
   }
-  return [
-    'image-sequence',
-    files.filter((file) => imageFilter(file) || csvFilter(file)),
-  ];
+  throw new Error('Function should not be called on empty array');
```
BryonLewis, line 42: Since this is a user facing error should it be more descriptive to the end user.  Something more like : "Upload should contain images or video files."?

client/src/components/Upload.vue
```diff
@@ -341,10 +363,12 @@ export default {
             <v-row>
               <v-col cols="auto">
                 <v-checkbox
-                  v-model="pendingUpload.createSubFolders"
+                  :input-value="pendingUpload.createSubFolders"
                   label="Create Subfolders"
+                  disabled
+                  hint="Enabled when many videos are being uploaded"
+                  persistent-hint
```
BryonLewis, line 370: Should we even display a unchecked disabled user input to the user? (If they are uploading an image sequence)  I know that reading it lets you know that you can upload multiple video files at once if you happen to be uploading image sequences.  Either way is fine, just a question.

subdavis: I sorta like it.

client/src/components/Upload.vue
subdavis: This should not be a user facing error.  the method should not be called with an empty array, and there is currently no code path for doing so.

JS doesn't have typed exceptions but if it did, this would be an unhandled type.

client/src/components/Upload.vue
subdavis: Fixed, ty

client/src/components/Upload.vue
BryonLewis: We don't filter the upload dialog to match the types we have listed (*.jpg, *.mp4 *.csv) and others.  If you upload a file or multiples that aren't classified as valid image or video types I believe the user would get this error.  While the array isn't empty, once it is filtered `videoFiles.length` `imageFiles.length` and `csvFiles.length` could all be zero.

image

subdavis: Fixed in b2848312cc949f1d9bcd5045953c7a8b1904fd46

Issue #43, opened 2020-04-09
Build VIAME with all of the pipelines for docker image


aashish24:

subdavis:
done in #77

PR #77, opened 2020-04-29
Bugfix/pipeline

### LINKED ISSUE #38: Fix running the pipelines

subdavis:
fixes #38 #43


server/viame_server/transforms.py
```diff
@@ -58,9 +58,9 @@ def transform(self):
 
         return self.folder_path
 
-    # def cleanup(self):
-    #     shutil.rmtree(os.path.dirname(self.folder_path),
-    #                   ignore_errors=True)
+    def cleanup(self):
+        shutil.rmtree(os.path.dirname(self.folder_path),
+                      ignore_errors=True)
```
AlmightyYakob, line 63: Was this just causing errors prior?

docker/docker-compose.yml
```diff
@@ -33,24 +35,21 @@ services:
...
+      - BROKER_CONNECTION_TIMEOUT=2
 
   girder_worker:
+    runtime: 'nvidia'
+    build:
+      context: ../
+      dockerfile: docker/girder_worker.Dockerfile
     volumes:
-      - ./server:/home/viame_girder
       - pipelines:/home/VIAME/build/install/configs/pipelines
```
AlmightyYakob, line 47: Do we have to coordinate this with the change in `girder_worker.Dockerfile`? As now these files should be at `/opt/noaa/viame/configs/pipelines/` in the girder worker image, right?

server/viame_tasks/tasks.py
```diff
@@ -44,34 +48,50 @@ def run_pipeline(self, path, pipeline):
...
+            "cd /opt/noaa/viame &&",
             ". ./setup_viame.sh &&",
-            "pipeline_runner",
-            "-p /home/VIAME/configs/pipelines/{}".format(pipeline),
+            "kwiver",
+            "runner",
+            "-p /opt/noaa/viame/configs/pipelines/{}".format(pipeline),
             "-s input:video_filename={}".format(image_list_file),
             "-s detector_writer:file_name={}".format(output_path),
         ]
```
AlmightyYakob, line 58: No direct action needed, but for this and the command above, it might be useful to utilize `pathlib`, as these paths are being shared, and it would make this command easier to read.

docker/docker-compose.yml
```diff
@@ -1,9 +1,11 @@
 ## Use this to start the girder/ girder-worker ecosystem
-version: "3.2"
+version: "2.4"
```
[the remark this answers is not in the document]
subdavis: See `/README.md`

docker/docker-compose.yml
```diff
@@ -1,9 +1,11 @@
...
 services:
 
   rabbit:
     image: rabbitmq:management
+    ports:
+      - "15672:15672"
```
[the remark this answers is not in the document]
subdavis: Allows browser access to the rabbit management console.  I find it handy for knowing the queue state.

docker/girder.Dockerfile
```diff
@@ -9,32 +9,19 @@ COPY client/ /app/
...
 COPY --from=builder /app/dist/ /usr/local/share/girder/static/viame/
 
-ENTRYPOINT ["/home/provision/girder_entrypoint.sh"]
+WORKDIR /home/viame_girder
+
+COPY docker/provision /home/provision
+COPY server/setup.py /home/viame_girder/
+RUN pip install --no-cache-dir .
+COPY server/ /home/viame_girder/
+RUN pip install --no-deps .
```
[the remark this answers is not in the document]
subdavis: This is considered a docker python best practice, and it's the same reason you copy `package.json` and `yarn.lock` and `yarn install` before you bring the rest of the code over.

Makes use of docker caching so that you aren't reinstalling PIP dependencies when all you changed was local code.

docker/provision/girder_entrypoint.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 python3 /home/provision/init_girder.py
-girder serve --host 0.0.0.0
+exec /sbin/tini -v -- girder serve --host 0.0.0.0
```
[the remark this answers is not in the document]
subdavis: There was a discussion about this in #girder recently.

You can check out https://github.com/krallin/tini -- the gist is that there are special responsibilities that being PID1 has, and regulare processes don't usually fulfill those responsibilities. However, `docker run --init` will inject tini, or you can include tini in your container.  

Docker Compose v2.4 doesn't include `tini` init support, so I added it to the image.

server/viame_server/transforms.py
subdavis: I think the previous author wanted to log into the container to look at the leftovers, so they commented this out.  I uncommented because any created file in the worker needs to be removed afterward.

server/viame_tasks/tasks.py
subdavis: Valid.  I'll do this.

docker/docker-compose.yml
```diff
@@ -33,24 +35,21 @@ services:
       - GIRDER_MONGO_URI=mongodb://mongo:27017/girder
       - GIRDER_ADMIN_USER=${GIRDER_ADMIN_USER}
       - GIRDER_ADMIN_PASS=${GIRDER_ADMIN_PASS}
+      - CELERY_BROKER_URL=amqp://guest:guest@rabbit/
+      - BROKER_CONNECTION_TIMEOUT=2
 
   girder_worker:
+    runtime: 'nvidia'
```
AlmightyYakob, line 42: Also, what's requried to run this on a machine without an nvidia GPU? Is just commenting this out sufficient?

subdavis: I don't know.  I sort of doubt it will work at all.

AlmightyYakob: It does seem to work (at least, I can run tasks no issue, gpu specific pipelines will probably fail). Maybe an override file like we have for `dev` would be good for people (like myself) trying to run this locally on non-gpu-enabled machines.

AlmightyYakob: If you want I can add this, since I've already got something locally

mattdawkins: We have CPU and GPU binaries of VIAME which have slightly different pipelines enabled, some pipelines will fail some wont in machines without GPUs if you install the GPU version on them

subdavis:
@BryonLewis Jobs have been added back into girder worker

docker/docker-compose.yml
subdavis: This is `./server`, a bind-mount of the source code.  I don't think the source code should be mounted in the production docker-compose.  It is, however, added in the dev compose.

I don't think this has anything to do with pipelines -- could you make sure that's true?

server/viame_tasks/tasks.py
subdavis: Does 27a6e60  resolve your comment?

docker/docker-compose.yml
AlmightyYakob: These lines https://github.com/VIAME/VIAME-Web/blob/001031fe99b5799ab4da7a441c1f4a65c778cd19/docker/docker-compose.yml#L24 https://github.com/VIAME/VIAME-Web/blob/001031fe99b5799ab4da7a441c1f4a65c778cd19/docker/docker-compose.yml#L45

This is what allows girder to expose the list of runnable pipelines through an API endpoint. I just tested this locally and it doesn't expose them properly, as it's trying to look at `/home/VIAME/build/install/configs/pipelines`. It seems like the place it should be pointing at is `/opt/noaa/viame/configs/pipelines/`, however I don't know if there's a difference there between all pipelines, and ones that are enabled from the install. There was before (which is why I pulled from the `install` sub-directory).  It doesn't seem like there's an `install` folder in that new path. I think the image this is using has them all enabled, which I think means that the list of files at `/opt/noaa/viame/configs/pipelines/` should be good. @mattdawkins could you weigh in?

server/viame_tasks/tasks.py
AlmightyYakob: Yep LGTM

docker/docker-compose.yml
subdavis: You're right, I'm dumb.

server/viame_tasks/tasks.py
```diff
@@ -6,72 +7,118 @@
...
+        "-show_streams",
+        file_name,
+    ]
+    process = Popen(command, stderr=PIPE, stdout=PIPE)
+    stdout = process.communicate()[0]
+    jsoninfo = json.loads(stdout.decode('utf-8'))
+    videostream = list(filter(lambda x: x["codec_type"] == "video", jsoninfo["streams"]))
+    if len(videostream) != 1:
+        raise Exception('Expected 1 video stream, found {}'.format(len(videostream)))
+    self.girder_client.addMetadataToFolder(folderId, videostream[0])
```
BryonLewis, line 121: Was the intention of this just for reference or was it supposed to override the fps data that is specified during upload?
If it should be added for reference should it be contained in it's own JSON object to prevent it from polluting the root of the metadata?
image

docker/docker-compose.yml
AlmightyYakob: 43ef72f seems to fix this issue.

docker/docker-compose.yml
```diff
@@ -21,36 +21,36 @@ services:
...
-      - pipelines:/home/VIAME/build/install/configs/pipelines
+      # Exposes pipeline directory to girder so it can see what pipelines are available.
+      - pipelines:${VIAME_PIPELINES_PATH}:ro
     depends_on:
       - rabbit
     environment:
-      - GIRDER_WORKER_BROKER=amqp://guest:guest@rabbit/
-      - GIRDER_WORKER_BACKEND=amqp://guest:guest@rabbit/
-      - GIRDER_WORKER_PLUGINS_ENABLED=girder_io
+      - VIAME_PIPELINES_PATH=${VIAME_PIPELINES_PATH}
```
AlmightyYakob, line 51: Small thing, but I don't think this needs to be here, as it's only used by girder and not girder worker.

subdavis: See https://github.com/VIAME/VIAME-Web/pull/77/commits/43ef72f6e22d7e3c807c7be1d400e064c087b700#diff-46d897de59fd6accaec892ee86d50f15L12

server/viame_tasks/tasks.py
subdavis: It was sort of a "we're going to all the trouble to compute this info, might as well hang onto it".

Also the FPS you get isn't usable because real FPS fractions can be nasty.

I think nesting it into it's own json object is a great suggestion.

docker/docker-compose.yml
AlmightyYakob: Oops, my bad.

server/viame_tasks/tasks.py
subdavis: Done in c0bbd61

Issue #96, opened 2020-05-08
Migrate to JSON internal schema.


subdavis:
It's wasteful and slow to convert CSV to JSON every time the annotation page is loaded.  We probably shouldn't read bytes from disk at all on the server, and instead let the client directly request a girder file.

IMO, the internal representation of tracks should be some JSON schema.  It would be simple enough to do a copy-on-read the first time the client requests a VIAME csv, then maintain the JSON data for annotation, and add export endpints to convert back to CSV if the user wants it.

This would also allow us to support other necessary annotation data types, like `kw18` and `kpf`, which are needed for general adoption within kitware.

AlmightyYakob:
When it comes to training, there's currently no supported JSON format for annotation in the training step of viame (it [supports](https://github.com/VIAME/VIAME/tree/master/examples/detection_file_conversions#integrated-detection-formats) csv, kw18, and HABCAM). So we'd definitely need to have the conversion in place for training.

AlmightyYakob:
I should add though I'm definitely in favor of us doing this.

subdavis:
I fully expect that any interaction with external systems would involve conversion in and out of our internal schema.

mattdawkins:
Training now supports json, the main issue is it needs to auto-detect json vs csv as csv is still in production (and probably always will be as we have some contracts which require it for excel users doing analytics on the output).

No objection to changing what is used internally as long as we always have optional support for csv at input (initial upload) and output (download through GUI) time as well.

subdavis:
Semi-related:

Ive been thinking a lot about how detections and tracks are represented in the client, and I think we may want to migrate to a track-centric data structure.

* representing tracks as a global detection array cannot ever support interpolation
* a flat detection array is efficient O(1) for frame lookup, but bad for almost everything else.
* For example, update/insert is always O(N) (where N is detection count)
* This O(N) cost is multiplied in practice because there are several different "views" of the data needed that all get computed on update: like the unfiltered track list for the left pane, the filtered detection list for rendering, etc.
* Right now the lookup cost for figuring out a track's start and end frames is also O(N) which is why  gotoTrackFirstFrame is such a pain:

```
function gotoTrackFirstFrame({ trackId }) {
      setTrackEditMode(trackId, false);
      const _frame = eventChartData.value.find((d) => d.track === trackId)
        .range[0];
      seek(_frame);
    }
```

[...]

PR #97, opened 2020-05-08
91 pyproject black isort

### LINKED ISSUE #91: Create pyproject.toml with python linting rules

subdavis:
Adds `black` and `isort`

The VS Code config to run both of these on save is:

```
{
  "python.formatting.provider": "black",
  "[python]": {
    "editor.formatOnSave": true,
    "editor.codeActionsOnSave": {
      "source.organizeImports": true
    }
  },
}
```

I also factored the parsing logic out into a serializers subfolder in preparation of #96 and other formats.  I don't think it's great style to do that all in the REST handler anyway.

FWIW, I think 80 char columns is too restrictive.  I prefer 90-100 myself, but ya'll should decide.

resolves #91

### DIFF
--- a/server/pyproject.toml
+++ b/server/pyproject.toml
@@ -1,0 +1,12 @@
+[tool.black]
+line-length = 80
+skip-string-normalization = true
+target-version = ['py37']
+exclude='\.eggs|\.git|\.mypy_cache|\.tox|\.venv|_build|buck-out|build|dist|venv'
+
+[tool.isort]
+line_length = 80
+use_parentheses = true
+include_trailing_comma = true
+multi_line_output = 3
+skip = "venv/**/*"
--- a/server/pyproject.toml
+++ b/server/pyproject.toml
@@ -1,11 +1,11 @@
 [tool.black]
-line-length = 80
+line-length = 88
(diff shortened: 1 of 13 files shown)


server/pyproject.toml
```diff
@@ -0,0 +1,12 @@
+[tool.black]
+line-length = 80
```
AlmightyYakob, line 2: I agree with your reservations, we should probably make this 88 (black default) or higher. 80 is kind of short.

server/pyproject.toml
```diff
@@ -0,0 +1,12 @@
...
+skip-string-normalization = true
+target-version = ['py37']
+exclude='\.eggs|\.git|\.mypy_cache|\.tox|\.venv|_build|buck-out|build|dist|venv'
+
+[tool.isort]
+line_length = 80
```
AlmightyYakob, line 8: Same here

subdavis:
Changed to 88 columns.

>  I'm not sure why to include the unrelated code change here

I was just trying to lump several code-quality changes into 1 PR.  None of these changes do any behavior changes or bug-fixes.

PR #115, opened 2020-05-13
107 enable download of independent parts

### LINKED ISSUE #107: API should allow downloading of images without an annotation file
For example the raw folder in https://viame.kitware.com/#/folder/5eb374f38d01d968cc1667f8 has images but no metadata/annotations by design, but you should still be able to download it.

subdavis:
dialog

This was surprisingly nontrivial.

* Includes a fair amount of refactoring in Home.vue because if I don't do stuff like that during feature work, I won't do it at all.  I wish we had more time for general code quality work, but taxing features / bug fixes with code quality is doable, IMO.
* Still an issue right now where, if a folder doesn't have detections, you get an unfriendly error message.

fixes #107

### DIFF
--- a/client/src/common/api.js
+++ b/client/src/common/api.js
@@ -8,7 +8,7 @@ function deleteResources(resources) {
       folder: resources
         .filter((resource) => resource._modelType === 'folder')
         .map((resource) => resource._id),
-      item: this.selected
+      item: resources
         .filter((resource) => resource._modelType === 'item')
         .map((resource) => resource._id),
     }),
--- a/client/src/common/api.js
+++ b/client/src/common/api.js
@@ -1,0 +1,48 @@
+import girderRest from '@/girder';
+
+function deleteResources(resources) {
+  const formData = new FormData();
+  formData.set(
+    'resources',
+    JSON.stringify({
+      folder: resources
+        .filter((resource) => resource._modelType === 'folder')
+        .map((resource) => resource._id),
+      item: this.selected
+        .filter((resource) => resource._modelType === 'item')
+        .map((resource) => resource._id),
+    }),
+  );
+  return girderRest.post('resource', formData, {
+    headers: { 'X-HTTP-Method-Override': 'DELETE' },
+  });
+}
+
+function getExportUrl(id, type) {
+  return `${girderRest.apiRoot}/viame_detection/${id}/export?type=${type}`;
+}
+
+function runConversion(itemId) {
+  return girderRest.post(`/viame/conversion?itemId=${itemId}`);
+}
+
+function runPipeline(itemId, pipeline) {
+  return girderRest.post(
+    `/viame/pipeline?folderId=${itemId}&pipeline=${pipeline}`,
+  );
+}
+
+function setMetadataForItem(itemId, metadata) {
+  return girderRest.put(
+    `/item/${itemId}/metadata?allowNull=true`,
+    metadata,
+  );
+}
+
+export {
+  deleteResources,
+  runConversion,
+  runPipeline,
+  getExportUrl,
+  setMetadataForItem,
+};
--- a/client/src/common/viame.service.js
+++ b/client/src/common/viame.service.js
@@ -1,5 +1,21 @@
 import girderRest from '@/girder';
 
+function makeViameFolder({
+  folderId, name, fps, type,
+}) {
+  return girderRest.post(
+    '/folder',
+    `metadata=${JSON.stringify({
+      fps,
+      type,
+      viame: true,
+    })}`,
+    {
+      params: { parentId: folderId, name },
+    },
+  );
+}
+
 function deleteResources(resources) {
   const formData = new FormData();
   formData.set(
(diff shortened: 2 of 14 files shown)


client/src/common/api.js
```diff
@@ -0,0 +1,48 @@
...
+
+function deleteResources(resources) {
+  const formData = new FormData();
+  formData.set(
+    'resources',
+    JSON.stringify({
+      folder: resources
+        .filter((resource) => resource._modelType === 'folder')
+        .map((resource) => resource._id),
+      item: this.selected
```
BryonLewis, line 11:
```suggestion
      item: resources
```

client/src/components/Export.vue
```diff
@@ -0,0 +1,125 @@
...
+              depressed
+              block
+              target="_blank"
+              :href="getExportUrl(folderId, 'media')"
+            >
+              {{ mediaType }}
+            </v-btn>
+          </v-card-actions>
+          <v-card-text class="pb-0">
+            Get latest detections csv only
```
BryonLewis, line 94: I'm assuming eventually we want to be more generic than only csv files.  Should we just use the term 'detections' generically?

client/src/components/Export.vue
```diff
@@ -0,0 +1,125 @@
...
+import { getExportUrl } from '@/common/api';
+import {
+  ImageSequenceType,
+  VideoType,
+} from '@/constants';
+
+const MediaTypeMap = {
+  [ImageSequenceType]: 'image sequence',
+  [VideoType]: 'video',
+};
```
BryonLewis, line 11: should we make the display names for the types universal within the program so if they are used more than once we don't have conflicting terms?

client/src/components/Upload.vue
```diff
@@ -4,16 +4,11 @@ import {
...
-
-// Utility for commonly used regular expressions
-const videoFilesRegEx = new RegExp(`${videoFileTypes.join('$|')}$`, 'i');
-const imageFilesRegEx = new RegExp(`${imageFileTypes.join('$|')}$`, 'i');
-const webFriendlyImageRegEx = new RegExp(`${webFriendlyImageFileTypes.join('$|')}$`, 'i');
+import {
+  videoFilesRegEx,
+  imageFilesRegEx,
+  webFriendlyImageRegEx,
+} from '@/constants';
```
BryonLewis, line 11: I like the idea of having these in a central location for access by multiple components.  Within this component we directly through a string define types as `video` and `image-sequence` should they by in the constants as well and imported in?

client/src/constants.js
```diff
@@ -0,0 +1,26 @@
...
+
+// Supported File Types
+const videoFileTypes = ['.mp4', '.avi', '.mov', '.mpg'];
+const webFriendlyImageFileTypes = ['.jpg', '.jpeg', '.png'];
+const imageFileTypes = [...webFriendlyImageFileTypes, '.bmp', '.sgi', '.tif', '.tiff', '.pgm'];
+
+// Utility for commonly used regular expressions
+const videoFilesRegEx = new RegExp(`${videoFileTypes.join('$|')}$`, 'i');
+const imageFilesRegEx = new RegExp(`${imageFileTypes.join('$|')}$`, 'i');
+const webFriendlyImageRegEx = new RegExp(`${webFriendlyImageFileTypes.join('$|')}$`, 'i');
```
BryonLewis, line 12: Feel free to disregard for now if I'm looking too far forward.:
Should we consolidate this more into a object like `MediaTypes` which contains the type name, a display name, supported file types.  One source of the supported media types and allows to add stuff like "Satellite Imagery" for stuff like geoTiff and others in the future.

subdavis:
I think I'll want to change the "export" endpoint to more of a "get export options" endpoint that returns URLs for each export rather than doing the actual export, then calling that options endpoint when the dialog is opened.  Otherwise, no way to know if the thing has detentions or not.

BryonLewis:
Small additions:

-  you can download an empty zip file if you attempt to export an empty folder
-  If you manually delete the aux annotations and then try to download, it pops open a new tab with a REST error.  I don't know if I would classify this as an error other than the fact that the menu still has the option to download them.  This occurs even if you delete the entire auxillary folder, or if you upload a sequence of images without a csv.
`{
    "message": "Detections not found in folder 5eac5e3d5b588f84005c2900",
    "type": "rest"
}`

client/src/constants.js
subdavis: Out of scope for this PR, IMO.

subdavis:
* Reverts a change from #110 because the `viame` metadata key was actually needed
* Splits API code into 2 services based on resource
* Moves transcode testing into the postUpload handler, corrects logic
* removal of some dead code for processing pipelines
* export endpoint now returns resource URLs rather than invoking them.

Here's how it looks when you try to dowload a video that doesn't yet have detections:

Screenshot from 2020-05-15 14-31-43

client/src/common/viame.service.js
```diff
@@ -0,0 +1,77 @@
...
+    null,
+    {
+      params: { folder },
+    },
+  );
+}
+
+function runPipeline(itemId, pipeline) {
+  return girderRest.post(
+    `/viame/pipeline?folderId=${itemId}&pipeline=${pipeline}`,
```
subdavis, line 59: should probably use the mode verbose syntax to let axios handle urlencoding here.

client/src/views/Home.vue
```diff
@@ -136,14 +129,20 @@ export default {
...
+      const transcodes = uploads.filter(({ results, folder }) => {
+        const videoTranscodes = results
+          .filter(({ name }) => videoFilesRegEx.test(name))
+          .map(({ itemId }) => runVideoConversion(itemId));
+        const imageTranscodes = results
+          .filter(({ name }) => !webFriendlyImageRegEx.test(name) && imageFilesRegEx.test(name));
+
+        if (imageTranscodes) {
+          runImageConversion(folder._id);
+        }
```
BryonLewis, line 142: Do you want the length of `imageTranscodes` not the actual result because wouldn't that be an empty array?  This is causing an image transcode task to be run on the EHU data.

subdavis: Thanks, fixed.

PR #116, opened 2020-05-15
Add generate_ pattern to allowed pipelines

### LINKED ISSUE #112: Add generate_ pipes alongside detector_ and tracker_
There weren't any in the last release, though the next one will also have a generate_empty_frame_lbls.pipe which needs to be included

AlmightyYakob:
Fixes #112.

Can't really be tested until the update mentioned in that issue adds those pipelines. @mattdawkins whenever that happens please let me know.

### DIFF
--- a/client/src/components/RunPipelineMenu.vue
+++ b/client/src/components/RunPipelineMenu.vue
@@ -50,6 +50,15 @@ export default {
         },
       });
     },
+    pipeTypeDisplay(pipeType) {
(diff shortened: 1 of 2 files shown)


AlmightyYakob:
As discussed with @mattdawkins, the menu for this in the front-end should be something like "Utility" instead of "Generate".

AlmightyYakob:
Tested running the new pipeline locally and it worked without issue.

PR #118, opened 2020-05-15
111 pipeline menu

### LINKED ISSUE #111: Run pipelines dialogue should have different dropdowns for different types
It's going to get crowded fairly fast, maybe have 3 sub dialogues, detectors, trackers, trained pipelines? I'm sure there's other ways to partition them.

At a minimum they should be listed in alphabetical order as a short-term solution

subdavis:
fixes #111

based on #115, will need to rebase

### DIFF
--- a/client/src/common/viame.service.js
+++ b/client/src/common/viame.service.js
@@ -54,6 +54,10 @@ function runImageConversion(folder) {
   );
 }
 
+function getPipelineList() {
+  return girderRest.get('viame/pipelines');
+}
+
 function runPipeline(itemId, pipeline) {
   return girderRest.post(
     `/viame/pipeline?folderId=${itemId}&pipeline=${pipeline}`,
@@ -69,6 +73,7 @@ function setMetadataForItem(itemId, metadata) {
 
 export {
   deleteResources,
+  getPipelineList,
   makeViameFolder,
   runImageConversion,
   runVideoConversion,
--- a/client/src/components/RunPipelineMenu.vue
+++ b/client/src/components/RunPipelineMenu.vue
@@ -1,6 +1,6 @@
 <script>
 import { runPipeline } from '@/common/viame.service';
-import { mapState, mapActions } from 'vuex';
+import { mapActions } from 'vuex';
 
 export default {
   props: {
@@ -11,19 +11,23 @@ export default {
   },
 
   computed: {
-    ...mapState(['pipelines']),
-
     selectedEligibleClips() {
       return this.selected.filter(
         ({ _modelType, meta }) => _modelType === 'folder' && meta && meta.annotate,
       );
     },
-    pipelinesRunnable() {
-      return this.selectedEligibleClips.length < 1 || !this.pipelines.length;
+    pipelinesNotRunnable() {
+      return this.selectedEligibleClips.length < 1 || !this.pipelines;
     },
   },
 
-  mounted() {
+  asyncComputed: {
+    pipelines() {
+      return this.fetchPipelines();
+    },
+  },
+
+  created() {
     this.fetchPipelines();
   },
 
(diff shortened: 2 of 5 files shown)


server/viame_server/__init__.py
```diff
@@ -32,11 +32,21 @@ def load_pipelines():
...
+    pipedict = {
+        'detector': {'pipes': [], 'description': '',},
+        'tracker': {'pipes': [], 'description': '',},
+    }
+    for pipe in pipelist:
+        nameparts = pipe.replace('.pipe', '').split('_')
+        name = ' '.join(nameparts[1:])
+        type = nameparts[0]
+        pipedict[type]['pipes'].append({'name': name, 'type': type, 'pipe': pipe})
+    return pipedict
```
subdavis, line 49: Process and parse the pipelines server-side.  Protect the client code from having to know the particulars of pipe naming conventions.  

For example, if convention switched from `TYPPE_PIPENAME.pipe` for determining pipe type to subfolders and folder name, the client code shouldn't be aware of this change.

client/src/components/RunPipelineMenu.vue
```diff
@@ -0,0 +1,136 @@
...
+
+  asyncComputed: {
+    pipelines() {
+      return this.fetchPipelines();
+    },
+  },
+
+  created() {
+    this.fetchPipelines();
+  },
```
AlmightyYakob, line 32: Here you're calling `this.fetchPipelines()` twice, but you're only relying on the one called in the `pipelines` computed property. IMO you should map state here.

client/src/components/RunPipelineMenu.vue
```diff
@@ -0,0 +1,136 @@
...
+      );
+      this.$snackbar({
+        text: `Started pipeline on ${clips.length} clip${
+          clips.length ? 's' : ''
+        }`,
+        timeout: 6000,
+        immediate: true,
+        button: 'View',
+        callback: () => {
+          this.$router.push('/jobs');
```
AlmightyYakob, line 50: I think it's generally better to use direct component names over paths:

```suggestion
          this.$router.push({ name: 'jobs' });
```

client/src/common/viame.service.js
```diff
@@ -54,6 +54,10 @@ function runImageConversion(folder) {
   );
 }
 
+function getPipelineList() {
+  return girderRest.get('viame/pipelines');
+}
+
```
AlmightyYakob, line 60: Is this used anywhere else but in the store? If not I think it can directly be part of the `fetchPipelines` action, unless you think it'll be used separately from the store (I guess if using the value in the store wouldn't make sense?) in the near future.

server/viame_server/__init__.py
AlmightyYakob, line 49: A few things here in my suggestion:

1. Double quotes to match black style
2. Changing the `type` variable to `pipe_type`, since `type` is a keyword in python
3. Use a shorthand for splitting the `nameparts` array into the first and remaining elements
4. Generalizing this to any "type", since we will probably be adding another once #112 is finalized. Since your approach already grabbed the first part of the pipeline name and used it as the dict key, this just takes full advantage of that.

Lmk what you think


```suggestion
    pipedict = {}
    for pipe in pipelist:
        pipe_type, *nameparts = pipe.replace(".pipe", "").split("_")
        pipe_info = {"name": " ".join(nameparts), "type": pipe_type, "pipe": pipe}

        if pipe_type in pipedict:
            pipedict[pipe_type]["pipes"].append(pipe_info)
        else:
            pipedict[pipe_type] = {"pipes": [pipe_info], "description": ""}

    return pipedict
```

client/src/common/viame.service.js
subdavis: The idea is to isolate `girderRest` into a single place in the code, mostly because our API is unstable and changes often.  I haven't gone through to do it everywhere, but I move it when it touches something I'm working on.

Issue #84, opened 2020-05-06
Represent data internally as tracks


subdavis:
## problem

Performing operations on these tracks throw a console error.  

* Go to first track
* delete

## Ideas

* Solution1: filter the track list to only show tracks above confidence filter
* solution 2: Perform tracklist operatons on all detections (instead of filteredDetections).  If `go to first frame` is selected, adjust the confidence automatically so that track is visible.
* Solution 3: Some visual indicator that the track is below confidence, and disable all interactions on that track list item.

I prefer 3.

Note: #66 may be related.

subdavis:
Here's what I'm thinking:

1. Load detections from the server.  In the future, the server should actually be track-centric as well, consistent with #96 
2. Transform to tracks (Will change when #96 is addressed, no transform will be needed)
3. Filter tracks by confidence  This will be the "master" track list for all uses in the UI.
4. Filter that list by selected types (this list will be used to populate the left pan track list)
5. Filter **that** list by individually selected tracks. (this list will be use to populate the timelines and will be rendered.

* Maintain a "materialized view" of detections populated by the final filtered list, a sparse JS array of arrays.  its indexes are frame numbers, its values are an array of all detections visible on that frame.   ( Only works with method 2 below)

## Interpolation method 1: slow read fast write

1. find out which tracks are active on a frame (Average case O(logN))
2. find the detections on either side of the frame you want: if you want frame 10, you need to find the previous and next keyframes, then interpolate them to get 10.  You can either maintain detections in a dense array where index is meaningless and binary-search to get the detection, or you can maintain a sparse array where index is frame number and `slice` to find the surrounding keyframes.  I have no idea which is faster.
3. Using the detections you found for each track, interpolate to get the current frame.

## Interpolation method 2: fast read slow write

* Basically the same as "no interpolation" except that, any time a keyframe is added, interpolation for keyframe - 1 through keyframe + 1 must be recomputed.  
* This is very efficient for reads, potentially slow for writes, and has a much larger memory footprint because you don't save the space from only keeping track of keyframes.

subdavis:
closed in #137

PR #124, opened 2020-05-19
Server-side JSON track processing

subdavis:
Add parameter for server to return JSON list of tracks.   This will support #84, but it's only the first step for that issue.

* It is my intention that the server would eventually save this data straight to a file and serve it to the app, then serialize to the VIAME csv format on demand.
* In order to use this on the client, serialization of this new structure to VIAME csv is still missing.  I think that should be a part of this PR, but I want to get some other eyes on the structure here first.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,7 +26,6 @@ devel/
 downloads/
 eggs/
 .eggs/
-lib/
 lib64/
 parts/
 sdist/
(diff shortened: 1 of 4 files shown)


aashish24:
@subdavis do we implement the track representation concept in a separate PR, is that right?

subdavis:
Yes.

PR #133, opened 2020-05-21
Bugfix: track attribute swap

subdavis:


subdavis:
> It's saving the annotation properly and loading it, but we should probably update the AttributeInput panel to handle the undefined value property passed to it.

I don't get that error after this update.  Do you?

AlmightyYakob:
@BryonLewis where in the frontend do you get that error? When I run this locally I don't seem to see observe this

BryonLewis:
> I don't get that error after this update. Do you?
Yeah I still do.  Even if `selectedTrack.trackAttributes` or `selectedDetection.detectionAttributes` is `null` it will default to Undefined which isn't supported by AttributeInput.vue.  You need to swap the view on an element which has no attributes bound. 
image

You can add it but then there is an issue in useSelection.js that I'm looking at.

PR #136, opened 2020-05-22
Add dataclasses with general abstraction

### LINKED ISSUE #131: Make minimum python version 3.7 and add dataclasses
A first use case for dataclasses are in support of things introduced by  https://github.com/VIAME/VIAME-Web/pull/124

AlmightyYakob:
Fixes #131 

This is a first pass at adding dataclasses, and also adds a bit of abstraction to the operations we're doing. I think this makes it easier to read.

This also adds a good bit of typing information, which I think adds clarity. The main drive for me adding type annotations during this PR was mainly any code that I touched, however I could add more if it's desired. Also, if we generally like this, it might be worth it to add `mypy` or something similar, so we can do static type checking and take full advantage of type hints.

### DIFF
--- a/server/viame_server/serializers/viame.py
+++ b/server/viame_server/serializers/viame.py
@@ -25,14 +25,14 @@ class Feature:
 class Track:
     begin: int
     end: int
-    key: int
+    trackId: str
     features: List[Feature] = field(default_factory=lambda: [])
     confidencePairs: Dict[str, float] = field(default_factory=lambda: {})
     attributes: Dict[str, Any] = field(default_factory=lambda: {})
 
 
 def row_info(row: List[str]) -> Tuple[int, int, List[float], float]:
-    trackid = int(row[0])
+    trackId = str(row[0])
     frame = int(row[2])
     bounds = [
         float(row[3]),
@@ -42,7 +42,7 @@ def row_info(row: List[str]) -> Tuple[int, int, List[float], float]:
     ]
     fish_length = float(row[8])
 
-    return trackid, frame, bounds, fish_length
+    return trackId, frame, bounds, fish_length
 
 
 def _deduceType(value: str) -> Union[bool, float, str]:
@@ -93,7 +93,7 @@ def _parse_row(row: List[str]) -> Tuple[Dict, Dict, Dict, List]:
 
 def _parse_row_for_tracks(row: List[str]) -> Tuple[Feature, Dict, Dict, List]:
     head_tail_feature, attributes, track_attributes, confidence_pairs = _parse_row(row)
-    trackid, frame, bounds, fishLength = row_info(row)
+    trackId, frame, bounds, fishLength = row_info(row)
 
     feature = Feature(
         frame,
@@ -153,12 +153,12 @@ def load_csv_as_tracks(file):
             track_attributes,
             confidence_pairs,
         ) = _parse_row_for_tracks(row)
-        trackid, frame, _, _ = row_info(row)
+        trackId, frame, _, _ = row_info(row)
 
-        if trackid not in tracks:
-            tracks[trackid] = Track(frame, frame, trackid)
+        if trackId not in tracks:
+            tracks[trackId] = Track(frame, frame, trackId)
 
-        track = tracks[trackid]
+        track = tracks[trackId]
         track.begin = min(frame, track.begin)
         track.features.append(feature)
 
@@ -173,4 +173,4 @@ def load_csv_as_tracks(file):
             for (key, val) in confidence_pairs:
                 track.confidencePairs[key] = val
 
-    return {trackid: asdict(track) for trackid, track in tracks.items()}
+    return {trackId: asdict(track) for trackId, track in tracks.items()}
--- a/server/viame_server/serializers/viame.py
+++ b/server/viame_server/serializers/viame.py
@@ -3,7 +3,7 @@ VIAME Fish format deserializer
 """
 import csv
 import re
-from dataclasses import dataclass, asdict, field
+from dataclasses import dataclass, field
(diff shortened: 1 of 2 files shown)


server/viame_server/serializers/viame.py
```diff
@@ -3,11 +3,49 @@
...
+    tail: Optional[Tuple[str, str]] = None
+    fishLength: Optional[float] = None
+    attributes: Optional[Dict[str, Union[bool, float, str]]] = None
+
+
+@dataclass
+class Track:
+    begin: int
+    end: int
+    key: int
```
subdavis, line 28: Let's keep this as `trackId` for consistency.   

Also, I'd like to make `trackId` a string and cast it to a string in the parser below.  It's number-ness is never used (for indexing, for example) so we might as well support other types of track identifiers, like UUIDs and things.

server/viame_server/serializers/viame.py
```diff
@@ -3,11 +3,49 @@
...
+    features: List[Feature] = field(default_factory=lambda: [])
+    confidencePairs: Dict[str, float] = field(default_factory=lambda: {})
```
subdavis, line 30: I found out while working on the client that it's also nicer to keep confidencePairs as a list of lists:

```python
[
  ['name', 0.86],
  ['v2', 1],
]
```

server/viame_server/serializers/viame.py
```diff
@@ -3,11 +3,49 @@
...
 
-def _deduceType(value):
+@dataclass
+class Feature:
+    """Feature represents a single detection in a track."""
+
+    frame: int = None
+    bounds: List[float] = None
+    head: Optional[Tuple[str, str]] = None
+    tail: Optional[Tuple[str, str]] = None
```
subdavis, line 19: should be `[number, number]` -- it's an (x, y) point

AlmightyYakob: Good catch, thanks.

server/viame_server/serializers/viame.py
AlmightyYakob: Good point, I'm in favor of that.

server/viame_server/serializers/viame.py
AlmightyYakob: I'm in favor of that, however that means we'd have to remove this section of code:
https://github.com/VIAME/VIAME-Web/blob/91b3c83b6d89bfa85e1ebb032777ab9b997efd98/server/viame_server/serializers/viame.py#L171-L174

Is there any downside to doing this? I'm not sure what that comment means.

server/viame_server/serializers/viame.py
AlmightyYakob: Is `int` sufficient? Or would it require `float`?

server/viame_server/serializers/viame.py
```diff
@@ -92,50 +147,24 @@ def load_csv_as_tracks(file):
...
-            track["end"] = frame
-            # final confidence pair should be taken as the
-            # pair that applied to the whole track
-            for (key, val) in confidence_pairs:
-                track["confidencePairs"][key] = val
-
-    return tracks
+            track.attributes[key] = val
+
+    return {trackId: asdict(track) for trackId, track in tracks.items()}
```
subdavis, line 170: This results in a different data object than the code it's replacing, and it's different than the client expects:

```
      {
        "attributes": {},
        "bounds": [
          654,
          743,
          173,
          266
        ],
        "fishLength": null,
        "frame": 2,
        "head": null,
        "tail": null
      },
```

Before, where a particular feature type is missing, the key will simply be omitted.  `feature.head` is never null, it's either a number or undefined.

subdavis: Also, `asdict` makes serialization slower as the entire dict has to be constructed from scratch: https://stackoverflow.com/questions/52229521/why-is-dataclasses-asdictobj-10x-slower-than-obj-dict

I'm actually not sure what value we're getting out of dataclasses, now that I think about it.  I think the idea is that they can be used when you need to make your structs hashable or comparable, but we aren't using any of that....

AlmightyYakob: The main benefit I get from dataclasses are:
- Description and type definition of our JSON
- Being able to work with our data in a more object oriented and clean way.

FWIW having `null`/`None` things be omitted is an easy thing to do, and I could try to replace the `asdict` call with calls to `__dict__`, but that may not work as I believe there are nested dataclasses in this PR. 

If we have serious performance concerns but still want to keep the benefit of typed data (which I think we should), we could switch to using `TypedDict` with `mypy`. Generally I think `mypy` would be a good addition anyway. My main point is that I'm not a fan of passing around these dictionaries willy nilly without any concrete definition of them somewhere. IMO, `TypedDict` would fill that roll, and wouldn't have a runtime performance hit, as they're purely for static type checking and are just a `Dict` at runtime. I think it would look a bit less clean than what I've proposed, but I'm also shrugging here.

AlmightyYakob:
> This I think is the real reason people choose msgpack. You get lightweight, robust payloads without having to make dumb schema decisions based on payload size.

I've never used MessagePack, and I'm not opposed to adding it, but from my understanding it would mean what we store at rest in girder would no longer be readable JSON files, correct? This would then not allow for direct interaction with these files on girder. I'm not saying that we necessarily need that, just something to consider.

subdavis:
@AlmightyYakob I wasn't suggesting moving to messagepack, just that this is the sort of problem it's intended to solve.

server/viame_server/serializers/viame.py
AlmightyYakob: @subdavis ef22b41 addresses your concerns around using `asdict`, lmk what you think.

PR #138, opened 2020-05-22
Add Continuous Deployment

### LINKED ISSUE #36: Set up testing and continuous deployment

AlmightyYakob:
Fixes #36 

Currently, this PR only updates code, and updating `deploy.yml` will have no effect, until updated manually on the server. If this is desired, the `main.yml` playbook would allow for this, as this runs both the systemd playbook, and the deploy playbook. However, the reason I haven't included it here is that I'm not sure if we even need to worry about the `systemd` files changing, and including that would mean that we need to do more work around storing a user's `sudo` password.

### DIFF
--- a/docker/ansible/README.md
+++ b/docker/ansible/README.md
@@ -1,5 +1,8 @@
 This folder contains the deployment configuration for Viame-Web
 
+
+Typically ansible is run using `ansible-playbook` over `ssh`, however since the server this is deployed on isn't accessable via `ssh`, we can't use normal CI/Deployment methods. Instead, ansible needs to be run with a `systemd` timer, and this `systemd` timer needs to be setup locally. Due to the uptime requirements, this deployment is run on a schedule, instead of directly after changes to the `master` branch.
+
 # Requirements
 The server must have the following things:
 - The repository cloned to `/home/viame`
@@ -8,7 +11,7 @@ The server must have the following things:
 
 
 # systemd service
-To initiate the systemd timer, run the following command:
+To initiate the `systemd` timer, run the following command:
 
 ```
 ansible-playbook docker/ansible/systemd.yml
--- a/docker/ansible/README.md
+++ b/docker/ansible/README.md
@@ -1,0 +1,24 @@
+This folder contains the deployment configuration for Viame-Web
+
+# Requirements
+The server must have the following things:
+- The repository cloned to `/home/viame`
+- Ansible
+- The Ansible [Docker Compose](https://docs.ansible.com/ansible/latest/modules/docker_compose_module.html) module
+
+
+# systemd service
+To initiate the systemd timer, run the following command:
+
+```
+ansible-playbook docker/ansible/systemd.yml
+```
+
+If you need to remove this timer and service, run the following command:
+
+```
+ansible-playbook docker/ansible/clean.systemd.yml
+```
+
+# Operation
+Once the service is initiated, it will run the `deploy.yml` playbook at the interval specified in `viame-deploy.timer`. This will update the server with any changes.
--- a/docker/ansible/clean.systemd.yml
+++ b/docker/ansible/clean.systemd.yml
@@ -1,0 +1,25 @@
+# Playbook to install the systemd service and timers files
+
+- hosts: localhost
+  become: yes
+  vars:
+    # Required to use docker_compose properly
+    ansible_python_interpreter: /usr/bin/python3
+  tasks:
+    - name: Disable timer
+      systemd:
+          name: viame-deploy.timer
+          state: stopped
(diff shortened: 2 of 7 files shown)


docker/ansible/README.md
```diff
@@ -0,0 +1,24 @@
+This folder contains the deployment configuration for Viame-Web
+
+# Requirements
+The server must have the following things:
+- The repository cloned to `/home/viame`
+- Ansible
+- The Ansible [Docker Compose](https://docs.ansible.com/ansible/latest/modules/docker_compose_module.html) module
```
subdavis, line 7: Typical ansible deployments involve running `ansible-playbook` to ssh into a server for provisioning.  I understand why that's not possible in this case, but I think you should explain the limitations and resulting architecture here, because these requirements would confuse a new developer.

Something to the effect of: "You have to run ansible with a systemd timer on the target server because it isn't accessible via SSH, so traditional CI/CD wont' work.  Also, this project's uptime requirements prohibit any kind of continuous deployment, and we have instead opted for scheduled overnight depolyments."

docker/ansible/deploy.yml
```diff
@@ -0,0 +1,21 @@
...
+
+    - name: Pull Repository Changes
+      shell: git pull
+      args:
+        chdir: /home/viame
+
+    - name: Restart and Build Service
+      docker_compose:
+        project_src: /home/viame/docker
+        build: yes
```
subdavis, line 21: Does this do the build before restarting, to minimize downtime?

docker/ansible/clean.systemd.yml
```diff
@@ -0,0 +1,25 @@
+# Playbook to install the systemd service and timers files
```
subdavis, line 1: This comment is incorrect.  Copy/paste?

docker/ansible/systemd.yml
```diff
@@ -0,0 +1,24 @@
...
+
+    - name: Copy systemd timer
+      copy:
+        src: /home/viame/docker/ansible/viame-deploy.timer
+        dest: /etc/systemd/system/viame-deploy.timer
+
+    - name: Enable timer
+      systemd:
+          name: viame-deploy.timer
+          state: started
```
subdavis, line 22: I haven't tested this, but are you certain this will cause changes in `viame-deploy.timer` to be picked up?

I see the `daemon-reload`, but since `state: started` doesn't cause the timer to bounce, I'm not sure if the changes in the timer file are picked up.  is `state: restarted` safer?   For example, if the schedule changes, but ansible doesn't actually restart the timer, do the schedule changes get noticed?

docker/ansible/viame-deploy.timer
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=Run the Viame deploy every Sunday and Thursday at 2 AM
+
+[Timer]
+OnCalendar=Sun 02:00
+OnCalendar=Thu 02:00
```
subdavis, line 6: This relies on the system clock, so these times are relative to the system time zone.    Should probably specify a time zone here.  

`Sun 02:00 UTC-5` would be EST, I believe, as would `Sun 02:00 America/New_York`

docker/ansible/clean.systemd.yml
AlmightyYakob: Yes, good catch

docker/ansible/deploy.yml
AlmightyYakob: This makes the command equivalent to:
```
docker-compose up --build
```

So rebuilds the images (much of which is skipped over because it hasn't changed), and then starts up the containers

docker/ansible/systemd.yml
AlmightyYakob: I've tested this and I believe what's there has the desired effect you're describing, but I can double check and change that to `restarted` to be safe.

docker/ansible/viame-deploy.timer
AlmightyYakob: I agree

docker/ansible/README.md
AlmightyYakob: ea4c365

docker/ansible/clean.systemd.yml
AlmightyYakob: d6053ee

docker/ansible/systemd.yml
AlmightyYakob: 96e5804

docker/ansible/viame-deploy.timer
AlmightyYakob: edd6ce2

PR #139, opened 2020-05-27
Save Track JSON

AlmightyYakob:
Depends on #136.

This PR changes the saving of detection into girder by always saving as Track JSON. This also maintains backwards compatibility. This is because if the latest detection file in girder is a CSV, it will call the functions that were already in place, but if not, it will either simply return the contents of the file, or convert it to `track_json` first, and then return it.

To maintain the CSV format in some way, I've added an `export_detections` endpoint, which serializes the JSON format back into the CSV format we've been using. This was mainly done by moving code from the `save_detection` endpoint (which used to do exactly this) into this endpoint.

Now, there are two cases for retrieving detections:
* Track JSON, from a CSV file
* Track JSON, from a JSON file

To my knowledge, I've made the response consistent between CSV/JSON backend storage, but if you find something I've missed, please lmk.

### DIFF
--- a/server/setup.py
+++ b/server/setup.py
@@ -9,6 +9,7 @@ requirements = [
     "girder_worker==0.6.0",
     "girder_worker_utils==0.8.5",
     "pysnooper",
+    "dacite",
 ]
 
 setup(
--- a/server/viame_server/event.py
+++ b/server/viame_server/event.py
@@ -2,10 +2,8 @@ from girder.models.folder import Folder
 from girder.models.item import Item
 
 from viame_server.utils import (
-    itemIsWebsafeVideo,
     validImageFormats,
     ImageSequenceType,
-    VideoType,
 )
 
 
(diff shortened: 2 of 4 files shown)


subdavis:
detection json is going to be removed soon, so I don't know how much sense it makes to merge code that stores data into detection json files, since we'd want to immediately transition to track json files.

This seems like it would cause a bit of extra headache to run migrations on. Would it make more sense to postpone this PR until after the client track work is done so that there are only 2 cases for retrieving and 1 for persisting?

AlmightyYakob:
> detection json is going to be removed soon, so I don't know how much sense it makes to merge code that stores data into detection json files, since we'd want to immediately transition to track json files.
> 
> This seems like it would cause a bit of extra headache to run migrations on. Would it make more sense to postpone this PR until after the client track work is done so that there are only 2 cases for retrieving and 1 for persisting?

Sounds good to me :+1:

AlmightyYakob:
I may still have to address the output of running pipelines.

AlmightyYakob:
It seems like the pipelines output will work fine, as when retrieving detections the backend can still read in CSV, but I can't definitely test it because I don't have a GPU on my machine.

subdavis:
FYI this branch has been integrated in #137 and can be closed

PR #146, opened 2020-06-04
Use python3.7 instead of python

AlmightyYakob:
This was never bumped in #136, so before this change it failed to build. We might want to consider adding CI that runs the build to ensure it succeeds.


AlmightyYakob:
I've tested a full build of this locally and it properly ran all services. Before this, the `viame_web_girder_worker` service wasn't running

PR #161, opened 2020-06-15
final tidying before merge

subdavis:
* Swap export manifest endpoint to return the `export_detections` endpoint Jacob wrote
* Allow cookie auth so the user can actually initiate a download
* Move `girder.ts` into the plugins folder


server/viame_server/viame_detection.py
```diff
@@ -75,8 +75,7 @@ def export_data(self, folder):
         clipMeta = self._get_clip_meta(folder)
         detection = clipMeta.get('detection')
         if detection:
-            itemId = detection.get('_id', None)
-            export_detections = f'/api/v1/item/{itemId}/download'
```
subdavis, line 79: @AlmightyYakob your export_detections endpoint works great, I just had to change these lines to actually use it.

server/viame_server/viame_detection.py
```diff
@@ -55,7 +55,7 @@ def _get_clip_meta(self, folder):
             'videoUrl': videoUrl,
         }
 
-    @access.public(scope=TokenScope.DATA_READ, cookie=True)
+    @access.user
```
AlmightyYakob, line 58: Why does this require user login, but the `export_detections` endpoint below is public?

subdavis: This endpoint returns JSON meant to be ajax'd on the page.

The `export_detections` endpoint is a download link, so the browser must put it in a `href` with `target=_blank` to open it in a new tab and trigger the download.  Naturally you can't set headers with an href, so you have to use cookie authentication, which is generally unsafe (CSRF) but works fine when you're doing read-only operations like download.

Public just means no authorization header, but you'll still 403 if you try to download it without a cookie for a user without read access.

All the existing girder views (item, file) work like this when you use `/<model>/download`.

subdavis:
> When downloading videos/everything for a video based sim it gives you both the source video and the ffmpeg converted video as well.

MM, yeah that should probably change.  Probably want to let them download the original non-transcoded video.

PR #176, opened 2020-06-25
TS Fix

subdavis:
Resolves the issue with coercing the JSON request from axios.

### DIFF
--- a/client/src/lib/api/viame.service.ts
+++ b/client/src/lib/api/viame.service.ts
@@ -13,6 +13,16 @@ interface Attribute {
   _id: string;
 }
 
+interface Pipe {
+  name: string;
+  pipe: string;
+  type: string;
+}
+export interface Category {
+  description: string;
+  pipes: [Pipe];
+}
+
 function makeViameFolder({
   folderId, name, fps, type,
 }: {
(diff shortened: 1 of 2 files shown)


client/src/lib/api/viame.service.ts
```diff
@@ -78,7 +88,7 @@ function runImageConversion(folder: string) {
 }
 
 function getPipelineList() {
-  return girderRest.get('viame/pipelines');
+  return girderRest.get<Record<string, Category>>('viame/pipelines');
```
subdavis, line 91: @BryonLewis I discovered this syntax recently and it really seems the most correct to me.

client/src/store/Pipelines.ts
```diff
@@ -27,10 +18,10 @@ const pipelineModule: Module<PipelineState, never> = {
   actions: {
     async fetchPipelines({ commit, state }) {
       if (state.pipelines === null) {
-        const { data } = await getPipelineList() as {data: PipelineState};
+        const { data } = await getPipelineList();
```
subdavis, line 21: I think the issue you were having is that `PipelineState` has a key `pipelines` that isn't present in the actual json response.

subdavis:
> Seems a little weird to import the Category from the viame.service

The source of the datatype is the server, so it makes sense (to me) that the library or module that brokers server requests would also own the data types it produces.

From what I've seen from talks from other organizations, a popular way to do this is with OpenAPI codegen, so the backend team produces a swagger document, then the frontend team autogenerates an axios-based typescript library with all the methods and data-types defined in the Swagger spec.  When the backend team changes the API, the library is rebuilt and the frontend fails to compile.

I've used https://github.com/OpenAPITools/openapi-generator once before and it works well if you're rigorous about keeping your swagger docs up-to-date.

PR #180, opened 2020-06-25
Fix root browser

### LINKED ISSUE #179: World button on viame.kitware.com doesn't seem to go back to the universal (shared) training data location
See images, if I press the 'world' it doesn't go back to the training data page like it use to, this should also probably be in the documentation somewhere what these buttons do and where they go, heard one comment about it being unclear, also how users can upload data and then share the links to the data with other users (public vs private). The only way to get back to the training data page appears to be to go to viame.kitware.com again, no GUI option which is bad

world_no_work

subdavis:
Clicking on the globe to go to the root was broken.  

Fixes #179

### DIFF
--- a/client/src/components/TrackList.vue
+++ b/client/src/components/TrackList.vue
@@ -51,7 +51,13 @@ export default Vue.extend({
   }),
 
   computed: {
-    virtualListItems() {
+    virtualListItems(): {
+      trackId: number;
+      selectedTrackId: number;
+      checkedTrackIds: number[];
+      editingTrack: boolean;
+      allTypes: string[];
+    }[] {
       const selectedTrackId = this.selectedTrackId.value;
       const checkedTrackIds = this.checkedTrackIds.value;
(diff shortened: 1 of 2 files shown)


client/src/components/TrackList.vue
```diff
@@ -51,7 +51,13 @@ export default Vue.extend({
...
 
   computed: {
-    virtualListItems() {
+    virtualListItems(): {
+      trackId: number;
+      selectedTrackId: number;
+      checkedTrackIds: number[];
+      editingTrack: boolean;
+      allTypes: string[];
+    }[] {
```
AlmightyYakob, line 60: Are the changes in this file unrelated?

subdavis: Yeah.   There's some dumb stuff you have to do to get typescript to work in Object API components.  

This block of code was missing, which is why I had to `// @ts-ignore` below.

PR #195, opened 2020-07-02
Move docker docs to another README

subdavis:
New README: https://github.com/VIAME/VIAME-Web/blob/fd6ac189499df3dde555a81badbf246003e4ae1e/README.md
New Docker README: https://github.com/VIAME/VIAME-Web/blob/fd6ac189499df3dde555a81badbf246003e4ae1e/docker/README.md

### DIFF
--- a/docker/girder.Dockerfile
+++ b/docker/girder.Dockerfile
@@ -1,5 +1,6 @@
 FROM node:latest as builder
 WORKDIR /app
+
 # Install dependencies
 COPY client/package.json client/yarn.lock /app/
 RUN yarn --frozen-lockfile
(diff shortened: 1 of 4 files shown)


docker/girder.Dockerfile
```diff
@@ -11,6 +12,12 @@ RUN yarn build
 
 FROM girder/girder as runtime
 
+ENV GIRDER_MONGO_URI mongodb://mongo:27017/girder
+ENV GIRDER_ADMIN_USER admin
+ENV GIRDER_ADMIN_PASSWORD viame
+ENV CELERY_BROKER_URL amqp://guest:guest@rabbit/
+ENV BROKER_CONNECTION_TIMEOUT 2
```
AlmightyYakob, line 19: Why was this added? Do these serve as a default for the variables specified in `docker-compose.yml`? Or if not, how do they interfere with them?

docker/girder_worker.Dockerfile
```diff
@@ -2,6 +2,9 @@ FROM kitware/viame:gpu-all-models-latest
 # expects VIAME install at /opt/noaa/viame/
 # expects VIAME pipelines at /opt/noaa/viame/configs/pipelines/
 
+ENV CELERY_BROKER_URL amqp://guest:guest@rabbit/
+ENV BROKER_CONNECTION_TIMEOUT 2
```
AlmightyYakob, line 6: Same question as above

docker/girder.Dockerfile
subdavis: Docker images that accept optional environment configuration should typically list them in the dockerfile.  This serves largely as documentation.  

These do serve as defaults if the images are used alone.  They're overridden by the vars supplied in docker-compose.yml

PR #196, opened 2020-07-02
Validfile endpoint

mattrix27:


client/src/components/Upload.vue
```diff
@@ -5,15 +5,17 @@ import {
...
   VideoType,
 } from '@/constants';
 import { makeViameFolder } from '@/lib/api/viame.service';
+import { mapActions, mapState } from 'vuex';
 import { getResponseError } from '@/lib/utils';
 
-function prepareFiles(files) {
+function prepareFiles(files, filetypes) {
+  const videoFilesRegEx = new RegExp(`${filetypes.video.join('$|')}$`, 'i');
+  const imageFilesRegEx = new RegExp(`${filetypes.image.join('$|')}$`, 'i');
```
subdavis, line 17: I see these lines repeated in several places.  Might be nice to just refactor them out into a Vuex Getter and get rid of the `mapState`.   See comment below.

client/src/store/Filetypes.ts
```diff
@@ -0,0 +1,31 @@
...
+      state.filetypes = filetypes;
+    },
+  },
+  actions: {
+    async fetchFiletypes({ commit, state }) {
+      if (state.filetypes === null) {
+        const { data } = await getValidFileTypes();
+
+        commit('setFiletypes', data);
+        return data;
```
subdavis, line 24: No need to return early here.   Easier to guarantee consistent behavior with fewer code paths.

client/src/lib/api/viame.service.ts
```diff
@@ -18,6 +18,13 @@ interface Pipe {
   pipe: string;
   type: string;
 }
+
+export interface Types {
+  video: [string];
+  image: [string];
+  web: [string];
```
subdavis, line 25: `string[]` ?

I belive `[string]` indicates an array with 1 element of type string.

client/src/components/Upload.vue
```diff
@@ -173,13 +181,13 @@ export default {
...
       try {
-        this.addPendingUpload(name, files);
+        this.addPendingUpload(name, files, this.filetypes);
       } catch (err) {
         this.preUploadErrorMessage = err;
       }
     },
-    addPendingUpload(name, allFiles) {
-      const { type, media, csv } = prepareFiles(allFiles);
+    addPendingUpload(name, allFiles, filetypes) {
```
subdavis, line 189: No need to make this a method param.
```suggestion
    addPendingUpload(name, allFiles) {
      // pull regexps from store, or use `mapGetters`
      const { regexps } = this.$store.getters.Filetypes;
```

In general, if you find yourself passing instance properties to instance methods, that's a smell, since they share the same context.

client/src/components/Upload.vue
mattrix27: Oh yeah, I was looking at this earlier, but when I tried to make the RegEx's in the module, it threw an error cuz 'state' was possibly null?

server/viame_server/viame.py
```diff
@@ -152,3 +156,8 @@ def update_attribute(self, data, attribute, params):
     @autoDescribeRoute(Description("").modelParam("id", model=Attribute, required=True))
     def delete_attribute(self, attribute, params):
         return Attribute().remove(attribute)
+
+    @access.user
+    @describeRoute(Description("Get Valid Image/Video filetypes for uploading"))
+    def get_valid_filetypes(self, params):
+        return {'image': validImageFormats, 'video': validVideoFormats, 'web': webValidImageFormats}
```
AlmightyYakob, line 163: We use double quotes throughout the codebase, and more generally, code is formatted with [black](https://black.readthedocs.io/en/stable/the_black_code_style.html). Formatting this with black gives me the following (also added an EOF newline):

```suggestion
        return {
            "image": validImageFormats,
            "video": validVideoFormats,
            "web": webValidImageFormats,
        }

```

client/src/views/Home.vue
```diff
@@ -32,6 +31,8 @@ export default {
     uploading: false,
   }),
   computed: {
+    ...mapState('Filetypes', ['filetypes']),
+    ...mapState('Location', ['location']),
```
AlmightyYakob, line 35: I don't see why `location` was added, there's already a computed getter and setter below.
```suggestion
```

client/src/store/Filetypes.ts
AlmightyYakob: I think that unless there's a reason to keep it here, that we should remove the return statements altogether. We're not currently saving the return value of this function anyway, and keeping the state as the source of truth seems like a good practice to me.

client/src/components/Upload.vue
```diff
@@ -113,11 +115,17 @@ export default {
...
   }),
   computed: {
+    ...mapState('Filetypes', ['filetypes']),
     uploadEnabled() {
       return this.location && this.location._modelType === 'folder';
     },
   },
+  created() {
+    this.fetchFiletypes();
+  },
```
AlmightyYakob, line 125: Both this and `Home.vue` contain this line. I believe `Home.vue` is the only place `Upload.vue` is used, so I think removing it from here and having it solely in `Home.vue` would be best. That way, we're not making unnecessary fetches/store updates, and it's less confusing where/when this action is dispatched.

client/src/components/Upload.vue
subdavis: I don't know about state itself being null.  but the getter would need to handle the case where filetypes is null.

client/src/components/Upload.vue
subdavis: The action contains logic to avoid the API call if the state already exists.

Home.vue does actually need the regexes in the uploaded hook to kick off transcoding.  IMO, that logic should be on the server, but that's an issue for another PR.

client/src/store/Filetypes.ts
```diff
@@ -0,0 +1,48 @@
...
+    filetypes: null,
+  },
+  mutations: {
+    setFiletypes(state, filetypes) {
+      state.filetypes = filetypes;
+    },
+  },
+  getters: {
+    getVidRegEx(state) {
+      if (state !== null) {
```
subdavis, line 20: `state` should never be null.  Did you mean to check `state.filetypes` ?

client/src/lib/api/viame.service.ts
```diff
@@ -110,6 +111,11 @@ function setMetadataForFolder(folderId: string, metadata: object) {
   );
 }
 
+function getValidFileTypes() {
+  return girderRest.get<Record<string, any>>('viame/valid_filetypes');
```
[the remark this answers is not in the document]
mattrix27: So I recreated the error I had before, but when I had it be the type definition I made (Types), it threw an error saying the state doesn't have the join property to create the regex. I tried casting as an Array but then doesn't seem to find the correct filetypes.

Screenshot from 2020-07-07 10-35-20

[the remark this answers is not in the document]
subdavis: https://github.com/VIAME/VIAME-Web/pull/206

client/src/components/Upload.vue
```diff
@@ -282,7 +272,7 @@ export default {
           const subfile = files.splice(0, 1);
           const subname = subfile[0].file.name.replace(/\..*/, '');
           // Only video subfolders should be used typically
-          const subtype = videoFilesRegEx.test(subfile[0].file.name) ? 'video' : 'unknown';
+          const subtype = this.vidRegEx.test(subfile[0].file.name) ? 'video' : 'unknown';
```
AlmightyYakob, line 275: Is this supposed to be `this.getVidRegEx`?

client/src/views/Home.vue
```diff
@@ -8,6 +7,11 @@ import RunPipelineMenu from '@/components/RunPipelineMenu.vue';
 import Upload from '@/components/Upload.vue';
 import NavigationBar from '@/components/NavigationBar.vue';
 import { getPathFromLocation, getLocationFromRoute } from '@/utils';
+import {
+  mapActions,
+  mapMutations,
+  mapGetters,
+} from 'vuex';
```
AlmightyYakob, line 14: Not a blocker, but just a note: Generally speaking, third party module imports are separated from internal imports, with third party imports usually going above internal ones.

subdavis: Actually, that should be a lint error.  I believe the `@` tricks eslint into believing everything is an external import.  I'll have to look into that, but please hoist this above the internal imports.

mattrix27: should I do this for Upload.vue too?

AlmightyYakob: Looks like it's fixed in 9fb16c7

client/src/components/Upload.vue
subdavis: It would appear so.

subdavis: @mattrix27 I think this is the last thing to check before merging.

mattrix27: Oh yeah, it is supposed to

client/src/store/Filetypes.ts
```diff
@@ -0,0 +1,48 @@
+import { Module } from 'vuex';
+import { getValidFileTypes } from '@/lib/api/viame.service';
+
+export interface FiletypeState {
+    filetypes: null | Record<string, string[]>;
```
AlmightyYakob, line 5: You actually could more strictly type this, since you know the `filetypes` object will consist of `image`, `video` and `web` at the top level. I don't think it's necessary though.

PR #199, opened 2020-07-02
Better form item alignment

subdavis:


client/src/components/TypeList.vue
```diff
@@ -74,45 +75,47 @@ export default Vue.extend({
...
+              dense
+              shrink
+              hide-details
+              class="my-1"
+            />
+            <v-spacer />
+            <v-tooltip
+              open-delay="100"
+              bottom
+            >
```
BryonLewis, line 100: I didn't like the behavior for long type names:
image

Much rather find a way to do wrapping like this with the editing button always in the same position and wrapping of the name below the current below:
image

subdavis: Reverted the removal of `type-checkbox`, this appears to be working now.

client/src/components/TypeList.vue
```diff
@@ -237,6 +237,11 @@ export default Vue.extend({
 </template>
 
 <style scoped lang='scss'>
+.type-checkbox {
+  max-width: 80%;
+  overflow: hidden;
```
BryonLewis, line 242:
```suggestion
  overflow-wrap: anywhere;
```
My original  `overflow: hidden` was probably the wrong thing to do.
This doesn't maintain item height for all types, but will prevent the name from being cut off.  Other option is to truncate it with a hover tooltip of the full name.

image

PR #206, opened 2020-07-07
Suggestion for any type

subdavis:
I took a look at your issue and this code is working for me.   Does this make sense, or am I missing something?


mattrix27:
Ohhh okay, I thought I had to create a custom interface for each (video, image, web), but just doing string[] seems to work!

PR #232, opened 2020-07-23
Continue if build failed, send failure report

### LINKED ISSUE #231: Send emails on failed deploy

AlmightyYakob:
Fixes #231 

This updates the behavior of the builds to start the service independent of building, so that if it fails, the service can come back online without issue. Furthermore, if the build fails, it will send a failure report to the email address we specify.

### DIFF
--- a/docker/ansible/.gitignore
+++ b/docker/ansible/.gitignore
@@ -1,0 +1,1 @@
+.vars.yml
--- a/docker/ansible/.vars.example.yml
+++ b/docker/ansible/.vars.example.yml
@@ -1,0 +1,4 @@
+---
+deploy_email_address: address.to.send.from@gmail.com
+notify_email_address: address.to.send.to@gmail.com
+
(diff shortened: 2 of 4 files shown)


AlmightyYakob:
I've tested this on the server, and the mail sending works without issue.

AlmightyYakob:
> This is CI config, so I'm just going to trust that it works.
> 
> Is there a reason you used a vars file rather than just entering the variable into CircleCI env config?

This runs on the viame server, not in CI.

subdavis:
Oh, I see.  Could you specify where to put `vars.yml` in the README?  Is it in the project root or inside `docker/ansible`?

AlmightyYakob:
> Oh, I see. Could you specify where to put vars.yml in the README? Is it in the project root or inside docker/ansible?
> 
> Better yet, could you place a .vars.example.yml in the place where it's supposed to go with the correct schema and dummy values?

Good idea, I'll do that.

PR #251, opened 2020-08-03
Run postprocessing after run_pipeline

subdavis:
@AlmightyYakob I remember I removed `viame: True` once before and something broke.  I did it again because it seems vestigial, but it would be nice to know what actually depends on it.

After this, `csv` and `yml` annotations will no longer have the `{ detection: <folderid> }` meta: they will be transformed into JSON immediately after being created by the postprocessing endpoint then moved into auxillary.

I didn't actually change that much, but I added a conditional around some of the stuff in postprocess that can potentially spin off jobs so that when postprocess is called after a `run_pipeline`, it doesn't accidentally run transcoding of media and stuff again.


subdavis:
@AlmightyYakob because CSV are no longer valid annotation files on the server, a migration will need to run.

All we have to do is list all annotation folders on the server, and for each one, call `/api/v1/viame/postprocess/<folderid>` with `skipJobs=true`.  I'm not sure what the best way to do that is.

server/viame_server/viame.py
```diff
@@ -126,74 +122,91 @@ def validate_files(self, files):
...
 
-        # Preprocess CSV if necessasry
+            Folder().save(folder)
+
+        # transform CSV if necessasry
         csvItems = Folder().childItems(
-            folder, filters={"lowerName": {"$regex": csvRegex}}
+            folder,
+            filters={"lowerName": {"$regex": csvRegex}},
+            sort=[("created", pymongo.DESCENDING,)],
```
mattrix27, line 200: A small thing mainly for me. Is Girder built directly off of MongoDb/could you grab this from girder? Or is it just easier/clearer this way?

subdavis: Girder models are built on pymongo.  These functions are just sort of thin wrappers around pymongo queries, and they support adding additional pymongo query params.

subdavis: https://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.find

AlmightyYakob:
@subdavis testing this locally I didn't notice anything missing regarding `viame: true` not being present :+1:

server/viame_tasks/tasks.py
```diff
@@ -91,21 +91,23 @@ def run_pipeline(self, input_path, pipeline, input_type):
...
+        output_path = detector_output_path
 
+    newfile = self.girder_client.uploadFileToFolder(output_folder, output_path)
 
-@app.task(bind=True)
-def convert_video(self, path, folderId, token, auxiliaryFolderId):
-    self.girder_client.token = token
+    self.girder_client.addMetadataToItem(newfile["itemId"], {"pipeline": pipeline,})
+    self.girder_client.post(
+        f'viame/postprocess/{output_folder}', data={"skipJobs": True}
```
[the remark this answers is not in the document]
subdavis: Yes.

subdavis:
@AlmightyYakob and @mattrix27  It's actually a little too hasty to drop support for CSV and try and do the migration all at once.

244d0f1 rolls back those changes so that no new CSV will ever be created, but the existing ones will continue to work.  This should remove the need for any migration at this time.  I'd appreciate if someone could verify my logic and make sure I didn't forget any edge cases.

If this works, we're good to start the merge train.

PR #257, opened 2020-08-04
Polygon/brandon

subdavis:
@BryonLewis I think I'm done with this.    Please consider this my active exploration of the organization of polygon state.  I didn't touch anything inside the actual layers, I mostly just rearranged stuff.

## Summary of suggestions

* Drop `useAnnotationMode`, combine with `useModeManager`
* Drop anything related to the help system.  I think a help system is a great idea, but I'd like for there to be a more holistic approach to help messaging and something that can be hooked into app-wide.  See #228 discussion for more on that.
* Drop unnecessarily complex `annotationModes` structure YAGNI.  I didn't find the `enabled/visible/hidden/selected/etc.` enum easy to understand, and I didn't think that transforming between the structure in `annotationModes.state` and the derived data structure used for the UI in `EditorMenu` seemed neccessary.  Replaced with a single string (for edit more) and an array of strings (for visible mode).  You can pass these directly to v-model and vuetify will handle a ton of that logic for you.
* Separete point deletion controls from mode toggle controls.
* Use `v-btn-toggle` for visibility toggles because it was designed for exactly this use case.
* Change colors for visible/editing mode.  We use blue everywhere else to indicate "selection" so it  was weird to have a red track turn blue when it enters edit mode, but a blue toggle menu turn red.  
* Use `editingTrack` state directly rather than setting up a watcher and writing derived state into another reactive value.

Screenshot from 2020-08-04 14-45-40

Screenshot from 2020-08-04 14-25-53

Here's how it looks.   I didn't really like the toolbar-within-toolbar look so I did this.  If  there are elements of the old style you prefer, LMK.  I put the select handle index in the delete button so you could see it update as sort of a confidence booster that yes, those things are connected.  It might be too much for the user.

## Things that are incomplete

* You're really good about adding tooltips to buttons.  I removed them while I was mucking around, and didn't put them back yet.

## Bugs I found that I didn't try to fix yet:

* When selecting a polygon handle point...
  * If the point on the polygon falls outside the bounding box, you get kicked out of edit mode rather than doing the selection
  * If you select a point, then toggle edit to bounding box and back to polygon, the handle state is still set (didn't reset to -1) but the point isn't highlighted.  I may have broken this myself, I'll have to check.


subdavis:
Thanks for these comments.

> Remove deletion options from rects,

Fixed

> Select rect for editing -> select handle -> click off to deselect -> notice that delete option is still available

Fixed, I think.

> Toggle for editing mode isn't behaving as a "select one",

Fixed

> View Mode Icon: It shouldn't be a button with a hover state

There's apparently no way to disable the hover state for a chip, but I can implement my own "chip clone".  This is easy: it's just a span with a background and border radius.

Fixed

> having the mode icon be smaller than the delete point icon seems off.

Fixed

> Also I don't know if the mode should have higher priority in a left -> right sort of flow like: Mode -> buttons -> deletion?

I try to avoid making things shift position to accommodate a v-if element.  It seems like bad UX to have something move when it's unrelated to a user action.

> toggling between selected and not in edit mode I think needs a bit more indication. 

Visually, I preferred your buttons too.  But making those buttons work required you to write the `v-model` mutation logic that `v-btn-toggle` gives you for free.   The trade-off is that v-btn-toggle is annoyingly not very customizable.  Any logic we can push off to Vuetify is something we don't have to worry about breaking ourselves, especially since we aren't writing component tests at this point in developemnt.

I think I'd prefer to keep it simple with the `v-btn-toggle` and the limited style choices available to that component.

PR #263, opened 2020-08-05
Fix small parser bug

subdavis:
Fixes parser errors for missing or empy CSV column data.


subdavis:
@AlmightyYakob this is pretty small, ptal?

PR #264, opened 2020-08-05
Hide button text on breakpoints

subdavis:
Screenshot from 2020-08-05 13-27-11

Remove button text in toolbar on page breakpoints to make room for the stuff we've added to the toolbar.

Below a certain width, this app becomes useless, so I'm not going to agonize over proper scaling at mobile widths.

### DIFF
--- a/client/src/components/Export.vue
+++ b/client/src/components/Export.vue
@@ -54,14 +54,15 @@ export default {
         :small="small"
         v-on="on"
       >
-        <v-icon
-          left
-          color="accent"
-          class="mdi-24px mr-2"
-        >
+        <v-icon color="accent">
           mdi-export
         </v-icon>
-        Download
+        <span
+          v-show="!$vuetify.breakpoint.mdAndDown"
+          class="pl-1"
+        >
+          Download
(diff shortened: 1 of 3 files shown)


subdavis:
> As always I'm going to suggest a separate issue for tool tips if we have a condition where we can't see the text for a button.

That makes sense to tackle here, so I'm going to consider that "requested changes"

subdavis:
* Updated with tooltips for download and run pipeline.  I think the help button is a universally recognized thing, so a tooltip below a question mark that just says "show help" seems noisy and unhelpful.  
* Fixed small bug in the export dialog to not show the threshold checkbox if no thresholds exist.
* Changed icon for download to a more traditional "download" instead of "export"

subdavis:
> Not needed to be worried about here, but probably once all stuff is merged. I didn't double check but if the Delete Point button doesn't have a placeholder space I don't want that to swap back and forth between collapsed and expanded button views when the delete point toggles.

Collapse is based purely on window width breakpoints.   Inner content overflow will just run off the page: it won't cause collapse.

PR #296, opened 2020-08-19
Add more interesting imagery

subdavis:
Intended to be eye-catching and show off a little more of what's been done over the last few months.

Link to diff: https://github.com/VIAME/VIAME-Web/pull/296/files?short_path=04c6e90#diff-04c6e90faac2675aa89e2176d2eec7d8

I don't want to confuse the messaging about what this tool is, but because these features were developed with general-purpose annotation in mind, it might make sense to showcase that?

In any case, I'd like to show an example of "video" type annotation with bounding boxes and more polygon-type annotation of a single still.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,9 +8,9 @@ Additional documentation will be available in the future for users.
 <p align="center">
 <br>
 <nobr>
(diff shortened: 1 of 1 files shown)


mattdawkins:
You may want to run the diva image by roddy or keith just to make sure as well

subdavis:
I checked, they approved.

PR #298, opened 2020-08-21
Add pan/zoom reset button

subdavis:
* Remove explicit column sizes in control bar.
* add button with tooltip to reset zoom/pan of video, mostly helpful while annotating a polygon.

Wait for upstream to merge, blocked on #270 

Screenshot from 2020-08-21 11-35-35


client/app/components/UserGuideDialog.vue
```diff
@@ -12,50 +12,53 @@ export default {
...
               name: 'Zoom In/Out', icon: 'mdi-mouse', actions: ['Scrollwheel Up/Down'], description: 'use scrollwheel to zoom in and out',
             },
             {
               name: 'Zoom Area', icon: 'mdi-mouse', actions: ['Shift + Mouse Movement'], description: 'Zoom into a specific area',
             },
+            {
+              name: 'Reset zoom', icon: 'mdi-keyboard', actions: ['R Key'], description: 'Reset pan and zoom',
+            },
             {
               name: 'Select Track', icon: 'mdi-keyboard', actions: ['Up/Down Arrows'], description: 'Select Track',
```
BryonLewis, line 30: `name` is used as the v-for key and causes a Vue warn here about duplicate keys because there are now two data items with the `name='Select Track'`, just suggest changing to the v-for to  ```:key="`${item.name}_${index}`"```.

client/src/components/controls/Controls.vue
```diff
@@ -116,35 +111,35 @@ export default {
...
       </v-container>
     </v-toolbar>
   </div>
 </template>
-
-<style lang="scss" scoped>
-.controls-row {
-  min-width:420px;
-}
-</style>
```
BryonLewis, line 150: This is really a minor thing and can be ignored.
Removing this and not replacing it with something similar, whether that be a min-width for the v-toolbar or another item causes this at small widths:
image
If you add in a width it will behave more like the toolbar at the top of the app where items are pushed offscreen instead of wrappingr.  There might be some no-wrapping arguments that can be used to do this as well:
image

subdavis:
Fixed.  Added index to key, added some min-width.   Since `app` injects arbitrary-width controls into those slots, I made controls only aware of the play/skip width and constrained the window width to at least 700.  Anything below 1200 is unusable to me, but we really shouldn't be attempting to support mobile-width layouts IMO.

PR #310, opened 2020-08-31
Fix folder metadata on import, use utils.csvRegex

AlmightyYakob:
Small fix for imports, we had not updated this function to add `annotate: True` to the folder metadata. It also uses the general `csvRegex` instead of looking for exactly `annotations.csv`.

### DIFF
--- a/server/viame_server/event.py
+++ b/server/viame_server/event.py
@@ -1,7 +1,7 @@
 from girder.models.folder import Folder
 from girder.models.item import Item
 
-from viame_server.utils import ImageSequenceType, validImageFormats
+from viame_server.utils import ImageSequenceType, csvRegex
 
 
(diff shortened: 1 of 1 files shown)


server/viame_server/event.py
```diff
@@ -10,16 +10,20 @@ def check_existing_annotations(event):
...
+    objectType = info.get("type")
+    importPath = info.get("importPath")
 
-    if "annotations.csv" in info["importPath"]:
+    if not importPath or not objectType or objectType != "item":
+        return
+
+    if csvRegex.search(importPath):
+        # Update file metadata
         item = Item().findOne({"_id": info["id"]})
```
subdavis, line 21: What if this is a folder named `folder.csv`?  Should there be a `_modelType` check here?

PR #312, opened 2020-09-01
Add linting and import sort to CI

subdavis:
Prevent merge if linting and import sorting aren't done.

This also cleans up missed linter and sorter issues, so it might have to wait for other server PRs to merge to avoid conflicts.

### DIFF
--- a/.github/workflows/blank.yml
+++ b/.github/workflows/blank.yml
@@ -42,5 +42,8 @@ jobs:
         pip install .;
         pip install -r dev-requirements.txt;
     - name: Test with pytest
-      run: |
-        pytest tests/
+      run: pytest tests/
+    - name: Lint with black
+      run: black --check .
+    - name: Check import linting with isort
+      run: isort --check .
--- a/server/dev-requirements.txt
+++ b/server/dev-requirements.txt
@@ -1,4 +1,4 @@
-black
+black>=20.8b0
 flake8
 isort
 pytest
--- a/server/pyproject.toml
+++ b/server/pyproject.toml
@@ -1,7 +1,7 @@
 [tool.black]
 line-length = 88
 skip-string-normalization = true
-target-version = ['py37']
+target-version = ['py38']
 exclude='\.eggs|\.git|\.mypy_cache|\.tox|\.venv|_build|buck-out|build|dist|venv'
 
 [tool.isort]
--- a/server/pyproject.toml
+++ b/server/pyproject.toml
@@ -9,4 +9,3 @@ line_length = 88
 use_parentheses = true
 include_trailing_comma = true
 multi_line_output = 3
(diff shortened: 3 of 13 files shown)


subdavis:
Lint failing on the server but passing locally.  I don't know what's going on.

subdavis:
Black upgraded to v20 and apparently has some new rules.  https://github.com/psf/black/blob/master/CHANGES.md#208b0

Note that you'll need to reinstall dev-requirements.txt to get the fresh version.

AlmightyYakob:
Instead of directly using `black` and `isort`, I'd prefer if we were to use flake8 for linting, and add plugins that suit our needs, which would include [flake8-black](https://pypi.org/project/flake8-black/) and [flake8-isort](https://pypi.org/project/flake8-isort/). This would also allow us to add any number of plugins that we like. What do you think?

subdavis:
I'm not really familiar with flake8 and what value it provides.  I like black and isort because they're pretty trivial to run and configure, but I'm open to something else if it's a D&A standard.

I've probably already spent more time this morning figuring this out than I should have, but if you'd like to add to this PR you're welcome to do so.

I don't really like that `flake8` doesn't support pyproject.toml because it's my understanding that pyproject.toml has sort of been blessed as the standard configuration method python, but this may not be a big deal.  I like that pyproject.toml config is picked up in my virtual env no matter what subfolder I'm in.  

This PR doesn't include any sort of precommit hook, which I'd also like to have, but I was too lazy.

AlmightyYakob:
> I'm not really familiar with flake8 and what value it provides. I like black and isort because they're pretty trivial to run and configure, but I'm open to something else if it's a D&A standard.

I've always seen it used when doing linting on a project, and it provides a really nice plugin system for linting.

> I don't really like that `flake8` doesn't support pyproject.toml because it's my understanding that pyproject.toml has sort of been blessed as the standard configuration method python, but this may not be a big deal. I like that pyproject.toml config is picked up in my virtual env no matter what subfolder I'm in.

The entire python community and myself agree that flake8 should support `pyproject.toml`, yet it's still missing. However I think you can still specify configs in `pyproject.toml` for tools that are used through flake8, like `black` and `isort`. If that's correct then the current config we have could remain unchanged. I can look into this.

> This PR doesn't include any sort of precommit hook, which I'd also like to have, but I was too lazy.

We've done this on [multinet](https://github.com/multinet-app/multinet-server), so I could add this if it's desired.

subdavis:
@AlmightyYakob this is blocking server work.  Can we merge this and follow up with flake8 later?

AlmightyYakob (dismissed):

subdavis:
@AlmightyYakob had to rebase, could you approve again? thanks.

PR #313, opened 2020-09-02
Fixes 307

### LINKED ISSUE #307: [BUG] Deleted polygon annotation on first frame of sequences stays there after deletion
**To Reproduce**
1. Go to https://viame.kitware.com/#/viewer/5eb371608d01d968cc1642dd
2. Draw a polygonal annotation
3. Finalize it by right clicking
4. Select it again in list
5. Press trash can in list to delete
6. Deletes from list but annotation stays there, also borks frame if you come back to it again doesn't show correct annotations for the first frame

subdavis:
Another consequence of https://github.com/alexbol99/flatten-interval-tree/issues/16

Switch to string keys so we never have falsey keys.

fixes #307

### DIFF
--- a/client/src/components/LayerManager.vue
+++ b/client/src/components/LayerManager.vue
@@ -115,11 +115,9 @@ export default defineComponent({
       featurePointing: boolean,
       visibleModes: EditAnnotationTypes[],
     ) {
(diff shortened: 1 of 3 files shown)


subdavis:
Adding a test for this, Hol up.

PR #318, opened 2020-09-03
Phase 1 recipes

subdavis:
## Done
* Recipe interface.
* Simplify after exit logic for changing bounding box
* Allow update geojson and update in progress geojson to use the same handler in useMOdeManager
* 
# To Do

- [ ] Bring back `h` and `t` hotkeys for starting with head and tail
- [ ] Support exiting early without laying the second point
- [ ] Document everything really well
- [ ] UI controls are currently a mess, I don't really know what I want to do with them


BryonLewis:
I'll look into the code and stuff for this in a bit, but I just wanted to log the issues and how to cause them so I can go through and start working on remedies.

Hitting Escape after switching to Head/Tails recipe and then clicking again produces an error: Related to calling `hanleUpdateGeoJSON` without a selected ID.  Under further testing it looks like anytime a new feature is being created, or the user is in edit mode it will error out after hitting escape.  This might have a relation to the editAnnotationLayer mode changing signal?

Deletion isn't working for points in a line because it is defaulting to attempting to delete the point of the line, not realizing that it is a head/tails line.  So it responds with the fact that there must be at least two points in a line.  I'm guessing point removal is something that needs to have an optional injection like the geoJSON update for recipes.

There seems to be an issue where I can trick it into a regular line mode instead of head/tails.
Steps:
1. Click an existing track and edit it to create a head/tails item
2. Right after placing the tails right click on a new track
3. You are now in Line mode and can create a line unbound by head/tails

Probably related to the above or the Escape while in creation mode issue.  But if you create a polygon or head/tail and then immediately select another track it doesn't know what 'edit' mode it should be in.  None of the options are selected.

Small UI things:

Editing buttons should be disabled (grayed out, no cursor pointer) unless you have a track selected, are in editing mode, or are creating a new track.  Right now you can click on them and they don't do anything when nothing is selected.

Tooltips for every icon which doesn't have a label to indicate what it means.  I think that means recipes will have to have a setting for configuring that (haven't looked in depth at the code yet).

This is purely a small personal nitpick but this radius and the aliasing:  I don't think I noticed as much without a 1px border thickness:
image

BryonLewis:
Okay so the first issue regarding Line creation and Escaping during creation and then clicking elsewhere:

> Hitting Escape after switching to Head/Tails recipe and then clicking again produces an error: Related to calling hanleUpdateGeoJSON without a selected ID. Under further testing it looks like anytime a new feature is being created, or the user is in edit mode it will error out after hitting escape. This might have a relation to the editAnnotationLayer mode changing signal?

Is caused by the fact that the `editAnnotationLayer` `featureLayer` still exists even when we aren't in editing mode.  Which means it emits signals up to the `handleUpdateGeoJSON` when the mouse clicked even with no track selected.  A conditional for emitting the signal is being in creation mode, we are never kicked to editing mode because we escape and deselect before creating a full line so it remains in creation mode allowing it to emit a shape in progress signal.

  A remedy to this would be to modify the `this.mode` for the editAnnotationLayer to also have a `'disabled'` mode which prevents emitting signals that are independent from the data that is loaded.  On change data anyways we set the `this.mode` to `'creation' | 'editing'` anyways.

BryonLewis:
Second issue with consecutive Head/Tails Creations:

>There seems to be an issue where I can trick it into a regular line mode instead of head/tails.
Steps:
>1. Click an existing track and edit it to create a head/tails item
>2. Right after placing the tails right click on a new track
>3. You are now in Line mode and can create a line unbound by head/tails

So  this issue may have to do more with how we define recipes in relation to editing modes.  So the issue here is that headtails recipe is disabled as soon as it is finished creating the head/tails item.  The system is now in LineString editing mode with head/tails recipe disabled.  That is why the next time we create a line it starts as a LineString without the head/tails enabled.  The only way to enable head/tails mode is through clicking it manually.  So the main question is if we want the head/tails to propagate as an editing mode across selections of different tracks.  Rectangles and Polygons do (meaning once in polygon mode if you create a polygon and click on another track you remain in polygon mode)

This can be fixed by simply removing the deactivation of head/tails mode when you finally draw the line.  Although there might be a discussion about further implications of this and what may occur if we have both LineStrings and Head/Tails at the same time.

subdavis:
> Hitting Escape after switching to Head/Tails recipe and then clicking again produces an error: Related to calling hanleUpdateGeoJSON without a selected ID. Under further testing it looks like anytime a new feature is being created, or the user is in edit mode it will error out after hitting escape. This might have a relation to the editAnnotationLayer mode changing signal?
> A remedy to this would be to modify the this.mode for the editAnnotationLayer to also have a 'disabled' mode which prevents emitting signals that are independent from the data that is loaded. It is set into this mode when disabled() is called. 

I agree that seems like the right solution.

> So this issue may have to do more with how we define recipes in relation to editing modes. So the issue here is that headtails recipe is disabled as soon as it is finished creating the head/tails item. The system is now in LineString editing mode with head/tails recipe disabled. That is why the next time we create a line it starts as a LineString without the head/tails enabled. The only way to enable head/tails mode is through clicking it manually. So the main question is if we want the head/tails to propagate as an editing mode across selections of different tracks. Rectangles and Polygons do (meaning once in polygon mode if you create a polygon and click on another track you remain in polygon mode)

> This can be fixed by simply removing the deactivation of head/tails mode when you finally draw the line. Although there might be a discussion about further implications of this and what may occur if we have both LineStrings and Head/Tails at the same time.

I think we will need to add some visual distinction for the "primitive" type and "recipe" type.  For example, if a recipe switches you from line to polygon, you don't want to see the polygon button light up.  I think they need some kind of outlined or dashed state to indicate "this is the mode you're in, but the recipe is what's controlling it".  And if you click it, it goes from dashed to solid and the recipe gets disabled.

This is a problem for later.  For now, I agree that removing deactivation seems reasonable.

BryonLewis (approved):
This is working really good except for a few minor things that we should handle at only one level distance away from master instead of this two levels away.  Especially if we have to any master merge requests.

1. Pseudo Edit state replicated by clicking on a track for annotation and then left clicking inside of the current bounds.  The Editing UI for the annotation goes away but the rest of the UI still considers you to be in editing mode with nothing selected and no way to create an object.
2. I think the above is related to an issue in which you left click on a head/tails annotation to drag it, and then you right click while the left button is still down.  This creates a phantom line in the edit annotation layer and puts you into that psuedo editing state mentioned above.
3. Finally the implementation of Deletion for head/tails point should be done and possibly the ability to delete full annotations (polygon, head/tails)

BryonLewis:
Ability to create single point Head/Tail points be a process of creating a head/tail line, deleting the head, creating only a head and then hitting escape.

Ability to change the mode while not actually in editing mode.  You are able to click on the buttons to swap modes but there is no indication of the mode being changed until you activate editing again.
changingModeWhilenotEditing

The behavior of disabling the visibility of a type while editing the type has some interesting implications.  Mostly that the editing item still exists and that the editing mode isn't properly selected
polygonDisableIssue

You can delete points while in selected mode, not indicating that you have a selected point at the time.  It's done by clicking on a point and left clicking to toggle edit mode on and off.  It doesn't clear out the selected point and allows you to delete a point while not being in edit mode.
DeleteWhileinSelectedMode

I may add some other stuff.

subdavis:
> Ability to create single point Head/Tail points be a process of creating a head/tail line, deleting the head, creating only a head and then hitting escape.

You can place the first point and hit escape.  You don't need to create a line and delete.

> Ability to change the mode while not actually in editing mode. You are able to click on the buttons to swap modes but there is no indication of the mode being changed until you activate editing again

Things get weird if you try and fix this.  I'd have to change how recipe activation works in order to fix it because activation happens before mode switch, and the state to know about ignoring a button press only exists in useModeManager.  There's no way for the modemanager to signal to the recipe to ignore the button event right now.

> The behavior of disabling the visibility of a type while editing the type has some interesting implications. Mostly that the editing item still exists and that the editing mode isn't properly selected

Agreed, fixed


> You can delete points while in selected mode, not indicating that you have a selected point at the time. It's done by clicking on a point and left clicking to toggle edit mode on and off. It doesn't clear out the selected point and allows you to delete a point while not being in edit mode.

This happens on master, and it's an unsolved problem with geojs mode management.

BryonLewis:
Thanks for fixing/addressing those items.

> This happens on master, and it's an unsolved problem with geojs mode management that makes my head hurt. I'd like to fix this, but I don't know how.

I took a very quick shot at resolving this.  This branch is based of yours and has some minor changes.  I just tried to sync up that when you click on an annotation and kick it into the the internal geoJS `null` state it makes the editing mode disabled.  This makes it consistent that when you left click its "Selecting" and then right click is used as toggle between editing and not.  We could force it so that right click is always "Editing Mode" to maintain consistency.  This will then propagate that upwards to the controls to reflect it.  This also should fix the issue of clicking on the center item and having the deletion point toggle on and off.  There is probably some corner case that I'm missed in here. 

https://github.com/VIAME/VIAME-Web/compare/brandon/layersplitting-recipes...annotationSync

PR #330, opened 2020-09-17
Allow annotations > filenames to not error

subdavis:
If there are more annotations than filenames, don't throw a 500 out of range, just ignore it.

### DIFF
--- a/server/viame_server/serializers/viame.py
+++ b/server/viame_server/serializers/viame.py
@@ -198,7 +198,7 @@ def export_tracks_as_csv(
                         feature.fishLength or -1,
                     ]
 
-                    if filenames:
+                    if filenames and feature.frame < len(filenames):
                         columns[1] = filenames[feature.frame]
 
                     for pair in sorted_confidence_pairs:


BryonLewis (approved):
I guess if you upload the CSV for a larger dataset than what you uploaded this issue would occur.

Like if you took half of the JRS8 images files but used the full JRS8 .csv file.

AlmightyYakob:
Just to clarify, this only addresses removing images from the end of the dataset, correct? I believe removing images from the middle or beginning would end up skipping detections

subdavis:
@BryonLewis yep that's exactly what I did.  I figured someone else might do it too.

@AlmightyYakob also yes.  It addresses the other 2 cases by squashing the error, but the CSV export would be wrong, which would be user error.  But in those cases, there would be no 500 error produced.

If you think this should be a 400 error instead, I could understand that argument.

AlmightyYakob:
> @AlmightyYakob also yes. It addresses the other 2 cases by squashing the error, but the CSV export would be wrong, which would be user error. But in those cases, there would be no 500 error produced.

I guess I'm just looking for clarification of what's considered correct usage and what isn't. That is, we allow for there to be more annotations than imagery, but only if the annotations extend past the imagery, i.e. to exist only on frame numbers [0, inf). But we can't allow for gaps in that range, and we don't allow for a range like [3, inf).

AlmightyYakob (approved):
I'm okay with this, just want to make sure it's clear what's allowed and what isn't.

PR #337, opened 2020-09-19
Client/bug 320

### LINKED ISSUE #320: [BUG] "Track Bounds mis-aligned with feature-array"
**To Reproduce**
1. Create new detections in continuous mode.
2. Skip forward to a different frame.
3. Create a detection.
4. Move backward in time.

**Related data**
All datasets affected

**Expected behavior**
track.begin should represent the true beginning of the detection, rather than the frame the detection was initialized on.

subdavis:
fixes #320

### DIFF
--- a/client/src/track.spec.ts
+++ b/client/src/track.spec.ts
@@ -52,4 +52,25 @@ describe('Track', () => {
     expect(track.begin).toBe(1);
     expect(track.end).toBe(1);
   });
+  it('should correctly reset begin and end on firstFeature', () => {
+    const track = new Track(0, {
+      meta: {},
+      begin: 8,
+      end: 8,
+    });
+    const feature = {
+      frame: 3,
+      bounds: [1, 2, 3, 4] as RectBounds,
+      keyframe: true,
+    };
+    track.setFeature(feature);
+    const f2 = track.getFeature(2);
+    const f3 = track.getFeature(3);
+    const f4 = track.getFeature(4);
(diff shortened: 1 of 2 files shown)


BryonLewis (commented):
There is an additional corner case that can happen which causes the same error.

1. Go ahead in the image-sequence
2. Set To Detection with Continuous
3. Click on the "+Detection" button but don't actually draw a rect
4. Hit left to go back a frame
5. Draw the Detection
6. Then hit right to go forward and the error will occur

Technically the `begin` is correct, but the `end` is a frame ahead and has no actual detection associated with it.  So you have a track that's X length based on the bounds, but there is no final detection.  Need a way to shrink it.

I haven't dived in enough with the feature array setup and interpolation.  I'm wondering if you always want it so the first Feature inserted reset the begin/end to the current frame.
Within `setFeature` inside of `track.ts`
```
 if (feature.keyframe) {
      if (this.featureIndex.length === 0) {
        this.begin = feature.frame;
        this.end = feature.frame;
      }
      listInsert(this.featureIndex, feature.frame);
    }
```

subdavis:
@BryonLewis can you write a failing test?

subdavis:
@BryonLewis really dumb typo, but valuable test case 

https://github.com/VIAME/VIAME-Web/pull/337/commits/f0d69b4dea7b7017fc8d83cf0fc31a1115b0a5d7

Thanks

BryonLewis (approved):

PR #338, opened 2020-09-19
Fixes #297

### LINKED ISSUE #297: [BUG] reset password button doesn't do anything
Hook it up so that the button takes you to the reset interface in girder.   We will also need to set up the SMTP gateway.    I think there's a D&A mailgun account we can use.

subdavis:
fixes #297

### DIFF
--- a/client/app/views/Login.vue
+++ b/client/app/views/Login.vue
@@ -63,7 +63,10 @@ export default {
           or email <a href="mailto:viame-web@kitware.com">viame-web@kitware.com</a>
         </div>
       </v-alert>
-      <GirderAuth :register="true" />
+      <GirderAuth
+        register
+        forgot-password-url="/girder#?dialog=resetpassword"
+      />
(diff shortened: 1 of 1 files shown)


BryonLewis (approved):
I understand this is an immediate feature so I get pushing it up, but I think we may want to try to keep them inside the VIAME-Web app in the future.  I'm just worried about what eventual mess they can get themselves into by poking around their user folder within girder instead of using the VIAME app.

I imagine we could just create our own internal dialog that does a PUT request to `/user/password/temporary` like the girder dialog does with the email address they enter.

subdavis:
> I understand this is an immediate feature so I get pushing it up, but I think we may want to try to keep them inside the VIAME-Web app in the future. I'm just worried about what eventual mess they can get themselves into by poking around their user folder within girder instead of using the VIAME app.

That would be a feature request on girder/girder_web_components -- feel free to log it!

PR #339, opened 2020-09-19
Fix 329

### LINKED ISSUE #329: [BUG] You can't edit attributes for track with ID 0 but can for every track ID 1+
**Describe the bug**
See title

**To Reproduce**
Try to edit attributes for detection ID 0 in https://viame.kitware.com/#/viewer/5eb372788d01d968cc165ced

subdavis:
fixes #329


BryonLewis (approved):
One of our missed TrackID's relying on a truthy value.  Did another search to see if anywhere it might be doing the same.  Didn't see anything but I could of missed stuff.

PR #340, opened 2020-09-20
[BUG] fix feature.interpolate=true getting set for new detections, simplify logic. 

subdavis:
# Bug

If you created a detection while track.interpolate was true, the detection would become a "Track" with interpolate=true.

There was also unnecessary complexity with newTrackMode and newDetectionMode: only a single state variable + `newTrackSettings.mode` was necessary to know.

Now, you can even change from track to detection while you're in the middle of creation (between track create and bounding box draw).

I consider this a critical fix, because for the demo, if I'm trying to explain the difference between detection and track, it's confusing to have detections acting like tracks in an unintended way.

### DIFF
--- a/client/app/components/CreationMode.vue
+++ b/client/app/components/CreationMode.vue
@@ -88,7 +88,7 @@ export default Vue.extend({
           Mode:
         </v-col>
         <v-col>
-          <v-combobox
+          <v-select
             v-model="newTrackSettings.value.mode"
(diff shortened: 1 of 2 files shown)


BryonLewis (approved):
Pulled, tested the various modes and switching between them.  Also tested alternated first geometries (Line, Polygon) in the different modes and it all worked.

PR #341, opened 2020-09-21
Add sentry integration in prod

subdavis:
Add sentry integration in production.

Note that `docker-compose up` results in a production build, so any errors in a local docker compose stack will get pushed to sentry.  I don't really care that much, but it might be a problem later.

Also, if anyone tries to run this stack on their own, they'll ship their errors to our sentry collection.  That may or may not actually be preferable.  Punt for now.

I'd like @AlmightyYakob to review this because I think he has used sentry a lot more than me.

### DIFF
--- a/client/.env.production
+++ b/client/.env.production
@@ -1,1 +1,2 @@
 VUE_APP_STATIC_PATH=./static/viame/
+VUE_APP_SENTRY_DSN=https://0ec02775cc734df98134cecf5c91f782@o267860.ingest.sentry.io/5436001
--- a/client/app/main.ts
+++ b/client/app/main.ts
@@ -1,6 +1,8 @@
 import Vue from 'vue';
 import VueCompositionApi from '@vue/composition-api';
 import NotificationBus from '@girder/components/src/utils/notifications';
+import { init as SentryInit } from '@sentry/browser';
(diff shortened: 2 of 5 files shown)


subdavis:
https://stackoverflow.com/questions/45802988/typescript-use-correct-version-of-settimeout-node-vs-window

/shrug

AlmightyYakob:
> Note that `docker-compose up` results in a production build, so any errors in a local docker compose stack will get pushed to sentry. I don't really care that much, but it might be a problem later.

This could get pretty annoying if we actually care about addressing, although I think you have the option of ignoring errors raised from localhost. The way I've usually addressed this is by not keeping the DSN in source control, and only specifying it in the `.env` file in your particular deploy (i.e. in prod). That way, if you run it locally and don't specify the value (default), it won't initialize a sentry value.

subdavis:
> The way I've usually addressed this is by not keeping the DSN in source control, and only specifying it in the .env file in your particular deploy (i.e. in prod)

Eh, then we'd have to put the file on the regular host and the AWS host, and I also think I prefer to get errors reported from customer's deployments (there may be a consent issue here).

Let's just take everything for now and we can refine it later if it's a problem?

AlmightyYakob (approved):
Tested locally :+1:

PR #354, opened 2020-09-24
Move .env file to docker/.env

### LINKED ISSUE #344: [BUG] Docker pipeline volume isn't recreated during deployment.
**Describe the bug**
During deployment, volume mount isn't recreated.

**To Reproduce**
Deploy

AlmightyYakob:
Fixes #344 

The reason the volume was failing to be removed was due to a mismatch between the a `.env` file in the `docker` directory, and the source-controlled `.env` file in the root directory. The deploy runs under the `docker` directory, and so environment files pertaining to the docker stack should live in that directory. I believe it's best practice to keep these files at the same level as the compose file anyway.

The only reason I could see to keep this file in the root directory would be to run girder/other parts of the stack locally. However that's not a use case we currently support/endorse, so I don't see a reason to keep the `.env` file in the root directory.


subdavis (changes requested):
This breaks LDC, pull latest kitware/ldc for support.  FYI @BryonLewis 

In order to support `ldc` from both `/` and `/docker`, could you add `LDC_BASE_DIR=.` to the environment file?  No other change is necessary, I closed my other branch.

BryonLewis (approved):
pulled latest ldc, ran it from the root dir (`ldc dev girder`) and everything seemed to come up and function properly.

subdavis (approved):

PR #360, opened 2020-09-28
Update girder_worker to fix failing builds

AlmightyYakob:
This fixes the failing builds we've been seeing recently, caused by a bug I've documented [here](https://github.com/girder/girder_worker/issues/369). I've tested this locally and it seems to work without issue.

### DIFF
--- a/server/setup.py
+++ b/server/setup.py
@@ -8,7 +8,7 @@ requirements = [
     "girder==3.1.0",
     "diva-boiler",
     "girder_jobs==3.0.3",
-    "girder_worker==0.6.0",
+    "girder_worker==0.8.0",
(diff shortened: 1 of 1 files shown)


server/setup.py
```diff
@@ -12,6 +12,7 @@
     "girder_worker_utils==0.8.5",
     "pydantic",
     "pysnooper",
+    "celery==4.4.7",
```
zachmullen, line 15:
```suggestion
    "celery<5",
```

zachmullen:
FYI I just pushed `v0.8.0` tag on girder worker, so you shouldn't need this.

subdavis (approved):
rubber stamp

PR #367, opened 2020-09-30
Fix LDC

subdavis:
Moves part of the env file back.  Otherwise, this will break `ldc`.

Actually, this doesn't work either.  the env file won't get picked up if you move it.

### DIFF
--- a/docker/.env
+++ b/docker/.env
@@ -1,4 +1,3 @@
-COMPOSE_PROJECT_NAME=viame_web
 PUBLIC_DATA_PATH=./girder_data
 VIAME_PIPELINES_PATH=/opt/noaa/viame/configs/pipelines/
 GIRDER_ADMIN_USER=admin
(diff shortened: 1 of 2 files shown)


docker/.env
```diff
@@ -1,4 +1,3 @@
-COMPOSE_PROJECT_NAME=viame_web
```
AlmightyYakob, line 1: It's actually important that this line stay here, as otherwise the volume and service names will be wrong

subdavis: Right, I'll come up with something else.  Just a sec.

AlmightyYakob:
> Actually, this doesn't work either. the env file won't get picked up if you move it.

The point is that you'd now have to run your `docker-compose` commands from the `docker` directory. Why does `ldc` require that the `.env` file be a directory above the `docker` directory (where all the relevant docker files live)? What's the benefit to removing it from that directory?

subdavis:
Supported in kitware/ldc@6cfe55ec4781540eb4a07ce3cd70355e0f122a79

PR #369, opened 2020-10-01
Smoke test feature array in constructor

subdavis:
Better error handling for track contstructor.  If you need to construct a new track initialized with features, use `fromJson()`, not `new Track()`.   See the new test case for an explanation.

## Wrong

``` typescript
const features = [
      { frame: 1, bounds: [1, 2, 3, 4] as RectBounds, keyframe: true },
    ];
    new Track(1, {
      meta: {},
      begin: 1,
      end: 1,
      features: features,
      confidencePairs: [['foo', 1]],
    });
```

## Right

``` typescript
const itrack: TrackData = {
      attributes: {},
      begin: 0,
      end: 100,
      confidencePairs: [
        ['foo', 1],
        ['bar', 0.9],
      ],
      features: [
        {
          frame: 0,
          bounds: [1, 2, 3, 4],
          head: [1, 2],
        },
      ],
      meta: {},
      trackId: 1,
    };
    const t = Track.fromJSON(itrack);
```

### DIFF
--- a/client/src/track.spec.ts
+++ b/client/src/track.spec.ts
@@ -83,7 +83,7 @@ describe('Track', () => {
       meta: {},
       begin: 1,
       end: 1,
-      features: features,
+      features,
       confidencePairs: [['foo', 1]],
     })).toThrowError();
   });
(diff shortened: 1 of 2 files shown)


BryonLewis (approved):
Pulled, tested and looks good.  Also confirmed that error isn't thrown if you initialize a sparse array when using the constructor (essentially what is done in fromJSON): 
```
    const features = new Array(2);
    features[1] = { frame: 1, bounds: [1, 2, 3, 4] as RectBounds, keyframe: true };
```

PR #371, opened 2020-10-01
Remove use of alias inside library

### LINKED ISSUE #304: [BUG] Mismatch between local alias and published `vue-media-annotator`
**Describe the bug**
When importing some components from the published `vue-media-annotator` npm package, they attempt to reference other components in the package. However, they're not actually importing from the published package, but from the alias defined in `jsconfig.json`. This alias resolves to `src/`, which means that in Viame-Web, the following example import succeeds no problem:

https://github.com/VIAME/VIAME-Web/blob/7cc58fe54b4398fa3dc45b0b396ec4a6a81350dd/client/src/components/LayerManager.vue#L12

Doing this import when using the published package, this import path is taken literally, which means it looks for `components` in the root package directory. This import fails, as the proper path is `src/components`.

**To Reproduce**
1. In an external application, add the `vue-media-annotator` package to your project.
2. Attempt the following import
3. There should be errors in the console when building/serving.

**Expected behavior**
The import succeeds.

---

I think ideally the publish package would be rooted at the `src` directory, as I don't think anything above that directory is needed? However I'm not sure if configuring the npm package that was is feasible/possible.

subdavis:
Removes `vue-media-annotator` alias from the library portion of the app. Keep using it in the app part.

Blocked on #361 
fixes #304

### DIFF
--- a/client/app/components/DeleteControls.vue
+++ b/client/app/components/DeleteControls.vue
@@ -1,6 +1,6 @@
 <script lang="ts">
 import Vue, { PropType } from 'vue';
-import { EditAnnotationTypes } from 'vue-media-annotator/layers/EditAnnotationLayer';
+import { EditAnnotationTypes } from 'vue-media-annotator/layers/';
 
 export default Vue.extend({
   name: 'DeleteControls',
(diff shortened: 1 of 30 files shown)


AlmightyYakob (approved):
LGTM, tested locally.

AlmightyYakob (approved):

PR #378, opened 2020-10-09
Add rollup for module release

subdavis:
Finally some progress toward an esm distributable of `vue-media-annotator`.

Apparently if the root import of your `.vue` files is a `.js` file, rollup-plugin-vue just builds the vue SFCs as if they were normal javascript.  You can't get type definitions, but you can at least build.

I've broken the library into 2 import roots: 1 for typescript with defs, 1 for components without.  Here's how you use it:

``` js
import { VideoAnnotator, LayerManager } from 'vue-media-annotator/dist/components';
import { providers, use } from 'vue-media-annotator';
```

I've confirmed with a consumer app that this works out-of-the-box in a regular vue app.


subdavis:
Released `1.0.0-beta.1` for testing.

client/src/index.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export {
+  listUtils,
+  providers,
+  Track,
+  utils,
+
+  use,
+  layers,
+  // components,
```
subdavis, line 18:
```suggestion
```

client/src/index.ts
```diff
@@ -0,0 +1,19 @@
+import * as listUtils from './listUtils';
+import * as providers from './provides';
+import Track from './track';
+import * as utils from './utils';
+
+import * as use from './use';
+import * as layers from './layers';
+// import * as components from './components';
```
subdavis, line 8:
```suggestion
```

client/src/components/TrackItem.vue
```diff
@@ -321,7 +321,7 @@ export default defineComponent({
     }
   }
   .input-box {
-    border: 1px solid rgb(255, 255, 255, 0.15);
+    border: 1px solid rgba(255, 255, 255, 0.15);
```
AlmightyYakob, line 324: Seems unrelated?

client/app/compositionplugin.ts
```diff
@@ -1,6 +0,0 @@
-import Vue from 'vue';
-import VueCompositionApi from '@vue/composition-api';
-
-Vue.use(VueCompositionApi);
-
-export default VueCompositionApi;
```
AlmightyYakob, line 6: Why was this file removed?

client/rollup.config.js
```diff
@@ -0,0 +1,72 @@
...
+const clean = true;
+const tsconfig = 'tsconfig.rollup.json';
+const external = [
+  '@vue/composition-api',
+  '@flatten-js/interval-tree',
+  'd3',
+  'geojs',
+  'lodash',
+  'vue',
+];
```
AlmightyYakob, line 18: Is this ultimately specified to reduce bundle size?

AlmightyYakob (commented):
Just a few comments, but generally looks good to me.

client/src/components/TrackItem.vue
subdavis: It was breaking build because the rollup sass plugin validates sass syntax.  Nothing in the webpack build was doing that.

client/app/compositionplugin.ts
subdavis: It was never being imported.

client/rollup.config.js
subdavis: It does that, but only because there's just no other option.  Otherwise you run into the issue where multiple instances of Vue (and every other dependency) are instantiated and nothing works.

The point is to defer dependency resolution to the consumer.

AlmightyYakob:
Of note, if you're not running the same (at least major?) version of `@vue/composition-api` as this is (`1.0.0-beta.14`), you get composition api errors down the call stack (I got mine in `useTrackStore` for example). Besides that, it seems to work fine. I have to update my app to use `provideAnnotator`, but that seems separate from this.

AlmightyYakob (approved):

PR #381, opened 2020-10-09
Fix process hanging when training with SVM config

AlmightyYakob:
This was due to an error in our use of `subprocess`, and the fact that using `subprocess.PIPE` to capture `stdout`/`stderr` can lead to hanging processes if there's enough output. [This](https://thraxil.org/users/anders/posts/2008/03/13/Subprocess-Hanging-PIPE-is-your-enemy/) article describes the problem well. This changes all of our calls to `Popen` to instead use temporary files to capture this output. I may try to refine this to make it less work on the caller of `read_and_close_process_outputs`, but I think it would be fine to merge as is.

I've tested this on a server with a GPU to confirm it stops the SVM training from hanging.

### DIFF
--- a/server/viame_tasks/tasks.py
+++ b/server/viame_tasks/tasks.py
@@ -328,7 +328,7 @@ def convert_video(self, path, folderId, auxiliaryFolderId):
         process, process_log_file, process_err_file
     )
 
-    output = str(stdout) + "\n" + str(stderr)
+    output = stdout + "\n" + stderr
     self.job_manager.write(output)
     new_file = self.girder_client.uploadFileToFolder(folderId, output_path)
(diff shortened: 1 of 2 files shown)


BryonLewis (approved):
Was able to reproduce the issue in the root `add-training` branch of a hanging training run.  It would run and stop and the job would report running while no process was doing any compute when I looked at nv-top.  This branch will successfully run and complete without any hanging.  I got back a pipeline but wasn't able to run it on anything.  This is further than I could get with the base branch.

Went through and tested the other jobs that could be done:
Ran a default pipeline on data and it completed properly.
Tested AVI to Mp4 conversion task.
Tested Tiff to Png conversions as well.

This seems to solve the issue and as far as I can tell has no adverse side effects outside of whatever else could be occurring in pipelines/training.

PR #385, opened 2020-10-16
Remove lingering running/cancelling jobs on girder start

AlmightyYakob:
~~Currently, these jobs are simply deleted when the server is spun up. Ideally, these jobs would be restarted, but so far attempts to do so result in new jobs remaining in an inactive state. I'd like to complete this task fully, but for now I think it's good to have this in here.~~ I've also added a warning to the client, so that users know that their training job could be killed if it's still running when the server is updated.

I've taken the advice of @subdavis, now if the server is restarted, existing running tasks will be restarted.

### DIFF
--- a/client/app/components/RunTrainingMenu.vue
+++ b/client/app/components/RunTrainingMenu.vue
@@ -75,7 +75,7 @@ export default {
 <template>
   <v-menu
     v-model="menuOpen"
-    max-width="320"
+    max-width="500"
     offset-y
     :close-on-content-click="false"
   >
@@ -105,7 +105,7 @@ export default {
 
     <template>
       <v-card v-if="trainingConfigurations">
-        <v-card-title>
+        <v-card-title class="pb-1">
           Run Training
         </v-card-title>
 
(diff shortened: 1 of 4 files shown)


subdavis:
I don't know if this is a reasonable way to handle user data.  Could you talk a bit more about what you've tried?  

Have you looked at acks_late? https://docs.celeryproject.org/en/latest/faq.html#should-i-use-retry-or-acks-late

I believe retry is useless in the case where the worker gets killed, but acks_late should cause it to get picked up again on restart.  I've never done this, but I've done similar stuff with other queue processors.

AlmightyYakob:
There's a case here that could arise, which is that if someone's job takes longer than a week to run, it would always be restarted, and never complete. I tried to set up a check at plugin load to fix this, but I was running into issues, so I've left it for now. If it's something that starts causing issues, we can always come back to it.

subdavis:
Did acks late solve the restart problem?

AlmightyYakob:
> Did acks late solve the restart problem?

Yes, but it (sort of) introduced a new problem, which is explained and fixed in 1af10a5

PR #395, opened 2020-10-21
Fixes 357

### LINKED ISSUE #357: [BUG] Do not resize bounding box when drawing a head tail loc for detection with existing box

subdavis:
fix #357 (see comments)

### DIFF
--- a/client/viame-web-common/recipes/headtail.ts
+++ b/client/viame-web-common/recipes/headtail.ts
@@ -11,6 +11,23 @@ export const HeadPointKey = 'head';
 export const TailPointKey = 'tail';
 const EmptyResponse: UpdateResponse = { data: {}, union: [], unionWithoutBounds: [] };
 
+/* Standard 10% padding */
+// const PaddingVector10Percent: [number, number][] = [
+//   [-0.1, -0.1],
+//   [-0.1, 0.1],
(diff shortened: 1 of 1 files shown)


client/viame-web-common/recipes/headtail.ts
```diff
@@ -11,6 +11,23 @@ export const HeadPointKey = 'head';
 export const TailPointKey = 'tail';
 const EmptyResponse: UpdateResponse = { data: {}, union: [], unionWithoutBounds: [] };
 
+/* Standard 10% padding */
+// const PaddingVector10Percent: [number, number][] = [
+//   [-0.1, -0.1],
+//   [-0.1, 0.1],
+//   [1.1, -0.1],
+//   [1.1, 0.1],
+//   [-0.1, -0.1],
```
subdavis, line 20: I want to keep this as an explanation of how to do nontrivial padding.

BryonLewis (approved):
Looks good to me.  I think eventually if we have to go back and use this a little bit more explanation about geoJS -> geoJSON and it's need for an extra coordinate might be needed.  I keep seeing 5 coordinates referenced and it was confusing me because we are talking about the bounds (And this is something I know about already).

PR #406, opened 2020-10-27
Fixes #404

### LINKED ISSUE #404: [BUG] [Regression] track list types disappear on keyboard navigation
**Describe the bug**

This happens:

Screenshot from 2020-10-26 15-56-31

shift-enter to edit is broken.  Escape to exit is broken.  enter to save is broken.

**To Reproduce**
1. Open any dataset
2. Hit the down arrow

Nothing useful in console

subdavis:
fixes #404

### DIFF
--- a/client/src/components/TrackItem.vue
+++ b/client/src/components/TrackItem.vue
@@ -97,7 +97,7 @@ export default defineComponent({
       }
     }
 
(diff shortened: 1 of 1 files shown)


BryonLewis (approved):
Pulled and tested, escape, down and enter as well as shift enter now works for setting the type.

Issue #410, opened 2020-10-28
[FEATURE] [Testing] Create unit tests for critical vue components


subdavis:
Add shallow-mount unit tests with vue-test-utils for...

# Highest priority

* TrackList
* TrackItem
* TypeList

# Next priority 

* LayerManager
* VideoAnnotator
* ImageAnnotator

PR #411, opened 2020-10-28
Fix toggleKeyframe regression introduced in #361

subdavis:
Oops:

https://github.com/VIAME/VIAME-Web/commit/80483a9f7ff0b1f3f5c58bb0b44bbcc9700f85e7
https://github.com/VIAME/VIAME-Web/blob/master/client/src/components/TrackItem.vue#L130L141

Demonstrated need for #410


BryonLewis (approved):
Pulled, tested and appears fixed.

PR #425, opened 2020-11-04
Add checkbuild.sh

subdavis:
I get a little impatient waiting for CI to run 10 minutes worth of checks, so this script will perform the same steps as CI in parallel.  It takes roughly 90 seconds on my machine.

That way you can be confident that your commit is functional before you push and waste less time fixing minor CI errors.

I don't want to add this as a commit hook.  It takes too long, and often I need to create a transient broken commit.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -110,7 +110,11 @@
       "ts-jest": {
         "tsConfig": "tsconfig.spec.json"
       }
-    }
+    },
+    "modulePathIgnorePatterns": [
+      "<rootDir>/dist/",
+      "<rootDir>/dist_electron/"
+    ]
   },
   "eslintConfig": {
     "root": true,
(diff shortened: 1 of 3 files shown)


client/package.json
```diff
@@ -110,7 +110,11 @@
       "ts-jest": {
         "tsConfig": "tsconfig.spec.json"
       }
-    }
+    },
+    "modulePathIgnorePatterns": [
+      "<rootDir>/dist/",
+      "<rootDir>/dist_electron/"
+    ]
```
subdavis, line 117: On linux, `build:electron` creates a `.snap` package in `dist_electron`, which is the same file extension of a jest snapshot.  The snapshot doesn't match and so tests fail.

AlmightyYakob (approved):
Ran locally no problem :+1:

PR #433, opened 2020-11-10
[Bugfix] inverse interpolation

subdavis:
I really don't know how we never noticed this.

Who wrote this stupid code anyway.


AlmightyYakob:
What behavioral bug is this fixing? And how would I test that this behaves correctly?

subdavis:
When you download a CSV, interpolation was being run backwards

That is, in `A[] -> B[]`, you were getting `A1, B.9*A.1, B.8*A.2, ..... B1`

BryonLewis (approved):

PR #435, opened 2020-11-11
Allow setting of arbitrary attributes on video player

subdavis:
# Example

`<video-annotator :video-player-attributes="{ muted: true, loop: true }" />`

You can absolutely footgun with this.  But I wanted to expose all possible configuration rather than having explicit props for everything.

I put it below `src, preload` to allow hacking on those, but above `onloadMetadata` because that would be unreasonable to try to hack.

### DIFF
--- a/client/src/components/annotators/VideoAnnotator.vue
+++ b/client/src/components/annotators/VideoAnnotator.vue
@@ -12,6 +12,10 @@ export default Vue.extend({
       type: String,
       required: true,
     },
+    muted: {
+      type: Boolean,
+      default: false,
+    },
   },
 
   computed: {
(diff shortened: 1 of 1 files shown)


BryonLewis (approved):
I understand the problem with shooting yourself in the foot with this.

Is this intended for vue-media-annotator instead of VIAME-Web in general.  Is it solely to get around the fact that we don't have audio controls or are there other features you want added in (looping)?
 I.E:  Looping, if specified in shouldn't we make the left/right arrows do the same looping action then?

I don't mind this is as a quick fix to solve specific needs.  I think if it is used too often we may want to think about specifically supporting audio controls as well as looping behavior (in terms of key controls)

subdavis:
This is a convenience for library consumers.

> I don't mind this is as a quick fix to solve specific needs. I think if it is used too often we may want to think about specifically supporting audio controls as well as looping behavior (in terms of key controls)

I don't know if I'd call it a quick fix.  There are plenty of attributes, like `muted`, `loop`, `poster`, etc that a lib consumer might want.  But 100% agree that if people start doing weird stuff, we should take that as a sign that something should change.

I don't intend to add support in VIAME-Web client.

PR #438, opened 2020-11-12
Add dummy generators

subdavis:
**This is a breaking change, and will upgrade to `1.1.0`.**

* Change arguments to `provideAnnotator` to accept an object, because the positional args were getting out of hand.
* Add methods to generate dummy state and a dummy handler compliant with the interface so that they can be easily overridden by downstream apps with much simpler needs.

## Example

``` typescript
provideAnnotator(
  { ...dummyState(), tracks: myTracks },
  dummyHandler(() => {}),
);
```

@AlmightyYakob and I have talked about this change

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vue-media-annotator",
-  "version": "1.0.3",
+  "version": "1.1.0",
   "scripts": {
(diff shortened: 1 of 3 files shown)


BryonLewis (approved):
Pulled tested, looks like a good improvement for users of the library.

PR #439, opened 2020-11-12
Downgrade urllib3

subdavis:


BryonLewis (approved):
for the sake of timeliness I'm going to rely on the checks.  Did a build with `1.25.11` so I would hope that this functions the same.

subdavis:
FWIW, none of the code paths in the checks are going to exercise anything related to urllib3.  But I can confirm that this gives you 1.25.11.

PR #453, opened 2020-11-18
Upgrade GWC 2.2.8

subdavis:
Fixes uploader performance concurrency bug.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -27,7 +27,7 @@
   "types": "lib/types/index.d.ts",
   "dependencies": {
     "@flatten-js/interval-tree": "^1.0.11",
-    "@girder/components": "^2.2.4",
+    "@girder/components": "^2.2.8",
     "@mdi/font": "^3.9.97",
     "@sentry/browser": "^5.24.2",
     "@sentry/integrations": "^5.24.2",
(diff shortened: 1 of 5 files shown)


client/package.json
```diff
@@ -44,7 +44,7 @@
     "vue": "^2.6.12",
     "vue-gtag": "^1.9.1",
     "vue-router": "^3.0.3",
-    "vuetify": "^v2.3.10",
+    "vuetify": "^2.3.10",
```
subdavis, line 47: There shouldn't have been a `v` here.

BryonLewis (changes requested):
Looks like vuetify swapped over to only a default export for 'Vuetify' where it used to have both the default and named (Vuetify) exports.  So it looks like that may need to be fixed in `useStyles.ts` to prevent the build error.

Other than that I pulled and ran it and tried to look at as much vuetify stuff as I could and nothing jumped out to me as broken or not working.  Looked specifically at everywhere we use inputs/icons and others.

BryonLewis (approved):
Tested again and everything seems to work well.  I didn't know about the type only import, that seems like it could be useful in the future.

PR #465, opened 2020-12-01
change getTrack

subdavis:
I didn't mean for you to add `getTrack` to the provided interface.  I don't know if you did that because you wanted to or you thought it's what I wanted.

* This would be a departure from the style of provided state.  In your change, getTrack is a partial function application where trackMap is already in the closure.  We don't have any other state like that, so it might be good to do it this way just for consistency.  
* More importantly, though, the point of `getTrack` is to to throw an error if the requested track is missing, which is useful in the common case where missing track would need to be a runtime error (i.e. selectedTrackId should always be in TrackMap).  This is how it worked in your change, but the type of `GetTrackType` included `undefined` which somewhat defeated the purpose.
* If you want a getter that just returns undefined if missing, you don't need a new function.  `trackMap.get(key)` already has this method signature.

### DIFF
--- a/client/src/provides.ts
+++ b/client/src/provides.ts
@@ -44,9 +44,6 @@ type TrackMapType = Readonly<Map<TrackId, Track>>;
 const TracksSymbol = Symbol('tracks');
 type TracksType = Readonly<Ref<readonly Track[]>>;
 
-const GetTrackSymbol = Symbol('GetTrack');
-type GetTrackType = (trackId: number) => Track | undefined;
-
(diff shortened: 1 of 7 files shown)


BryonLewis (approved):
Sorry that was my confusion.  I also assumed that everywhere we use getTrack we included the trackMap anyways.  So I think I was trying to rid ourselves of two ways of accessing the track data and it was probably a mistake including the undefined during initial testing to solve some TS errors (or it could of been something with where one specific need for it is to returned undefined).

PR #476, opened 2020-12-04
Small fix

subdavis:
Removed the check for `target` because I couldn't see a valid execution path where that would ever be false.

If it's undefined or null, that should be a runtime error, right?


BryonLewis (approved):
I could see removing it as long as we know that all of our change events are tied directly to HTMLInputElements or at least something that has a target and a string based value.

PR #506, opened 2020-12-16
Lazy load pipelines

subdavis:
I'm testing this locally while CI runs.


subdavis:
Works for me locally, but that doesn't count for much.

BryonLewis (approved):
Giving the approval, better than no pipelines working.

PR #511, opened 2020-12-23
Add current frame image filename to display for image-sequence

### LINKED ISSUE #300: [FEATURE] Display filename alongside imagery for image sequences

subdavis:
This looks like a lot of changes, but it's really just a couple indentation changes.  I removed the `v-toolbar` and `v-container` wrappers in the toolbar because they provided no value, replaced with a fluid-height `v-card` so that text wrapping looked correct at narrow widths.

Had to move controls into the `timeline-wrapper` so that `frame` prop from annotator slot could be used.

Got rid of `annotatorType` derived property because we already have an explicit `meta.type` associated with the dataset.  

Screenshot from 2020-12-23 09-35-53

# Location

I put it in the time control bar because that's where it made most sense to me.  It's tightly coupled to time/frame controls, so having it in the header seemed like a violation of locality.  Also, I haven't done this yet, but on videos the filename text would be replaced with a timestamp, and that's objectively the correct place to put a timestamp.

resolves #300

### DIFF
--- a/client/src/components/controls/Controls.vue
+++ b/client/src/components/controls/Controls.vue
@@ -60,90 +60,81 @@ export default defineComponent({
       { bind: 'd', handler: mediaController.prevFrame, disabled: $prompt.visible() },
     ]"
   >
-    <v-toolbar
-      height="80px"
-      flat
-    >
-      <v-container
-        fluid
-        class="pa-0"
-      >
-        <v-row>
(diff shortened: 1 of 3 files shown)


BryonLewis (approved):
Pulled and it works well.  A little different having the offset controls, but it's worth it to add the image names in there.

PR #514, opened 2020-12-24
Only source pipeline volume from one place

subdavis:
Because the volume was being populated by 3 containers, there's a race condition for whichever loads first, and others will try to "copy" their data over to the host mount and fail.

Only expose the volume to girder from a single worker.   It doesn't matter which one, since they all have the pipelines directory internally.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -8,9 +8,6 @@ x-worker: &base-worker
     context: ../
     dockerfile: docker/girder_worker.Dockerfile
   image: kitware/viame-worker:${TAG:-latest}
-  volumes:
-    # Exposes pipeline directory to girder so it can see what pipelines are available.
-    - pipelines:${VIAME_PIPELINES_PATH}:ro
   depends_on:
     girder:
         condition: service_healthy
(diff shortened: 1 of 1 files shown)


BryonLewis (approved):
Works well, was able to duplicate the error I saw on the server and the new version prevents it from occurring.  Just including the below for searching/future reference if needed.

Testing by bringing down the containers, removing them and the pipeline volume I could get the  error attached at the bottom.  Not every time, but it could happen.  Restarting the containers would fix it.  During testing it would hang on training just like the prod server did and I could see that the training girder worker failed to start successfully.

Using the updated compose file no error occurs through multiple down/up/clean cycles.
```
Creating viame_web_girder_worker_training_1  ... error

ERROR: for viame_web_girder_worker_training_1  Cannot create container for serviCreating viame_web_girder_worker_default_1   ... error
elines/_data/embedded_single_stream: mkdir /var/lib/docker/volumes/viame_web_pipelines/_data/embedded_single_stream: file exists

ERROR: for viame_web_girder_worker_default_1  Cannot create container for service girder_worker_default: failed to mkdir /var/lib/docker/volumes/viame_web_pipelCreating viame_web_girder_worker_pipelines_1 ... done
s: file exists

ERROR: for girder_worker_training  Cannot create container for service girder_worker_training: failed to mkdir /var/lib/docker/volumes/viame_web_pipelines/_data/embedded_single_stream: mkdir /var/lib/docker/volumes/viame_web_pipelines/_data/embedded_single_stream: file exists

ERROR: for girder_worker_default  Cannot create container for service girder_worker_default: failed to mkdir /var/lib/docker/volumes/viame_web_pipelines/_data/models: mkdir /var/lib/docker/volumes/viame_web_pipelines/_data/models: file exists
```

PR #516, opened 2020-12-30
Schema/refresh phase2

subdavis:
## Phase 2: what really happened.

Phase 2 went differently than I expected.  I doubled back on some choices from Phase 1, and it turned out less involved than I originally thought, so rather than try to make a bunch of changes in 1, then roll them back in 2, I'd prefer to just review 2 as if it were the only change.  This will close #515 

* Revert to `loadDetections` / `loadMetadata` instead of `loadDataset`.  There was no need to remove the decoupling between these methods.  I was wrong that detections requires metadata: it only actually requires knowlege of the base data path.  Extra reasoning is explaining inline in a comment in `desktop/frontend/store/index.ts`
* Reorganize the folder structure a bit with `desktop/frontend` and `desktop/backend` for clarity.
* Explain the 1 exception: keep loading detections directly from the renderer thread because it's a waste of compute and memory to buffer the whole file into memory _twice_ (in the server thread, then in the client thread) to load it.
* Keep recents and settings in localStorage.  When I originally wanted to move them, I forgot the chicken-and-egg problem: you have to know where the file is on disk before you can load it.  LocalStorage provides everything I was going to go to the trouble of implementing myself in the server (levelDB at a fixed location on disk) so it seems foolish to write code to end up with what you've already been given for free.  
  * This leaves the minor annoyance of the backend having to ask the frontend for settings, but that's what IPC is for. It's not really as bad as I was making it out to be.  
* MIgrate to express.js webserver, add new endpoints for metadata save, metadata load, and import.
  * Media import could have been done with REST or IPC.  Doesn't seem to matter to me which is used.
* Changed the signature of `SaveDetectionArgs` since the object provided no benefit over a simpler array.  This involves a simple server-side change.

## Not included

Persistence for job run history (logs will not be populated if a job is loaded from this storage area)
Error handling for importMedia issues. Right now they just dump to console and nothing happens in the UI.
Error handling for when source files change location.

In this sense, this is more of a phase 1.5.  

## Some notes on the size of the changes

The size of this PR is misleading.  I renamed some files (most notably `frontend/api.ts`) and reordered the contents.  Many other files have been renamed, and git is considering them 100% changed for some reason. 

It's probably closer to 850+, 500-


client/platform/desktop/backend/server.ts
```diff
@@ -100,18 +139,33 @@ router.register('/api/media', withErrorHandler((req, res) => {
...
-}));
-
-// We need a server which relies on our router
-const server = http.createServer((req, res) => {
-  const handler = router.route(req);
-  handler.process(req, res);
 });
 
-export default server;
+app.use(cors({ origin: 'http://localhost:8080' }));
```
BryonLewis, line 152: very minor and only applies to `yarn serve:electron` - this locks you into CORS only being accepted from localhost:8080, so if you have other things running on that port it will fail to give to access because vue-cli-server will switch to 8081 or others.  Like I said minor, but just something to note.

client/platform/desktop/backend/ipcService.ts
```diff
@@ -1,57 +1,64 @@
...
   ipcMain.handle('open-link-in-browser', (_, url: string) => {
     common.openLink(url);
   });
+  ipcMain.on('update-settings', async (_, s: Settings) => {
+    settings.set(s);
+  });
+  ipcMain.handle('import-media', async (_, path: string) => {
+    const ret = await common.importMedia(settings.get(), path);
+    return ret;
+  });
```
BryonLewis, line 40: This just here for possible future uses?  It is never called now that we are using the REST version.

client/platform/desktop/backend/native/common.ts
```diff
@@ -0,0 +1,473 @@
...
+    type: datasetType,
+    id: dsId,
+    fps: 5, // TODO
+    originalBasePath: path,
+    originalVideoFile: '',
+    originalImageFiles: [],
+    transcodedVideoFile: '', // TODO: this is empty (see above)
+    transcodedImageFiles: [], // TODO: this is empty
+    name: dsName,
+  };
```
BryonLewis, line 382: This may be a niche case, but we may want to add either a timestamp or a user custom Name to the JSONMeta.  Feel free to dismiss or mark it for another PR - During testing I opened the same dataset multiple times, it properly creates the different unique directories and everything, but then I can't tell the difference between them and some have pipelines run on them, some have other input data that I used.  Just wondering if in metadata we should have either a timestamp or something else to let the user know more information about dataset or a user configurable name?
image

client/platform/desktop/backend/native/common.ts
```diff
@@ -0,0 +1,473 @@
...
+      npath.join(projectDirAbsPath, npath.basename(trackFileAbsPath)),
+    );
+    foundDetections = true;
+  }
+  /* Look for other types of annotation files as a second priority */
+  if (!foundDetections) {
+    const csvFileCandidates = contents
+      .filter((v) => CsvFileName.test(v))
+      .map((filename) => npath.join(jsonMeta.originalBasePath, filename));
+    const { fps, processedFiles } = await processOtherAnnotationFiles(
```
BryonLewis, line 442: very minor, in the JSON loading you warn the user if they have mutiple .json track files that are being loaded.  Here it will load multiple CSVs and only take one of them.  Reminder the processAnnotationFiles can take multiple CSVs because tracking pipelines will produce two csvs, a detector.csv and a tracking.csv.

client/platform/desktop/backend/native/common.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+
+  it('importMedia video success', async () => {
+    const meta = await common.importMedia(settings, '/home/user/data/videoSuccess/video1.mp4');
+    expect(meta.name).toBe('video1');
+    expect(meta.originalImageFiles.length).toBe(0);
+    expect(meta.originalVideoFile).toBe('video1.mp4');
+    expect(meta.originalBasePath).toBe('/home/user/data/videoSuccess');
+  });
+
+  it('importMedia various failture modes', async () => {
```
BryonLewis, line 237: minor `failture` -> `failture`

client/platform/desktop/frontend/store/index.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  /**
+   * loadDetections loads JSON data directly from disk into the
+   * renderer thread. It relies on the node runtime being enabled on the browser.
+   *
+   * This is done such that large annotation files do not need to be loaded into memory
+   * twice, serialized and deserialized twice, and transmitted over the local network.
+   *
+   * In a future version, this could me moved to the backend and streamed directly
+   * to the client using something like https://github.com/uhop/stream-json
```
BryonLewis, line 42: interesting side-note, exact package I was looking at when wondering if there should be streaming for large json file conversions to CSV, but decided against it because all JSON files need to be fully loaded in memory anyways.

BryonLewis (changes requested):
Big PR and seems fully functional on Linux at least.  Went through most of the standards of testing the loading, swapping, running pipelines and others for the electron version.  I also tested to make sure the changes to loading meta/detections didn't have any adverse effect on the server.  All seems stable.  I didn't worry about windows testing as of this moment.

Most of what I have are just my standard annoying questions and wondering where stuff is heading towards.  I do think we shouldn't do the moving of track.json in source media folder and instead do a copy.

client/platform/desktop/backend/native/common.ts
```diff
@@ -0,0 +1,473 @@
...
+
+  let foundDetections = false;
+
+  /* Look for JSON track file as first priority */
+  const trackFileAbsPath = await _findJsonTrackFile(jsonMeta.originalBasePath);
+  if (trackFileAbsPath !== null) {
+    /* Move the track file into the new project directory */
+    await fs.move(
+      trackFileAbsPath,
+      npath.join(projectDirAbsPath, npath.basename(trackFileAbsPath)),
```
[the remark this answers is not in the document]
subdavis: Agree.

client/platform/desktop/backend/server.ts
subdavis: No, this is important.  It would be better to find a way to discover the devserver address and pull from that.

client/platform/desktop/backend/ipcService.ts
subdavis: Duplicate, should be removed.

client/platform/desktop/backend/native/common.ts
subdavis: Good suggestion, we should maintain a `createdAt` property in metadata.  I don't know if we should show it in recents because it may look messy.  But it would be good to have.

client/platform/desktop/backend/native/common.ts
subdavis: I thought about this when I was doing it, but IMO the current behavior is correct.

1) in normal operation, two result_.json files should never be created.  We wouldn't know which to load.
2) Likewise, multiple .csv files _can_ be created.  We still don't know which to load, so we allow the caller to specify a priority order because creation of these files is out of our control.

I don't know of a better strategy.  What would you suggest.

client/platform/desktop/frontend/store/index.ts
subdavis: I know, I got this from you.

I put the comment in just as a relevant bookmark since github issue / PR search is noisy. Can remove if you want.

client/platform/desktop/backend/native/common.ts
BryonLewis: I am talking about the caller being the media Importer here and that the mediaImporter should specify what occurs instead of leaving it up to filename order.  I wouldn't change how `processOtherAnnotationFiles`, but maybe here we handle in the return that there are multiple `processedFiles` and we either notify the user or error like we do with the JSON loader.  The specifics were importing an image sequence which had multiple csv files in the folder then it takes one of those CSV files (I think by filename order) and declares it as the trackJSON.  I was just including some background as to why `processOtherAnnotationFiles` will handle multiple csv files (mostly for my own recollection).

client/platform/desktop/frontend/store/index.ts
BryonLewis: yeah leave it in as a reminder.  I'm assuming that at some point we are going to be hitting track files that may be 10's of MBs and transferring them all at once and waiting to load may not be feasible.

client/platform/desktop/backend/native/common.ts
subdavis: Only added to metadata, didn't render it in the resents list.  It looks messy and wraps badly.

BryonLewis (approved):

PR #534, opened 2021-01-19
Disable unit testing on windows for now

subdavis:
Allows release to succeed.  Can consider a more suitable fix later.


BryonLewis (approved):
We can create an issue  and assign it to me with getting the tests to pass on windows.  I'm assuming using some os statements for the tests would work.  Also looks like there are some npm packages for conditional testing among different operating systems.

subdavis:
It's just path matching.

The solution is as simple as `(\/|\\)` for matching slashes, I think.  But feel free to explore.

PR #565, opened 2021-02-05
Add env to configure default viame path

### LINKED ISSUE #560: [FEATURE] Need way to pass in VIAME installation dir on command line to launcher in Desktop
for the .zip installers of VIAME (including DIVE), we already know where VIAME is so the user shouldn't need to set the path, alternatively if there's a way to hard-code it somewhere in source and I can custom build would also work

subdavis:
Setting `DIVE_VIAME_INSTALL_PATH` on all platforms will override the platform default.

If a user sets the value themselves on the settings page, the environment var will be ignored.

fixes #560


BryonLewis (commented):
haven't pulled but won't this reset itself on each launch?  Meaning you're locked into your env variable location and have to reset it each time you launch the app if you want to use a different version or something else.  Should changing the install location reset the env variable?

subdavis:
> haven't pulled but won't this reset itself on each launch?

I don't think so, but I'm not sure what you're asking.  If a user sets their install path through the GUI, the default, and therefor the environment variable, will be totally ignored.

Basically, the env var works up until the point where the user hits the settings save button, then it's ignored forever.

BryonLewis (approved):

PR #567, opened 2021-02-08
Optimize docker build

subdavis:
Fixes for https://github.com/pyca/cryptography/issues/5771

Also re-ordered for cache optimization.  things that change less frequently should happen earlier in build.


BryonLewis (approved):
Pulled, built and ran some pipelines to make sure there were no problems.

PR #571, opened 2021-02-12
Rename to DIVE

### LINKED ISSUE #502: Rename VIAME-Web to DIVE
Blocked on #487

subdavis:
* Rename most references from VIAME Web to DIVE
* Rename server from `viame_*` to `dive_*`
* Change default branch from `master` to `main` because this is the new standard and has been for many months.

---- 

Also fixes a mistake in docker-compose where, on yml section merge, the abstract section's `environment` was getting dropped.   We didn't notice because so far we have only used the default values.  Removed the section getting overwritten.

fixes #502


client/src/components/LayerManager.vue
```diff
@@ -171,7 +171,7 @@ export default defineComponent({
           editingTracks.push(trackFrame);
         }
         if (editingTracks.length) {
-          if (editingTrack && !editAnnotationLayer.creationIncomplete()) {
+          if (editingTrack) {
```
aashish24, line 174: is this change intentional?

subdavis: No, it was an outdated branch.

BryonLewis (approved):
Sorry it took me a bit to get the container launched correctly.

Everything I can think of seems to be changed except endpoints (future PR).
Impressed you even got the `Written on {date.time} by dive:python writer`

PR #572, opened 2021-02-12
Docker/bloat

subdavis:
Rebase after #571 lands.

Reduces from 33gb to 25gb.

We need to remove the pipelines from the image build.


subdavis:
This will be rolled into #574

PR #584, opened 2021-02-18
[Critical] Attribute processing bug

subdavis:
See recent email thread.

Detection attributes were being missed on ingestion.


BryonLewis (approved):
Thanks for finding that.

PR #587, opened 2021-02-18
Proper insert location and delete selection behavior

### LINKED ISSUE #580: [BUG] Removing track should switch selection to nearest neighbor, not 0.
UX bug.

"This is very time consuming to have to scroll back down to the detection number we left off at"

subdavis:
closes #580


BryonLewis (approved):
Works as intended, just wondering when we will get the subsequent issue that the track Ids are no longer in sequential order.

subdavis:
Track IDs were never in sequential order.    It's a little weird.  I'm sure it will come up.

PR #602, opened 2021-02-25
Transcode anamorphic video.

### LINKED ISSUE #570: [BUG] Video display for unusual pixel ratios
**Describe the bug**

Brandon Davis  11:51 AM
Good morning, David.
Bryon and I have just run headlong into a weird video codec / rendering issue.  Here's some ffprobe output.
```
ffprobe version 3.4.8-0ubuntu0.2 Copyright (c) 2007-2020 the FFmpeg developers
  built with gcc 7 (Ubuntu 7.5.0-3ubuntu1~18.04)
  configuration: --prefix=/usr --extra-version=0ubuntu0.2 --toolchain=hardened --libdir=/usr/lib/x86_64-linux-gnu --incdir=/usr/include/x86_64-linux-gnu --enable-gpl --disable-stripping --enable-avresample --enable-avisynth --enable-gnutls --enable-ladspa --enable-libass --enable-libbluray --enable-libbs2b --enable-libcaca --enable-libcdio --enable-libflite --enable-libfontconfig --enable-libfreetype --enable-libfribidi --enable-libgme --enable-libgsm --enable-libmp3lame --enable-libmysofa --enable-libopenjpeg --enable-libopenmpt --enable-libopus --enable-libpulse --enable-librubberband --enable-librsvg --enable-libshine --enable-libsnappy --enable-libsoxr --enable-libspeex --enable-libssh --enable-libtheora --enable-libtwolame --enable-libvorbis --enable-libvpx --enable-libwavpack --enable-libwebp --enable-libx265 --enable-libxml2 --enable-libxvid --enable-libzmq --enable-libzvbi --enable-omx --enable-openal --enable-opengl --enable-sdl2 --enable-libdc1394 --enable-libdrm --enable-libiec61883 --enable-chromaprint --enable-frei0r --enable-libopencv --enable-libx264 --enable-shared
  libavutil      55. 78.100 / 55. 78.100
  libavcodec     57.107.100 / 57.107.100
  libavformat    57. 83.100 / 57. 83.100
  libavdevice    57. 10.100 / 57. 10.100
  libavfilter     6.107.100 /  6.107.100
  libavresample   3.  7.  0 /  3.  7.  0
  libswscale      4.  8.100 /  4.  8.100
  libswresample   2.  9.100 /  2.  9.100
  libpostproc    54.  7.100 / 54.  7.100
Input #0, mpeg, from 'wrasse.mpg/wrasse.mpg':
  Duration: 00:00:07.16, start: 0.224400, bitrate: 4824 kb/s
    Stream #0:0[0x1e0]: Video: mpeg2video (Main), yuv420p(tv, top first), 720x480 [SAR 32:27 DAR 16:9], 29.97 fps, 29.97 tbr, 90k tbn, 59.94 tbc
    Stream #0:1[0x80]: Audio: ac3, 48000 Hz, stereo, fltp, 256 kb/s

```

When transcode to mp4, the width, height, DAR, PAR are the same.  When we play it in a browser, the width is reported by a video player as 853.  Indeed, the displayed aspect ratio (as checked with a measuring tape) is 16:9 (not 3:2, as 720x480 actually is)
Naturally, when we store annotations from this stretched coordinate space, they don't match up with the video pixels.  Have you had to deal with this problem before?  What's a good way to manage these?  Does GeoJS have tools that will put drawn shapes into the video's coordinate space given the PAR? Should we transform them at export time?
@Bryon Lewis please add any detail I may have missed. (edited) 

manthey  1:33 PM
If you know you have non-square pixels, there are several ways to apply a transform in GeoJS.  ...
1:35
You can construct an appropriate proj4 string to do the transform.  Or you could add a position() accessor to the shape feature layers to transform the coordinates for rendering (e.g., feature.position((d, i) => {x: d.x * 1.2, y: d.y}))
1:36
I think if the video is already being stretched to the correct shape in the quad, then you'd want to use the position accessor.

Brandon Davis  1:48 PM
I'd like to try both and see which makes sense for us.  the feature position thing seems relatively straightforward.
Do you have any references to an example where geo.transform is used?  If I construct an affine transform and apply it to the whole map through geo.map.spec.gcs, will that do what I epxect?

manthey  1:49 PM
The feature position will work for rendering, but if you are drawing annotations it will probably be more work than the transform.  I know I have an example of applying the transform, but will take me a short while to dig it up.
1:50
The transform is applies via gcs, yes.

Brandon Davis  1:53 PM
Thank you.

manthey  2:09 PM
Try a gcs string like +proj=longlat +axis=enu +s11=1.2 +s12=0 +s21=0 +s22=1 +xoff=0 +yoff=0.
2:10
You would do something like layer.gcs('+proj=longlat +axis=enu +s11=1.2 +s12=0 +s21=0 +s22=1 +xoff=0 +yoff=0') for the annotation layer. (edited) 

Brandon Davis  2:11 PM
I think this will get us most of the way there.  Thank you!

**Related data**

https://viame.kitware.com/girder#user/60243dea4d85a3cb4e2250c5/folder/602441854d85a3cb4e2250d2

subdavis:
Pixel aspect ratio !== 1:1 is not supported.  Force it in transcoding.

This may slightly alter the actual width/height of the video

https://viame.kitware.com/#/viewer/601a0582503d63fdb4d021f5
https://viame.kitware.com/#viewer/602441854d85a3cb4e2250d2

fixes #570


subdavis:
Screenshot from 2021-02-24 23-17-06

Works a lot better when you use `SAR 1:1`.

Interestingly, the tracker output is different too, so I think even the tracker hates weird SAR values.

subdavis:
I also tried using gcs.

```
      this.featureLayer.gcs([
        '+proj=longlat',
        '+axis=esd',
        '+s11=1', '+s12=0', '+s13=0',
        '+s21=0', '+s22=1', '+s23=0',
        '+s31=0', '+s32=0', '+s33=1',
      ].join(' '));
```

Only parts of this work.  the `+axis` bit works on everything.  But the rest of the affine transform only works on text features, and is ignored by polygons.

This reinforces my fear of client-side transformations.  If we had a regression, we would find out about it much later because this is such a niche use case, and we would have to locate and correct all the corrupted manual annotations that users may have created.  This kind of mistake could be very costly in time and customer relationship (since we could have corrupted hours of manual labor in creating annotations)

Imagine if we had to script updates to all line geometry only because some future change cause it to start ignoring this transform.

I also don't like the additional mental hoops I'd have to jump through.  "Does 175px really mean 175?"

One thing I can do is distort the video rather than the annotations.  This doesn't seem ideal for users, but it seems a lot safer than hoping all annotations types apply the same transform.

Screenshot from 2021-02-25 08-19-15

Screenshot from 2021-02-25 08-22-22

BryonLewis:
did you attempt this on the wrasse.mp4 data?  It was giving me an error related to the width being non-divisible by 2.  It wants to scale the video from 720x480 to 853x480 but that then is a non-even width.
https://stackoverflow.com/questions/20847674/ffmpeg-libx264-height-not-divisible-by-2
Adding that filter to the end of the scaling resets it to 854x480 and allows it to transcode properly.
I.E: ` '-vf "scale=iw*sar:ih,setsar=1,pad=ceil(iw/2)*2:ceil(ih/2)*2"',`

subdavis:
I have a similar local change that I forgot to push, but they aren't the same.

`ceil(iw/2)*2:ceil(ih/2)*2` is not the same as `-vf "scale=ceil(iw*sar):ih,setsar=1"`

Mine causes the final aspect ratio to be different than the input by some fraction of a pixel width  Yours seems like it must be doing the same.

BryonLewis:
With a really weird SAR you could end up getting a non-even width.
Source file: `523x222` SAR: `49:33`
523*(49/33) = 776.575757 | ceil of that is 777

The padding way would take that 776.5757/2 = 388.2878 | ceil = 389 | double = 778

I think mine will add up to 1.9999999  pixels black letter box like scaling on the ends while maintain aspect ratio, an updated version of your command (`ceil(iw*sar/2)*2:ih`) will stretch it the additional maximum of 1.99999 pixels.  I don't know which one is proper to do.

BryonLewis:
Update:  Just doubled checked and it looks like ffmpeg supports round so you could cut the maximum amount of being off to 1 pixel instead of 2.  But then I would limit to squashing the output using your method, don't think you can pad negative space:
 `'-vf "scale=round(iw*sar/2)*2:ih,setsar=1"',`

That would have an output width of 776 instead of the 778 for the weird source video.

server/dive_tasks/tasks.py
```diff
@@ -415,7 +412,14 @@ def convert_video(self: Task, path, folderId, auxiliaryFolderId, itemId):
     if process.returncode == 0:
         manager.updateStatus(JobStatus.PUSHING_OUTPUT)
         new_file = gc.uploadFileToFolder(folderId, output_path)
-        gc.addMetadataToItem(new_file['itemId'], {"codec": "h264"})
+        gc.addMetadataToItem(
+            new_file['itemId'],
+            {
+                "source_video": False,
+                "transcoder": "ffmpeg",
```
BryonLewis, line 419: Just a question:
Is `transcoder: "ffmpeg"` there with the assumption that we may use the VIAME stuff in the future for image/color enhancement?

BryonLewis (changes requested):

client/platform/desktop/constants.ts
```diff
@@ -20,6 +21,9 @@ export const fileVideoTypes = [
   'avi',
   'mov',
   'wmv',
+  'mpg',
+  'mpeg',
+  'mp2',
```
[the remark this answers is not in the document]
BryonLewis: `.mp4;.MP4;.mpg;.MPG;.mpeg;.MPEG;.avi;.AVI;.wmv;.WMV;.mov;.MOV;.webm;.WEBM;.ogg;.OGG`
List of supported file types that Matt gave me.

server/dive_tasks/tasks.py
subdavis: Yes.  `source_video=False` was needed as a difinitive marker that this is a derived video, not source.  `transcoder` indicates what did the conversion.   `codec` just has the codec, which shouldn't be used as a proxy for any other properties as it was before.

BryonLewis (approved):
Tested the web version and desktop version with some weird SAR datasets and it works properly.  Web has the benefit of converting everything and the desktop was properly detecting x264/mp4s that had non 1:1 SAR values and converting them.  I then ran pipelines through the results making sure they were applied to the converted versions.

BryonLewis (approved):
did another pass through with my test datasets and all seem to be working in both modes.

PR #616, opened 2021-03-05
Upgrade pipelines

subdavis:
supercedes #574 

* Reduce worker image size from 38GB to 12GB.
* Remove all usage of ansible for deployments
* Use docker hub for weekly deployments with simple pull-and-restart workflow (containerrr/watchtower)
* No longer need to remove pipeline volumes ever.
* Adds ability to upgrade addons without redeployment, and optional caching of addon archives.


BryonLewis (commented):
This is just general stuff.  I haven't done a complete review but I like the behavior so far. 

 Easy enough to trigger the pipeline download.  I may wish for a little more transparency in the download/extract process (first zip file taking like 8 minutes before logging it was downloading).

It properly showed the pipelines afterwards and it ran the subset of them.  I think is a much more sustainable way of adjusting pipelines in the future instead of requiring the volume destroy and rebuilds.

BryonLewis (approved):

subdavis:
Will merge after the redeploy tonight. This will require a bit of downtime to do.

PR #627, opened 2021-03-09
Compensate for Chromium MPEG inaccuracy.

### LINKED ISSUE #610: [BUG] Off-By-One video sync issue
**Describe the bug**

Bounding boxes can end up off-by-one on fast-moving targets in video. I think this has to do with how we treat annotations as rising-edge vs falling-edge, so for wide frame widths, you can end up quite far from the exact point where the annotation was created.

**To Reproduce**
 Open dataset, play, pause, and step around until you see fish outside their boxes.

**Related data**

* https://viame.kitware.com/#/viewer/602ac90e2ccc1bc096d6d3f1

subdavis:
This problem was exactly what I expected: falling on the wrong side of the frame boundary due to small variations in real frame size and maybe floating point errors.

Also force video playback pause events to snap to nearest frame boundary so that detections created on video that was paused cannot happen in the middle of a frame's width. 

fixes #610


BryonLewis (dismissed):
Sorry it took me so long to conceptualize what was going on.

subdavis:
Updated to reflect the true cause of this bug: Chrome!

See detailed inline comment about how and why this fix is correct.

BryonLewis (approved):

PR #634, opened 2021-03-11
Transcode audio to AAC to prevent incompatible container formats

### LINKED ISSUE #353: [BUG] Debug transcoding failures
**Describe the bug**
Some user uploads fail to transcode.  We should figure out why.

**Related data**
https://viame.kitware.com/#/folder/5e97308a95a54f6165d90dbf

There may be more info in girder logs or sentry.

subdavis:
Note that this fixes one of the only failures I could find related to video where the sample data had not been removed.  This may not  be the cause of every transcoding issue.

fixes #353

### DIFF
--- a/client/platform/desktop/backend/native/linux.ts
+++ b/client/platform/desktop/backend/native/linux.ts
@@ -43,7 +43,8 @@ const ViameLinuxConstants = {
       '-c:v libx264',
       '-preset slow',
       '-crf 26',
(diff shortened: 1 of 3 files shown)


BryonLewis (approved):
There is that new issue I added this morning regarding conversion of tiled Tiffs #632 in ffmpeg.  I haven't dived into that yet, but placed it in triage.

PR #643, opened 2021-03-15
upgrade gwc to 3.0.1

subdavis:
* Tested and builds with better type coverage than before.
* Apparently also manages to remove a double-copy of vue and axios....
* Fixes google analytics
* **important** implements notificationBus using EventSource in TS.  GWC EventSource isn't functional.
* Add support for eventSource in webpack dev server

This removes the notification check every few seconds and should dramatically reduce the amount of noise in server logs.


client/platform/web-girder/main.ts
```diff
@@ -42,10 +43,13 @@ Promise.all([
...
-    provide: { girderRest, notificationBus, vuetify },
+    provide: { girderRest, vuetify },
     render: (h) => h(App),
   })
     .$mount('#app')
     .$snackbarAttach()
     .$promptAttach();
+
+  /** Start notification stream if everything else succeeds */
+  registerNotifications(girderRest).connect();
```
BryonLewis, line 54: Should we rely on the `login` signal from the `restClient` to connect instead of doing it here?
If you sit at the login screen it is constantly attempting to reconnect right now because there is a 401 error so it's in the error/reconnect loop.  I could be missing something though.

client/dive-common/components/RunTrainingMenu.vue
```diff
@@ -130,10 +130,11 @@ export default defineComponent({
 
         <v-alert
           dense
-          type="warning"
+          color="warning"
+          class="mx-2"
```
BryonLewis, line 134: Feel entirely free to ignore:  `outlined`? 
image
image

BryonLewis (changes requested):
Just a few questions regarding the jobList notification and when it should connect.

client/src/notificatonBus.ts
```diff
@@ -0,0 +1,84 @@
...
+ * Register notifications directly on the girderRest instance using
+ * the EventSource api.
+ *
+ * @param rc Girder RestClient
+ */
+export default function registerNotifications(rc: RestClient) {
+  const ES = window.EventSource;
+  const withCredentials = true;
+  const timeoutSeconds = 300;
+  const retryMsDefault = 8_000;
```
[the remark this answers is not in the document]
subdavis: Underscores are ignored in js numbers.  I use them to signify orders of magnitude.  This is 8 seconds. I originally had it at 10 seconds, so the underscore is particularly nice for readability there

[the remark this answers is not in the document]
BryonLewis: Thank You for the info.  I didn't know and did a quick search but didn't see anything immediately in the first page (mostly because my google terms were using underscore so I got a lot of underscore.js results).  Now adjusting the terms I see the info.

client/dive-common/components/RunTrainingMenu.vue
```diff
@@ -118,47 +118,65 @@ export default defineComponent({
...
+          <p>
+            Input data:
+          </p>
+          <v-chip
+            v-for="id in selectedDatasetIds"
+            :key="id"
+            small
+            class="mr-1 mb-2"
+          >
+            <pre>{{ id }}</pre>
```
BryonLewis, line 167: I like it, but these numbers only mean stuff you and I and people who understand the routing.  If it was the folder name that might help a bit more for regular users.

client/dive-common/components/RunPipelineMenu.vue
```diff
@@ -149,32 +152,34 @@ export default defineComponent({
...
-                    {{ pipeTypeDisplay(pipeType) }}
                     <v-icon
-                      left
+                      right
                       color="accent"
-                      class="ml-0"
+                      class="mr-0"
                     >
-                      mdi-menu-right
+                      mdi-menu-left
```
BryonLewis, line 174: I'm guessing it's mostly from only knowing right-to-left languages but it feels different. This isn't me requesting a change or anything and I'd wait to see what others say.  I played around for a bit seeing if I could do some stuff width nudging and moving it around but nothing was better than what you currently have.

BryonLewis (dismissed):
I'm good with this as is and it addressed all my previous comments.  Feel free to modify the `datasetID` in training to a folder if you want, or you can leave it as an Id.  I've noticed we are a bit inconsistent in the jobs panel.  It looks like conversions are run on Ids and pipelines are run on the folder name so we do use both in other locations.

client/dive-common/components/RunPipelineMenu.vue
subdavis: Yeah, it's  odd.  If you want to make them open to the right and just not worry about the case where it covers the lower menu, I'm good with that.

client/dive-common/components/RunTrainingMenu.vue
subdavis: It wasn't available without adding more props.  I agree it's a bit odd.  Mostly I wanted a clear visual indicator of quantity.

I could just change it to say "Train on {count} datasets" if that's preferred.

client/dive-common/components/RunPipelineMenu.vue
BryonLewis: That sounds good.

BryonLewis (approved):

PR #670, opened 2021-03-28
Fix circular import crash, Remove VTK js again.

subdavis:
* Critical: Refactor constants to `dive_utils` to prevent circular import crash.  `main` is currently broken, workers fail to start
* Adds very stupid test to confirm that imports all work.  This test would have broken on the initial problem, but I forgot to write the test first to show it failing.
* Removes VTK.js again because apparently it got added back in.
* Refactor prod `TIMEZONE` for duplicati and watchtower.  Watchtower is currently updating at the wrong time (2AM UTC).  Apparently `/etc/timezone` isn't working as advertised.  w/e.


BryonLewis (changes requested):
Sorry I dropped the ball on my second review and not thoroughly checking the requested change to utilize the constants across more locations.
Did a more thorough test and review this morning.
I see that I also missed this:
https://github.com/Kitware/dive/blob/fix/circular-vtk-regressions/server/dive_tasks/tasks.py#L416-L418
I believe that this should be the `TrainedPipelineMarker`.  I believe we want the meta to be `trained_pipeline` instead of `trained`

subdavis:
Great catch, fixed in ebdc7bb

BryonLewis (approved):

PR #679, opened 2021-04-01
Add data portal options sidebar

### LINKED ISSUE #654: [FEATURE] Some way to display author information / usage rights alongside datasets
One group requested the ability to add a readme with annotation authorship at one point to those that are in training_data

subdavis:
Experimentation.  We're running out of room in the data browser header to keep adding buttons, and it'd be nice to have a place to show more information about datasets than just "launch viewer"

Screenshot from 2021-04-01 13-11-43

Or reverse (which makes more sense for menu opening)

Screenshot from 2021-04-01 13-23-39

This will give us more room to build controls and information.

Most of the code came from copy/pasting the DataDetails component from GWC because it's not flexible enough to support this and it's easier just to copy/modify

fixes #654


subdavis:
Screenshot from 2021-04-01 16-49-25
Screenshot from 2021-04-01 16-49-04

Final pass.

client/platform/web-girder/views/DataDetails.vue
```diff
@@ -0,0 +1,141 @@
...
+        };
+        const typeCounts = this.value.reduce(reducer, {
+          item: 0,
+          folder: 0,
+          size: 0,
+        });
+        const countMessages = ['item', 'folder']
+          .filter((k) => typeCounts[k] > 0)
+          .map((k) => `${typeCounts[k]} ${k}(s) selected`);
+        const sizeMessage = `Total size: ${this.formatSize(typeCounts.size)}`;
```
BryonLewis, line 134: how do you feel about not display size of it is 0 for folders?  It just feels a little off with the size being there for a folder which clearly has items in it?

client/platform/web-girder/views/Home.vue
```diff
@@ -157,12 +180,53 @@ export default Vue.extend({
...
+      fill-height
+      :fluid="$vuetify.breakpoint.mdAndDown"
+    >
       <v-row
         class="fill-height nowraptable"
-        no-gutters
       >
-        <v-col :cols="12">
+        <v-col cols="3">
+          <DataDetails
```
BryonLewis, line 191: scroll sticky?  I know that isn't something that the file browser has so it isn't really that important.
image

client/platform/web-girder/views/DataDetails.vue
```diff
@@ -0,0 +1,141 @@
...
+        return data;
+      }
+      return null;
+    },
+  },
+  computed: {
+    title() {
+      return this.details
+        ? (this.details.name || this.formatUsername(this.details))
+        : `${this.value.length} Selection(s)`;
```
BryonLewis, line 99: maybe a specialized display for Object that is just `type:"root"` and `type:"'collections"` instead of the ` Selections(s)
image

image

BryonLewis (changes requested):
My first look and it works really well and I like it more than the toolbar menu system.

Just a few questions and comments about the display during some corner cases.

subdavis:
Screenshot from 2021-04-06 16-49-25
Screenshot from 2021-04-06 16-49-02


Responded to comments, also moved details markdown below the editor like this because I like the GitHub vibe and that much information would clog up the sidebar.  I'd like to start documenting our public datasets (not individually, but each subfolder group would be nice to say where the data came from, how it was annotated, and how close to ground truth we think it is)

client/platform/web-girder/views/Home.vue
subdavis: I wasn't really feeling this one. You're welcome to add it

BryonLewis:
I like that details markdown added in there.

client/platform/web-girder/views/Home.vue
```diff
@@ -1,7 +1,9 @@
 <script>
 import Vue from 'vue';
 import { mapMutations } from 'vuex';
-import { getLocationType, GirderFileManager } from '@girder/components/src';
+import {
+  getLocationType, GirderFileManager, GirderMarkdown, GirderBreadcrumb,
```
BryonLewis, line 5: minor linter, import never used.

BryonLewis (changes requested):
Minor linter warning.  Only additional thing looks like GWC and it's allowing a user to 'attempt' to create a folder inside of a container they don't have write access on.  It errors properly but it's a little annoying that the button is enabled.

subdavis:
> Only additional thing looks like GWC and it's allowing a user to 'attempt' to create a folder inside of a container they don't have write access on. It errors properly but it's a little annoying that the button is enabled.

Yeah, this is an upstream issue in GWC's FileManager.  You have to make an extra request to check permissions, which at some point we decided was not we wanted to do.   The AccessControl context menu does this when a user does a right-click to decide whether or not to show the button.  It's obnoxious.

client/platform/web-girder/views/Home.vue
subdavis: fixed

BryonLewis (approved):

PR #689, opened 2021-04-08
Docker nvidia docs out of date

subdavis:
* Nvidia docker2 is not deprecated, it is the current supported nvidia docker runtime.
* GPU support for compose has now landed.  I didn't upgrade the actual compose version because if-it-aint-broke, but if we wanted to use more fine-grained GPU device selection, we could now.
* Explicitly mention that the docker install is linux-only, and point to alternatives.


subdavis:
I also messed up the watchtower command when I enabled rolling release.  It interpreted the comment string as a container name, couldn't find any containers with that name, and noop'd.

BryonLewis (approved):

PR #690, opened 2021-04-13
Sort order image fix

### LINKED ISSUE #494: [BUG] Improve filename resolution for image sequence ordering
There's a couple fringe cases where image sequences aren't loading in the right order. While hopefully people usually use things like 0001, 0002, ....  I've also seen (1), (2), ... (10), (11), which is loading out of order, e.g. attached below

e1
e2
e3

subdavis:
Fixes issues with incorrect load order in this data.

https://viame.kitware.com/#/viewer/6074b9b0244544940f7755fe
https://viame.kitware.com/#/viewer/6074b9b3244544940f775a3f

In string sorting, `1500` is considered smaller than `20` because the sort is character-by-character.  This change would consider all numeric values in the string, and only move on to comparing non-numeric values if the numbers were equal, i.e. `1000a1`, `1000b1`.

fixes #494


BryonLewis (commented):
This is my quick first look at this so I might be making some mistakes or interpreting wrong, please let me know.

I think the issue with some of the other data was un-padded numbers not properly sorting.  I'm not sure if this would fix that without implementing a grouping by numbers compare.  Where the regex looks for sequential numbers and moves left to right comparing them in files if the groups match.  A flattening of all numbers could give "incorrect" (I don't know what correct is) ordering.

A little test example:
```
  [
    [
      'avi_0002.1.png',
      'avi_400005.0.png',
      'avi_400004.12.png',
      'avi_1100004.0.ng',
      'avi_400003.0png',
      'avi_400006.0png',
    ],
    [
      'avi_0002.1.png',
      'avi_400003.0.png',
      'avi_400004.12.png',
      'avi_400005.0.png',
      'avi_400006.0.png',
      'avi_1100004.0.png',

    ],
  ],
```
Because we have a second un-padded (12) number the `avi_400004.12.png` will always be placed as last because it is the largest number.  Where as sorting should maybe move number group by number group in the two arrays generated from the regex.  This will be much longer but may give a more accurate sorting order?  I haven't thought of the weird corner cases that may come up with this method yet.

subdavis:
That's a great example case.  Let me see what I can do.

BryonLewis (approved):
I think this is a much better sorting mechanism that should hopefully address a majority of the edge cases.  There are probably still some out there (AM/PM) but I think that should be outside the scope of what we handle.

PR #718, opened 2021-04-22
New user email

subdavis:


BryonLewis (dismissed):
I configured everything up until the actual sending of email and it looks like it kicked off everything properly.

One thing I did note is that even a clean docker build (--no-cache) didn't place the mail_templates in the container `/opt/lib/python3.8/site-packages/dive_server` directory  I had to uninstall my dive_server from my host.  It must of been picking up some cached stuff.

BryonLewis (approved):

PR #751, opened 2021-05-12
Bump versions

subdavis:
Replaces #749 #747 #742


BryonLewis (dismissed):

subdavis:
I found a better way of exlcuding VTK js with webpack externals.  This will allow us to merge dependabot PRs in the future.

BryonLewis (approved):
ran through some basic tests and it seems to work.  Confirmed the size difference (2.2MB vs 4.0) between the builds and using vtkjs

PR #778, opened 2021-05-28
Trained pipes missing in desktop

subdavis:
Not sure why this used to be "2"

The contents of a directory looks like this:

```
~$ ll /home/brandon/VIAME_DATA/DIVE_Pipelines/sample_detector
.rw-rw-r--   92 brandon 27 May 15:00 custom_cfrnn.lbl
.rw-rw-r-- 8.2k brandon 27 May 15:00 custom_cfrnn.py
.rw-rw-r-- 4.1k brandon 27 May 15:00 detector.pipe
```


subdavis:
Invalid change.

PR #787, opened 2021-06-01
Switch from checkbuild.sh to tox for server

### LINKED ISSUE #775: Switch from `checkbuild.sh` to `tox` for server.
See other projects in D&A for examples of how to do this.

f4str:
fixes #775 

Replaced `server/checkbuild.sh` with a `server/tox.ini` script. Incorporated `server/dev-requirements.txt` into `server/setup.py` to follow tox conventions. Modified `.github/workflows/blank.yml` to run tox instead of manually running each test. Updated documentation in `server/README.md` to specify how to run the new tox script (also moved parts from `CONTRIBUTING.md` since it makes more sense to be here).


f4str:
Since tox automatically creates virtualenv environments, the only dependency that needs to be installed from now on is tox itself
```bash
pip install tox
```

The CI in `.github/workflows/blank.yml` will run the tox command by itself which is set to run the lint check for `black` and `isort`, the type check for `mypy`, and the tests for `pytest`. This command is simply 
```bash
tox
```
which is set by default to run the environments
```bash
tox -e check, type, test
```

When running locally, the linter needs to invoked to actually format the files and not just perform checks. It's also worth running the type checker and tests. This is all done through tox by manually selecting the environments
```bash
tox -e format, type, test
```
This is documented in `server/README.md`

subdavis (approved):
LGTM﻿

PR #794, opened 2021-06-04
When exporting detections to csv also add option to only export selected types

### LINKED ISSUE #695: [FEATURE] When exporting detections to csv also add option to only export selected types
The user might want to only export checked types (I think this option should still be unchecked by default, however)

f4str:
fixes #695 

Web version new checkbox: 
image

Desktop version new checkbox:
image


client/platform/web-girder/views/Export.vue
```diff
@@ -34,15 +34,18 @@ export default defineComponent({
...
     const savePrompt = ref(false);
     const currentSaveUrl = ref('');
     let save = () => Promise.resolve();
     let pendingSaveCount = ref(0);
+    let checkedTypes = ref([] as readonly string[]);
 
     if (props.blockOnUnsaved) {
       save = useHandler().save;
       pendingSaveCount = usePendingSaveCount();
+      checkedTypes = useCheckedTypes();
```
BryonLewis, line 48: I don't know if this should be in the `blockOnUnsaved` prop and maybe it should be outside of it where the variable is initialized?  I know `blockOnUnsaved` set outside of here in the `ViewLoader.vue` but the default properly is false.  `checkedTypes` once bound in the beginning should auto update anytime the checked types are changed.  I could be mistaken on this and it's handling an edge case I don't know about.

server/dive_server/serializers/viame.py
```diff
@@ -280,17 +280,20 @@ def export_tracks_as_csv(
     filenames=None,
     fps=None,
     header=True,
+    typeFilter=set(),
```
BryonLewis, line 283: Might be an opportunity to add some more dataset export testing to `test_serialize_viame_csv.py`?  I'd check with @subdavis and see what he thinks about possible adding the typeFiltering to here.  Could also be an indicator that we start adding nodejs serializer testing to match the python.  Ensuring that same data in = same data out between the two.

BryonLewis (commented):
In my testing this works really well on desktop/web.  I just have one general question about the placement of `useCheckedTypes` in `Export.vue` (I could very well be missing something).  Other than that the possibility of maybe adding some additional testing the python export script and maybe some initial testing to the node.js/desktop side.

server/dive_server/serializers/viame.py
subdavis: Agree with Bryon

client/platform/web-girder/views/Export.vue
subdavis: `blockOnUnsaved` is a proxy indicator for "this export dialog was mounted on the viewer instead of the data browser".

If this were outside this block, it would cause a runtime exception because the Provides api was not injected.

There's a conditional on menuOpen below to prevent requests to the server unless the menu is actually open.

No changes needed, IMO.

client/platform/desktop/frontend/components/Export.vue
```diff
@@ -183,6 +187,16 @@ export default defineComponent({
               </span>
             </div>
           </template>
+
+          <template v-if="checkedTypes.length">
+            <v-checkbox
+              v-model="data.excludeUncheckedTypes"
+              label="export checked types only"
```
subdavis, line 194: Maybe add a `details` prop with `Export only the types currently enabled in the type filter`.  Get rid of `hide-details`.

I think the extra explanation is needed.

subdavis (changes requested):
* Agree with bryon about adding a test case or two in python
* Couple other minor comments.

Looks great, confirmed working, no issues.

client/platform/web-girder/views/Export.vue
BryonLewis: thanks, that was the case I was missing.

f4str:
Added some more unit tests for exporting with type filters and importing with missing types. Also added the checkbox hint.
image

subdavis (approved):
:100: looks great, thanks.

I noticed in the unit tests that we aren't testing the confidence threshold either.  It would be good to test that, especially in the case where both confidence and track type are filtered and a track has multiple types where the behavior might conflict.

PR #799, opened 2021-06-09
Fix log buffering issue

subdavis:
Don't keep the entire log history in renderer: splice to truncated length on every update.


BryonLewis (approved):
looks good

PR #804, opened 2021-06-10
Refactor prompt-service to use composition-api

### LINKED ISSUE #664: [Housekeeping] Refactor $prompt to use composition-api
Prompt is a nice feature that needs to be re-implemented in typescript with composition-api.

f4str:
fixes #664 

The files `client/dive-common/vue-utilities/prompt-service/Prompt.vue` and `client/dive-common/vue-utilities/prompt-service/index.ts` are both rewritten using composition-api and typescript.

Instead of using `root.$prompt({...}` to call the prompt box, instead import the prompt provider
```typescript
import { usePrompt } from 'dive-common/vue-utilities/prompt-service'
```
Call the prompt box by getting the `PromptService` class and calling the show function
```typescript
const prompt = usePrompt();
prompt.show({...});
```
This will be updated in the README.md later


client/dive-common/vue-utilities/prompt-service/index.ts
```diff
@@ -0,0 +1,106 @@
...
+let promptService: PromptService;
+
+export function usePrompt() {
+  // in vue 3 should use inject instead of singleton
+  return promptService;
+}
+
+export default function (vuetify: Vuetify) {
+  return function install(Vue: VueConstructor) {
+    Vue.extend({ vuetify, ...Prompt });
```
f4str, line 94: There are issue with this line. This currently does not properly inject vuetify into `Prompt.vue` and the colors of the prompt box are incorrect. The code in the original javascript file was
```typescript
Prompt.vuetify = vuetify
```
However, this is not valid in typescript and does not compile. Needs further investigation on a fix.

subdavis (changes requested):
@f4str I made a couple changes here to address the vuetify issue and simplify the "resolve" logic.

Please LMK if this looks good to you and we can merge.

f4str:
@subdavis everything looks good to me. Seems to work on both web and desktop as far as I can see. I also updated the README to remove the redundant statement.

subdavis (approved):

PR #813, opened 2021-06-17
Fix prompt delete

subdavis:
Simple bug.  promptShow didn't return the awaited value.  Remove function, pass prompt direclty.


BryonLewis (approved):
Tested in a few instances.

PR #824, opened 2021-06-29
Updated head/tail keypoint appearance to filled/hollow

### LINKED ISSUE #243: Update head/tail position keypoint appearance
I think as opposed to having to differently colored dots, it would look better if there were 2 dots which match the box color, where they are just slightly different (e.g. one being a filled in circle and one being a hollow circle or the like) with a line between them

f4str:
fixes #243 

Updated head and tail keypoints to be the same color as the bounding box. Instead of the arbitrary blue/orange colors, the head keypoint is filled in and the tail keypoint is hollow.

Screenshot when not selected:
Screenshot from 2021-06-29 16-54-57

Screenshot when selected:
Screenshot from 2021-06-29 16-55-20


subdavis:
Looks like a good opportunity to revisit the head/tail keypoint docs.

https://kitware.github.io/dive/Annotation-QuickStart/#head-tail-annotations

Could you make sure that's up to date?  I know the GIF won't have the new style, and that's fine, but maybe add a screenshot of a single fish with head/tail and explain the hollow/filled thing?

BryonLewis (approved):
Looks good.  Can you just confirm one thing if you are on ubuntu 20.04.  Can you see the CSS grab icon when hovering over points to edit them.  A secondary confirmation could be done here: https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

UPDATE:  I found the reason and the default Ubuntu 20.04 Cursor theme lacks Grab, it can be resolved with a symlink but going forward we may want to swap the use of grab in our icon set to 'grabbing'.

f4str:
> Looks good. Can you just confirm one thing if you are on ubuntu 20.04. Can you see the CSS grab icon when hovering over points to edit them. A secondary confirmation could be done here: https://developer.mozilla.org/en-US/docs/Web/CSS/cursor
> 
> UPDATE: I found the reason and the default Ubuntu 20.04 Cursor theme lacks Grab, it can be resolved with a symlink but going forward we may want to swap the use of grab in our icon set to 'grabbing'.

@BryonLewis In Ubuntu 20.04, the default yaru theme does not come with a mouse grab icon. I tried installed a custom mouse icon theme which does have it and this what it looks like:
Screenshot from 2021-06-30 09-17-56
This only appears when on editing mode for the points (as expected).

PR #832, opened 2021-07-08
Checkbox to use annotated frames only when training

### LINKED ISSUE #771: [FEATURE] Checkmark when training models to indicate only use frames with groundtruth on them
TODO: Matt D put the setting here which does this on the train side

f4str:
resolves #771

Checkbox added in training menu to choose option to only train on annotated frames and ignore frames that are not annotated. This is enabled by default for both web and desktop. This checkbox is passed as an additional parameter to the kwiver call.

Web:
Screenshot from 2021-07-08 16-50-42

Desktop:
Screenshot from 2021-07-08 16-45-17

Also modified CLI to incorporate this new change by adding the `--annotatedFramesOnly` flag which is false by default.

### DIFF
--- a/client/dive-common/apispec.ts
+++ b/client/dive-common/apispec.ts
@@ -105,7 +105,9 @@ interface Api {
   runPipeline(itemId: string, pipeline: Pipe): Promise<unknown>;
 
   getTrainingConfigurations(): Promise<TrainingConfigs>;
-  runTraining(folderIds: string[], pipelineName: string, config: string): Promise<unknown>;
+  runTraining(
+    folderIds: string[], pipelineName: string, config: string, annotatedFramesOnly: boolean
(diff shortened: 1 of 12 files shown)


f4str:
From previous discussions, changed the checkbox to be off by default on both web and desktop.

Ran tests on both web and desktop to verify the `--gt-frames-only` flag is being passed to kwiver. Also tested the desktop cli tool where running `yarn divecli run-training ARGS [--annotatedFramesOnly]` does pass the flag to kwiver when appropriate.

The final step is to final the placement for the checkbox. On web, this is most likely already the best placement:

image


For desktop, the current placement is below the dataset table and left of the button:

image

Another option is to place the checkbox above the dataset table:

image

I'm open to any other suggestions.

subdavis (approved):

PR #852, opened 2021-07-20
Desktop environment paths and readonly mode

### LINKED ISSUE #640: [BUG] Always honor DIVE_VIAME_INSTALL_PATH environment path
If I extract the VIAME desktop installers to a directory, run dive, manually change the VIAME install path in settings then despite reloading the GUI with a new DIVE_VIAME_INSTALL_PATH it still retains the old manual setting.

(the install path probably shouldn't even be a setting when the env var is specified regardless because that's just gonna confuse users)

### LINKED ISSUE #727: [FEATURE] Unauthenticated mode

f4str:
fixes #640 
fixes #727 

Added environment overrides on desktop under the `DIVE_VIAME_INSTALL_PATH` and `DIVE_READONLY_MODE` environment variables. If they are provided, they will override the settings for `viamePath` and `readonlyMode` respectively. If the environment variable `DIVE_VIAME_INSTALL_PATH` is provided, it will also prevent the user from changing the path in any way:
image

Implemented read only mode on desktop. This can be either toggled on/off from the settings or the `DIVE_READONLY_MODE` environment variable can be provided. Even if this variable is provided, the user can still toggle it on or off. When read only mode is active, a warning will be displayed in the settings:
image

In read only mode, the user is able to make changes as usual, however, they cannot save any changes. The save icon will be disabled with a warning logo to catch attention:
image

When hovering over the save icon, it will display a descriptive message:
Kazam_screenshot_00000

However, the Run Pipeline and Import functionality still work (as requested) which can still modify the data even in read only mode. Although this is not good in terms of the application perspective, this is still the desired feature.

Fixed the issue with "fake" saving on the settings page on desktop. Previously when changing any settings (such as the viame path), the desktop app would update immediately before clicking the save option. When navigating away to a different page, these changes would persist. However, when reloading the app, these changes would not truly save until the save button was actually clicked. Now, when making local changes, nothing will update until clicking the save button. Navigating away to a different page will also discard any unsaved changes on the settings page. 

_Read only mode is only for desktop, it is not implemented for web._


client/platform/desktop/frontend/store/settings.ts
```diff
@@ -30,30 +30,38 @@ function isSettings(s: any): s is Settings {
...
-      }
+    // Client applies user-configured settings from localstorage
+    const settingsStr = window.localStorage.getItem(SettingsKey) || '{}';
+    const maybeSettings = JSON.parse(settingsStr);
+    if (isSettings(maybeSettings)) {
+      settingsValue = {
+        // Populate from defaults to include any missing properties
+        ...settingsValue,
+        // Overwrite with explicitly persisted settings
+        ...maybeSettings,
```
BryonLewis, line 47: This could be a two-part because the correct `localSettings` aren't written unless you save the data So see the other comment further down.
If you don't have a DIVE_VIAME_INSTALL_PATH set and have a previous version when you load up the data it will set override to `{}` in the local settings.
If you then set DIVE_VIAME_INSTALL_PATH afterwards and load it up, the locaStorage will always have the `overrides: {}`.  I'm guessing because your spreading the `...maybeSettings` over the top of `...settingsValue` so it is wiping out the default `overrides:` with the blank one.
Even if you get the correct `settings` set here it doesn't update the localSettings for the client.  I don't know if that is required during this part.

client/platform/desktop/frontend/store/settings.ts
```diff
@@ -30,30 +30,38 @@ function isSettings(s: any): s is Settings {
...
-
+  // Client applies external force overrides
+  if (settingsValue.overrides.viamePath !== undefined) {
+    settingsValue.viamePath = settingsValue.overrides.viamePath;
+  }
+  if (settingsValue.overrides.readonlyMode !== undefined) {
+    settingsValue.readonlyMode = settingsValue.overrides.readonlyMode;
+  }
+  settings.value = settingsValue;
   ipcRenderer.send('update-settings', settings.value);
```
BryonLewis, line 61: I don't know exactly if this would be a incorrect/correct option but if the settings are different than the base `localSettings` like the case where there is an environment variable.  Here is where I think you would want to just call `updateSettings` from further down to update the `localStorage` as well

client/platform/desktop/frontend/components/Settings.vue
```diff
@@ -69,23 +75,26 @@ export default defineComponent({
...
               />
             </v-col>
+
             <v-col cols="3">
               <v-btn
                 large
                 block
                 color="primary"
-                class="mb-6"
+                :disabled="!!(localSettings.overrides && localSettings.overrides.viamePath)"
```
BryonLewis, line 97: I think these two need some explicit text stating why they are disabled.  Something about the environment variable name being set and using that instead of manually allowing changing.

client/platform/desktop/frontend/components/Settings.vue
```diff
@@ -20,10 +21,12 @@ export default defineComponent({
...
+    // create a local copy of the global settings
+    const localSettings = cloneDeep(settings);
     const { arch, platform, version } = process;
-    const settingsAreValid = ref(true as boolean | string);
+    const settingsAreValid = ref(false as boolean | string);
     const gitHash = process.env.VUE_APP_GIT_HASH;
+    const readonlyMode = computed(() => settings.value.readonlyMode);
 
     onBeforeMount(async () => {
       settingsAreValid.value = await validateSettings(localSettings.value);
```
BryonLewis, line 32: Read the other comments first indicating that this may have `overrides: {}` even when an environment variable exists.
If the environment variable isn't a valid location should we still have this locked to prevent the user from manually changing it or should we indicate that they should adjust their environment variable to a valid location?

BryonLewis (changes requested):
Couple of things primarily related to adding environment variables after launching it without them.  Only other thing is either instructions or options if the install path is disabled as well as if it is disabled and the environment path is invalid.

client/platform/desktop/frontend/store/settings.ts
f4str: I was a bit confused what you meant at first, but after understanding this is a great catch. Really appreciate you pointing out this major issue. I was actually having trouble with getting the environment variables to load in and thought I was going insane, but this was actually the issue.

A quick temporary fix I'm trying for now is just to delete the `overrides` property from `maybeSettings` (since it's a JSON and not a proper type). Then spreading it on top of `settingsValue` will overwrite everything except the overrides. I'm open to any better solutions.

client/platform/desktop/frontend/components/Settings.vue
f4str: Good point. I'll add an info box saying the environment exists and changes are being locked.

client/platform/desktop/frontend/store/settings.ts
BryonLewis: I think you might want to check the `overrides` to ensure the `Object.keys(maybeSettings).length !== 0` or something similar before deleting it or setting it to undefined.  If there exists the path/readonly settings you probably want them to spread and overwrite the defaults.

f4str:
The save button is now disabled when there are no pending changes. This is just a simple computed ref checking if the local settings match the global settings.

image

Fixed the major issue where the `overrides: {}` property from local settings can overwrite the actual overrides property from environment variables. For now a simple solution is to delete the `overrides` property from the local settings json since that's the only problematic property. Now running
```bash
DIVE_VIAME_INSTALL_PATH=something yarn serve:electron
```
will properly cause the path to be set and will lock changes.

Added an info box to indicate the `DIVE_VIAME_INSTALL_PATH` environment variable is set and that changes are locked. For now its placed right under the kwiver initialization message, but putting on the top (before the GPU message) might be better. I'm open to any suggestions for placement and text.

image

f4str:
There are still a few outstanding issues
* As pointed out before, should the VIAME install path folder option still be disabled if the environment variable is provided but is a bad path
* Should the readonly mode switch be disabled if the environment variable is provided.
* There is a race condition between loading settings and initializing settings if reloading the electron app on the settings page (ctrl + R). Looks something like this
  image
  where the settings aren't initialized and everything is empty. It's fixed by going to a different page and back. This isn't caused by this PR and has always been an issue (same issue on the main branch), but should probably be addressed now rather than later.

client/platform/desktop/frontend/store/settings.ts
subdavis: Could we add a little comment above the `delete` line about what that's for?

subdavis (approved):
LGTM, tested and works quite well!

BryonLewis (approved):
Nice, looks good.

BryonLewis (approved):

client/platform/desktop/frontend/components/Settings.vue
```diff
@@ -33,8 +33,13 @@ export default defineComponent({
     const pendingChanges = computed(() => isEqual(localSettings.value, settings.value));
 
     onBeforeMount(async () => {
-      settingsAreValid.value = await validateSettings(localSettings.value);
       smi.value = await nvidiaSmi();
+      settingsAreValid.value = await validateSettings(localSettings.value);
```
subdavis, line 37: You flipped the order here.  Does this change anything?  If this order matters, you have a race condition that needs to be addressed.

subdavis (dismissed):
One quick question about a command reorder.

f4str:
After discussion, there is no longer a race condition and entry points of error in `Settings.vue` so the two lines can be swapped back to their original order.

Also updated the desktop settings screenshot for the docs.

subdavis (approved):

Issue #837, opened 2021-07-13
[FEATURE] "Shared with me" view on Data Browser page.


subdavis:
**Is your feature request related to a problem? If so, Please describe.**

Users are consistently frustrated when someone shares data with them but they can't find it.  They share stuff with eachother, but don't understand that you still have to navigate to it in the browser.

**Describe the solution you'd like**

A new tab that switches the data browser with a "Shared with me" view that is a flat list of all the datasets and folders shared with a user, just like the "Shared with me" feature in google drive.

I think it should be a list with dataset or folder name, checkboxes like the browser, and a column for who owns the resource.  This could be produced with a mongo aggregation pipeline query.  This feature will require a new endpoint in the `viame` resource.

![Screenshot from 2021-07-13 08-48-01](https://user-images.githubusercontent.com/4214172/125454769-19c6d136-1b40-4be7-91eb-1fe05581cc21.png)

## More details

* Create a new endpoint in the web server that returns a list of shared folders and datasets
  * This would be a mongo query for Girder folders where the user has read, write, or admin access and is not the owner.
  * It would be cool to add a query aggregation that joins the original owner's usernam as well so you can show the true owner on the page.
 * Above the data browser and sidebar in the Home.vue page, add tabs for "Browse data" and "Shared with me".  Don't use `v-tabs-items`, they add unsightly animation that won't look good on this page.  You can use `v-tab` but render the actual page elements with `v-if` based on the value of the tab model.
 * Switch out the data browser with some kind of list, maybe a `v-list dense` but also maybe a `v-data-table` with the results of th endpoint.

The sidebar will be visible in both views and should generally work the same for both.  Both components should support multi-select.

![Screenshot from 2021-07-14 11-05-34](https://user-images.githubusercontent.com/4214172/125646284-dd5a28d1-613f-4310-bbf9-5be1eb0b1260.png)

PR #861, opened 2021-07-27
**WIP** adds shared tab and data table

marySalvi:
Feature Development #837 

- Adds 'Browse Data' and 'Shared with Me' tabs
- Adds Data Table to display the 'Shared with Me' Data 
- Currently uses Datasets as data for testing purposes
- *** Does Need Work*** I am still working on trying to link the selected data from the Shared with me Data Table to the DataDetails Menu (as pictured below)
Screenshot from 2021-07-27 08-55-23


subdavis:
I think this is definitely headed in the right direction.

Couple recommendations:

- [ ] if a row is a dataset, put a button to launch the viewer same as in the data browser.
- [ ] format the file size as human readable.  There's a method to do this in the girder web components mixins, you can either use it in-template or in a computed function.
- [ ] I want to make shared by the user's name, which will require some backend changes, probably something to get to later.
- [ ] We can work on getting the selected boxes synced up if you want.

PR #887, opened 2021-08-13
Add arch diagram

subdavis:
For reference later today.


BryonLewis (approved):

aashish24:
the diagram is nice. Just wondering if we could have used the wording "girder server" as opposed to "girder REST API" as workers coming back to REST API may not makes sense or does it?

subdavis:
> as workers coming back to REST API may not makes sense or does it?

workers `POST` their results via the Girder REST interface, and the diagram attempts to capture that.  I chose "REST API" because "server" is sort of an overloaded term.  Does this answer your question, and is there a way we could better communicate that?

aashish24:
@subdavis that made sense. I think in general this arch diagram is good but just wanted to provide another perspective. For instance load balancer, mongodb kind of fall into physical entities where as REST API is more of a software (API) side of things but they are represented same in the diagram. So perhaps REST API is the technique how the messages are sent (the arrow between components). Also, where is the QUEUE hosted? Is that on the web server or a separate instance?

subdavis:
> For instance load balancer, mongodb kind of fall into physical entities where as REST API is more of a software (API) side of things but they are represented same in the diagram

I don't think I'd agree with that.  A load balancer, a database, and a REST api are all running processes that represent Layer 7 applications running inside containers.   Even if we were using some layer 4 proxy like HAProxy, I'd still probably add it to the diagram because it's relevant to how requests move through the system.

I agree it sorta fuzzy when you want to show where physical bytes are located (on disk? In S3?) because S3 is also a service but disk is a physical device that the operating system, but it's still useful for the diagram reader to understand where the bytes are coming from.  I left this part out for that reason.

> Also, where is the QUEUE hosted? Is that on the web server or a separate instance?

This diagram is for external folks to understand how DIVE works.  For them, the queue would be on the same host.  Our Queue is actually in AWS, but nobody outside of our team really needs to know that.

PR #895, opened 2021-08-16
Add sorting method dropdown for type list

### LINKED ISSUE #872: [FEATURE] option to alphabetize types in type list.

When there are a ton of features, an alphabetical list is better than sorting by count.

At some point in the past, types with no visible tracks used to disappear. What happend there?

Minious:
Fixes #872

Adds a dropdown which allows to sort the list of types by instance count or alphabetically.


subdavis:
Thanks for your PR.  I tweaked this a bit and so this PR will be replaced by #899

PR #902, opened 2021-08-24
adds alert message banner

### LINKED ISSUE #815: [FEATURE] Site-wide maintenance banner
To announce scheduled maintenance and other issues, we need to be able to put a banner at the top of the page for all logged in users.  Using the brand info metadata object seems like a reasonable option. 

Just having info on the login page means most users won't see the announcement (or notice it at all).

marySalvi:
Resolves #815 

- Adds site-wide banner if value for alertMessage field is present on brandData (currently using a static variable. will change)
- Adds an alert button and dialog modal on the viewer page in an attempt to mindful of over crowding the page.

Screenshot from 2021-08-23 23-29-12
Screenshot from 2021-08-23 23-28-45
Screenshot from 2021-08-23 23-28-54


client/platform/web-girder/api/viame.service.ts
```diff
@@ -25,6 +25,7 @@ export interface BrandData {
   logo: string;
   name: string;
   loginMessage: string;
+  alertMessage: string;
```
subdavis, line 28: Most of the time this value will be empty.

`alertMessage: string | null`  should be the type.

client/platform/web-girder/store/Brand.ts
```diff
@@ -31,9 +31,10 @@ const brandModule: Module<BrandState, RootState> = {
       favicon: null,
       logo: defaultLogo,
       name: 'DIVE',
-      loginMessage: `DIVE is automatically updated 
+      loginMessage: `DIVE is automatically updated
         at 2AM EST/EDT on Thursdays. Downtime is typically
         less than 10 minutes.`,
+      alertMessage: 'viame.kitware.com will be down for maintenance on 8/20/21',
```
subdavis, line 37: Should default to null.

client/platform/web-girder/views/NavigationBar.vue
```diff
@@ -42,49 +42,59 @@ export default {
...
+      <v-btn
+        text
+        @click="logout"
+      >
+        Logout
+      </v-btn>
+    </v-app-bar>
+    <v-alert
+      v-if="brandData.alertMessage"
+      type="info"
```
subdavis, line 89: Let's unify the color and icon across all places this alert may appear to warning yellow?

`type=warning`

Also, I know `v-banner` is specifically designed to fit below a `v-app-bar`.  Is there any difference between these, such as how the page layout is impacted?

client/platform/web-girder/views/ViewerAlert.vue
```diff
@@ -0,0 +1,60 @@
+<script lang="ts">
+import { defineComponent, ref, toRef } from '@vue/composition-api';
+
+export default defineComponent({
+  name: 'ViewerAlert',
+  setup(_, { root }) {
+    const dialog = ref(false);
+
+
```
subdavis, line 9: extra blank line.

client/platform/web-girder/views/ViewerAlert.vue
```diff
@@ -0,0 +1,60 @@
...
+<template>
+  <!-- This component is for alert messages on the Viewer Page only.
+    See NavigationBar for sitewide alert banner -->
+  <v-dialog
+    v-model="dialog"
+    max-width="600px"
+  >
+    <template v-slot:activator="{ on, attrs }">
+      <v-icon
+        v-if="brandData.alertMessage"
```
subdavis, line 30: move the `v-if` up to the `v-dialog` so that when alertMessage is undefined, the whole dialog is excluded, not just its activator.

client/platform/web-girder/views/ViewerAlert.vue
```diff
@@ -0,0 +1,60 @@
...
+      dialog,
+      brandData,
+    };
+  },
+});
+</script>
+
+<template>
+  <!-- This component is for alert messages on the Viewer Page only.
+    See NavigationBar for sitewide alert banner -->
```
subdavis, line 23: Remove comment?  I don't see anything inside this component that requires it only be rendered inside the Viewer context.

This comment is a maintenance burden because we have to remember to change it 1) if we want to use the component elsewhere or 2) if we change how NavigationBar works.

client/platform/web-girder/views/ViewerAlert.vue
```diff
@@ -0,0 +1,60 @@
+<script lang="ts">
+import { defineComponent, ref, toRef } from '@vue/composition-api';
+
+export default defineComponent({
+  name: 'ViewerAlert',
+  setup(_, { root }) {
+    const dialog = ref(false);
+
+
+    // this will change once REST updates are merged
```
subdavis, line 10: remove comment.

client/platform/web-girder/views/ViewerAlert.vue
```diff
@@ -0,0 +1,60 @@
...
+    </v-card>
+  </v-dialog>
+</template>
+
+<style lang='scss'>
+.theme--dark.v-dialog--active {
+  background: green;
+}
+
+</style>
```
subdavis, line 60: Generally best to avoid hacking into private CSS classes in vuetify as they make version upgrades difficult and require you to use unscoped (global) css.  These classes are not part of the Vuetify public API, so they may change without warning.

This block will also effect every `v-card` in the application because it's unscoped.

client/platform/web-girder/views/ViewerAlert.vue
```diff
@@ -0,0 +1,60 @@
...
+    <v-card
+      color="blue"
+      class="pa-4"
+    >
+      <v-icon
+        class="float-right"
+        @click="dialog = false"
+      >
+        mdi-close
+      </v-icon>
```
subdavis, line 49: use `<v-btn icon>` to wrap clickable icons.  They appear in the tab select index and give a cursor:pointer when hovered.

client/platform/web-girder/views/ViewerAlert.vue
```diff
@@ -0,0 +1,60 @@
...
+  <v-dialog
+    v-model="dialog"
+    max-width="600px"
+  >
+    <template v-slot:activator="{ on, attrs }">
+      <v-icon
+        v-if="brandData.alertMessage"
+        dark
+        large
+        color="red"
```
subdavis, line 33: to use the same colors everywhere, change to `color="warning"`?

client/platform/web-girder/views/ViewerAlert.vue
```diff
@@ -0,0 +1,60 @@
...
+        v-bind="attrs"
+        v-on="on"
+      >
+        mdi-alert-circle
+      </v-icon>
+    </template>
+    <v-card
+      color="blue"
```
subdavis, line 41: either warning color here too, or maybe just `outlined` with no color?  Having a red warning button open a blue info dialog seems busy to me.

client/platform/web-girder/views/ViewerAlert.vue
```diff
@@ -0,0 +1,60 @@
...
+  },
+});
+</script>
+
+<template>
+  <!-- This component is for alert messages on the Viewer Page only.
+    See NavigationBar for sitewide alert banner -->
+  <v-dialog
+    v-model="dialog"
+    max-width="600px"
```
subdavis, line 26: In most of the other dialogs, we've added `overlay-opacity="0.90"` to make the background fade more while the dialog is open.  Could you add that here as well?

client/platform/web-girder/views/ViewerLoader.vue
```diff
@@ -86,6 +88,7 @@ export default defineComponent({
     ref="viewerRef"
   >
     <template #title>
+      <ViewerAlert />
```
subdavis, line 91: I think this might need to move down after the `v-tabs` so that the tab's horizontal position doesn't jump when navigating between data browser and viewer.  What do you think?

subdavis (changes requested):
Handfull of small things.

After playing with it, I like the icon on the viewer page.

Thanks!

client/platform/web-girder/api/viame.service.ts
marySalvi: Should this still be `string | null` or should it just be optional?

client/platform/web-girder/views/NavigationBar.vue
marySalvi: I am not easily finding an answer to the page layout impact. I suspect that since it looks like Banner was created after Alert that this might be an added feature. I am fine with going with the component known to work with our layout and will change it to `v-banner`

client/platform/web-girder/views/ViewerLoader.vue
marySalvi: I personally prefer it before before the tabs for a couple of reasons:
1. I think it gets lost a bit when its at the end of the tabs
2. The placement at the end makes me think the warning is associated with the image/video being viewed.

I am fine with whatever everyone thinks is best though

Screenshot from 2021-09-22 12-10-16
Screenshot from 2021-09-22 12-09-46
.

subdavis (changes requested):
Looks like something may have happened to this branch.  Diff is showing +3900 -2700 which isn't right.

marySalvi:
@subdavis I think merge issues with this branch have been fixed.

subdavis (approved):
Looks good!

One thing I had to change: somewhere along the way `alertMessage` got dropped from `BrandData`, so if you tried to enable it, you got an error:

Screenshot from 2021-10-04 15-40-22

I assume you were testing this feature by adding the string to the client state?  It's generally a good idea to test the backend integration for features that require it, because even little things like this can be missed.

:shipit:

PR #908, opened 2021-08-29
Fix Cancellation

subdavis:
While reviewing on friday, I noticed there were some unhandled cancellation conditions, most notably in `check_and_fix_frame_alignment()`.  I implemented cancellation poorly to begin with such that any function call that could interrupt the function has to be checked on return (in the same way that golang error handling works) so you have to check all the way down the call stack.

This is the whole purpose of built-in error handling language features (throw early/catch late).  This gets rid of most instances of `check_canceled` so that forgetting handle the cancel somewhere won't introduce bugs.

It also replaces `self` with `task` in some functions.  Self shouldn't be the name of an argument in a function that isn't an instance method (celery tasks are a special case)

This is untested (I wasn't done with it) but I wanted to go ahead and let you know about the possible change because it looks like you were doing stuff.

### DIFF
--- a/server/dive_tasks/frame_alignment.py
+++ b/server/dive_tasks/frame_alignment.py
@@ -9,10 +9,12 @@ from typing import Dict
 from girder_worker.task import Task
 from girder_worker.utils import JobManager
 
-from dive_tasks.utils import check_canceled, stream_subprocess
+from dive_tasks.utils import stream_subprocess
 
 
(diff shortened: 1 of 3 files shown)


subdavis:
Summary of changes

* stream_subprocess now throws a CanceledError, so @task methods must catch and swallow this for any calls to this method or methods that might call it.

That's pretty much it.

BryonLewis (approved):

PR #910, opened 2021-08-30
Fixes #903 Import annotation error handling

### LINKED ISSUE #903: [BUG] Import Annotations Button has no error handling
Upload a junk CSV or json using the import button

Result: spins forever, no error message.

subdavis:
fixes #903

Until now, apparently the client would let you upload whatever json you wanted and postprocess would tag it as an annotation file without checking the formatting.  Now, it's forced through a pydantic schema validation same as if you `save`.

Also fixes a bug where `api.importAnnotationFile` was not updated to use the new postprocessing endpoint.


BryonLewis (approved):
:+1:

PR #916, opened 2021-09-02
Updates for docker compose

subdavis:
* The girder health check is worthless and spams the access log with 90% empty noise, making logs rather hard to look at.
* Traefik was missing restart:always, so the service doesn't come back up when the server restarts, which was an issue today.


BryonLewis (approved):
Sorry it took me a bit, I was double checking and looking up what exactly the worker needed from the girder container and why we were reliant on the server being responsive before.

PR #917, opened 2021-09-03
Upgrade vue, template compiler, composition api

subdavis:
Will require a minor version bump for consumers of `vue-media-annotator`


BryonLewis (approved):
served, built and rand through some basic tests.  Everything that I tested seemed to function properly.

PR #944, opened 2021-09-22
Allow raw FPS on the client

subdavis:
Remove FPS floor on ingest, allow user to choose native FPS.


BryonLewis (approved):
tested with 29.97... frame embedded video and it works properly.
This will deviate from the Web version in which we probably need to add an option that says (annotate at video frame rate) considering we don't know the Frame rate until ffprobe processes it.

PR #985, opened 2021-10-11
fixes TypeError on delete

marySalvi:
The error message on delete is happening because
` this.$refs.fileManager.$refs.girderBrowser.refresh();`
is throwing an undefined TypeError.

Current fix just ignores that error but I am going to investigate the cause of it further.


marySalvi:
Undefined $ref error fixed by adding a ref to the `router-view`

also needed to update to use the store for setSelection

Issue #600, opened 2021-02-23
[FEATURE] Support KWIVER pipelines for performing video re-encoding and image enhancement on upload


mattdawkins:
Some people might want to enhance their videos/images when being uploaded into the system, so we probably want a check-mark for this. Example KWIVER pipeline: filter_enhance.pipe

subdavis:
@mattdawkins can you describe the kwiver options/command necessary for running this in a pipeline?

This issue needs detailed requirements.

* Client side enhancement options?
* pipeline enhancement options?

mattdawkins:
There's a couple pipes which could apply here beginning with the filter_* prefix (so maybe another wildcard option would make sense)

filter_extract_frames.pipe  - Just extracts frames using KWIVER downsampler no enhancement
filter_enhance.pipe - Runs image enhancement and outputs image list
filter_extract_left.pipe - Crops out left half of image in case stereo composites
filter_debayer_and_enhancepipe - Runs debayering on input camera raws and enhancement

Inputs are the same as every other pipe covered so far, outputs are as follows for 6 0 padded #s:

-s output:file_name_template=great_folder/frame%06d.png

There's a few which can be set via parameters which are fixed for the job, though probably lesser priority

  :filter:ocv_enhancer:apply_smoothing         false
  :filter:ocv_enhancer:smooth_kernel           3
  :filter:ocv_enhancer:apply_denoising         false
  :filter:ocv_enhancer:denoise_kernel          3
  :filter:ocv_enhancer:denoise_coeff           2
  :filter:ocv_enhancer:auto_balance            true
  :filter:ocv_enhancer:force_8bit              false
  :filter:ocv_enhancer:apply_clahe             true
  :filter:ocv_enhancer:clip_limit              3
  :filter:ocv_enhancer:saturation              1.20

Out of these saturation, float [ 0, 10.0 ] might be most useful and clip limit (int, [ 0, 10 ])

waxlamp:
@subdavis do you have the information you need for this?

PR #1015, opened 2021-10-25
Add brightness prop and svg filter

AlmightyYakob:
This arose from a request in a downstream project. I've tested this with `ImageAnnotator.vue` and it works as expected. I haven't tested it with `VideoAnnotator.vue`, so I can't explicitly verify that functionality, although I'd expect it to work just the same.

I also realize this might not be in line with the existing coding style / organization, so please point out / suggest these fixes if they apply.

related to #600


subdavis (dismissed):
This doesn't seem to work for monochromatic images.

BryonLewis (approved):
I've tested both video and image sequence and it seems to operate as intended where only the media layer has the filter applied.  I think this accomplishes what you need without influencing any of the core DIVE functionality.

PR #1023, opened 2021-11-01
RE: Add brightness prop and svg filter

AlmightyYakob:
This is a redo of #1015, since the commits from that were merged but not squashed.


subdavis (approved):
Rubber stamp approval based on #1015

PR #1026, opened 2021-11-02
Replace all uses of `vue-media-annotator` alias under src/ with relative imports

AlmightyYakob:
When using `vue-media-annotator` externally, attempting to import anything from `src/layers/TextLayer.ts` (in my case the `TextData` type), the following error is thrown:

```
1:29 Cannot find module 'vue-media-annotator/use/useStyling' or its corresponding type declarations.
  > 1 | import { TypeStyling } from 'vue-media-annotator/use/useStyling';
      |                             ^
    2 | import BaseLayer, { BaseLayerParams, LayerStyle } from './BaseLayer';
    3 | import { FrameDataTrack } from './LayerTypes';
    4 | 
```

I suspect this is due to the fact that in the project itself, `vue-media-annotator` is an alias to the `src` directory.
This PR just replaces that with the use of a relative path.

### DIFF
--- a/client/src/components/TrackList.vue
+++ b/client/src/components/TrackList.vue
@@ -4,8 +4,8 @@ import {
   defineComponent, reactive, computed, ref, Ref, watch,
 } from '@vue/composition-api';
 
-import { TrackWithContext } from 'vue-media-annotator/use/useTrackFilters';
 import { usePrompt } from 'dive-common/vue-utilities/prompt-service';
+import { TrackWithContext } from '../use/useTrackFilters';
 
 import { TrackId } from '../track';
 import {
--- a/client/src/components/TypeList.vue
+++ b/client/src/components/TypeList.vue
@@ -3,11 +3,11 @@ import {
   computed, defineComponent, reactive, Ref,
 } from '@vue/composition-api';
 
+import { usePrompt } from 'dive-common/vue-utilities/prompt-service';
 import {
   useCheckedTypes, useAllTypes, useTypeStyling, useHandler,
   useUsedTypes, useFilteredTracks, useConfidenceFilters,
-} from 'vue-media-annotator/provides';
-import { usePrompt } from 'dive-common/vue-utilities/prompt-service';
+} from '../provides';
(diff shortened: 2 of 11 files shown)


subdavis:
Can you find anywhere else "vue-media-annotator" is used inside src/ ?

There used to be something in the readme explaining this, but I guess this slipped in.

AlmightyYakob:
> Can you find anywhere else "vue-media-annotator" is used inside src/ ?
> 
> There used to be something in the readme explaining this, but I guess this slipped in.

Sure, meaning if I find anymore occurances of this, replace them with relative imports?

subdavis:
Yep

client/src/components/TrackList.vue
```diff
@@ -4,8 +4,8 @@ import {
   defineComponent, reactive, computed, ref, Ref, watch,
 } from '@vue/composition-api';
 
-import { TrackWithContext } from 'vue-media-annotator/use/useTrackFilters';
 import { usePrompt } from 'dive-common/vue-utilities/prompt-service';
```
subdavis, line 7: FYI: Because this module still imports prompt service, this component is unusable.

client/src/components/TypeList.vue
```diff
@@ -3,11 +3,11 @@ import {
   computed, defineComponent, reactive, Ref,
 } from '@vue/composition-api';
 
+import { usePrompt } from 'dive-common/vue-utilities/prompt-service';
```
subdavis, line 6: Unusable.

subdavis (approved):

waxlamp:
> Can you find anywhere else "vue-media-annotator" is used inside src/ ?
> 
> There used to be something in the readme explaining this, but I guess this slipped in.

Do you mean there's a reason why we'd make `vue-media-annotator` an alias for `src/` and that reason has been lost?

subdavis:
No.  vue-media-annotator is used to alias this library from other packages (platform, dive-common).  But the code in src/ can't use this alias itself because the alias will get bundled into the lib, and then when someone installs it from NPM, that alias points nowhere.

PR #1029, opened 2021-11-03
Fix #1028

### LINKED ISSUE #1028: [BUG] Very large upload requests return err 431
**Describe the bug**
GET validate_Files returns https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/431

**To Reproduce**
1. Try to pload a few thousand imags
2. see bug

subdavis:
Switch to POST for file validation so that params can be sent as json body.  

Read https://stackoverflow.com/questions/978061/http-get-with-request-body

Apparently elastic search uses GET with body, but cherrypy didn't like this, and it's not totally good with the HTTP spec.  POST is harmless here.

fixes #1028

### DIFF
--- a/client/platform/web-girder/api/dataset.service.ts
+++ b/client/platform/web-girder/api/dataset.service.ts
@@ -127,9 +127,7 @@ interface ValidationResponse {
 }
 
 function validateUploadGroup(names: string[]) {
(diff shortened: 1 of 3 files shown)


BryonLewis (approved):
pulled, built, tested various size uploads and it worked.
Also ran integration tests successfully.

PR #1034, opened 2021-11-05
Client/type search 1003

### LINKED ISSUE #1003: Enable bulk toggling of ground truth annotations
There may be ground truth or other kinds of associated data belonging to the same track type, etc. It is currently possible to group some activity types together and turn them on and off, but we also need a way to toggle the subtypes as a group as well. That is, "row-wise" toggling is already possible; we also want "column-wise" toggling.

Some way to accomplish "column-wise" toggling.

A filter on activity name could be used, as long as the data encodes activity subtype in regex-compliant ways.

subdavis:
https://user-images.githubusercontent.com/4214172/140565388-0c2cef1f-3360-47f8-ab3d-4eed9e3a6265.mp4

fixes #1003

### DIFF
--- a/client/src/components/TrackItem.vue
+++ b/client/src/components/TrackItem.vue
@@ -276,7 +276,7 @@ export default defineComponent({
         v-model="data.trackTypeValue"
         type="text"
         list="allTypesOptions"
-        class="input-box"
+        class="input-box freeform-input"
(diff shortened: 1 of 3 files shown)


BryonLewis (approved):
Tested and looks good.

PR #1037, opened 2021-11-09
Remove MEVA KPF support

### LINKED ISSUE #246: Run KPF processing in worker task
Should run KPF generation in a worker task cause it's hefty.

This actually gets into a "race" condition if a video is uploaded with KPF: the video and the KPF processing must complete before the dataset can be marked for annotation.  How do we handle this?  

CC @AlmightyYakob and @mattrix27

subdavis:
Drops KPF support for upload.  This current implementation has no users.

We may have to reimplement at some point, but we should do so from scratch rather than having a diva-boiler dependency.  Our implementation has no tests and is not compatible with the features needed.

Closes #246.


subdavis:
Still interested in dropping KPF, but this PR needs a rebase.

subdavis:
Stale PR, closing.  Can revisit later if we want.

PR #1044, opened 2021-11-11
Aggregate dependabot updates

subdavis:


BryonLewis (approved):
pulled and did some standard client (web/desktop) testing, seems good.

PR #1060, opened 2021-11-18
Fix customTypeStyling key name

subdavis:


BryonLewis (approved):
I could have sworn I did a direct import from the web version configuration file into the desktop with specifically a style having the fill turned on with opacity and it reflected properly in desktop.  I actually did it two ways, once with the config.json and then obviously with the meta.json file (how I noticed the merge issue).  I didn't re-import all the same stuff after the `pick()` change though and that is where I missed it.

PR #1067, opened 2021-11-22
Mongo 2: Rollback API

subdavis:
## Reviewer notes

This could have been included in #1066 but it's a new feature, so I made it separate.  It provides a read API for rolling back to earlier versions.  

## Important

This implementation is naive and I don't want to keep it forever.  It's a true rollback such that it actually deletes rows from the DB.  I would prefer a soft rollback like `git revert` which replays changes in reverse.  But that sounded complicated and I didn't feel like writing it.

### DIFF
--- a/server/dive_server/crud_annotation.py
+++ b/server/dive_server/crud_annotation.py
@@ -79,6 +79,18 @@ def get_revisions(
     return cursor, total
 
 
+def rollback(dsFolder: types.GirderModel, revision: int):
+    """Reset to previous revision."""
+    # TODO implement immutabble forward-rollback (like git revert)
+    # Logic: delete everything created after revision
+    # And erase deletions for anything deleted after revision
+    RevisionLogItem().removeWithQuery({REVISION: {'$gt': revision}})
+    AnnotationItem().removeWithQuery({REVISION_CREATED: {'$gt': revision}})
+    AnnotationItem().update(
+        {REVISION_DELETED: {'$gt': revision}}, {'$unset': {REVISION_DELETED: ""}}
+    )
+
+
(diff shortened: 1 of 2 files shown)


BryonLewis (approved):
Much simpler with the re-base.  I pulled and tested it and it seems to work properly removing the annotations after the revision number.

The only thing I would comment on is that it's a little unclear that when you give a DELETE with Revision = 5 - it really means DELETE all those revisions that are > 5 and not remove everything including 5.  I think as long as it's clear in the documentation that should be okay though.

waxlamp:
> Much simpler with the re-base. I pulled and tested it and it seems to work properly removing the annotations after the revision number.
> 
> The only thing I would comment on is that it's a little unclear that when you give a DELETE with Revision = 5 - it really means DELETE all those revisions that are > 5 and not remove everything including 5. I think as long as it's clear in the documentation that should be okay though.

I'm uneasy with this being a DELETE operation at all, because it's not really (in the REST or CRUD sense). However, since this is a tentative implementation with a long-term aim of a replacement, non-destructive rollback mechanism, it's probably ok.

The main difficulty is that the clients will need to change out DELETE calls for a PATCH or POST call later. To the extent that performing a rollback is a *state change* from the user's POV, rather than a *change in the model itself*, a POST operation with a verb like `rollback` sounds like the right abstraction to me; that would also avoid the need for clients to change their calling sequences later. (There's a name for this supplementation of CRUD operations with arbitrary (English) verb-based actions, but I can't find it right now; it's related to DRF's use of the [`@action` decorator](https://www.django-rest-framework.org/api-guide/viewsets/#marking-extra-actions-for-routing) in the same context.)

Issue #1065, opened 2021-11-18
[BUG] Dataset summary permissions error


subdavis:
**Describe the bug**
After the summary job is executed, the accessControlModel for each summary object should be set to public when `create()` is called.

Regular users have **never** been able to see the summary page, but it's never been reported.  I modified the database to fix for now, but the code needs to be changed

subdavis:
Summary feature was removed in the mongo track thing.

PR #1071, opened 2021-11-22
Mongo 5: Remove broken summary and report generation

subdavis:
Drops a whole heap of untested and unused code.  I found out in #1065 that users have never been able to see the summary page, and nobody ever mentioned it, so this drops the page, the report generation, and the Max-N csv generation that was also broken for a long time, untested, and seemed to be of dubious value.

We should have a follow-up discussion

* about how to replace summary with a version that Matt would find useful.
* about what kinds of report generation we actually need rather than having an unneeded demo report.


waxlamp:
> Drops a whole heap of untested and unused code. I found out in #1065 that users have never been able to see the summary page, and nobody ever mentioned it, so this drops the page, the report generation, and the Max-N csv generation that was also broken for a long time, untested, and seemed to be of dubious value.
> 
> We should have a follow-up discussion
> 
> * about how to replace summary with a version that Matt would find useful.
> * about what kinds of report generation we actually need rather than having an unneeded demo report.

I'll need quite a bit of background on what this was supposed to be, what kinds of problems it was meant to solve, etc. Let's discuss it the next time we have a planning meeting.

BryonLewis (approved):
Pulled, tested and ran integration tests.   Looks like all instances of the summary stuff were removed properly.  Assuming that we may add something back in once the switch to mongo is done, probably using your aggregation pipline stuff you were working on the other day.

PR #1081, opened 2021-11-30
update girder to current viable version

marySalvi:
Updating the girder version fixes our build.


BryonLewis (approved):
Pulled, built and ran integration tests which all passed.

PR #1087, opened 2021-12-08
Poetry fix

subdavis:
The virtual env location changed, but the static site wasn't moved. 

Otherwise, the frontend is missing.

Screenshot from 2021-12-08 14-37-21


BryonLewis (approved):
Okay I pulled and re-ran integration test, plus uploaded a video, transcoded and ran a pipeline this time.

PR #1092, opened 2021-12-17
Add Documentation for Labels.txt

subdavis:
I added some docs and added what I think are better labels and hints in the form.

If there are any issues here, I think they're on the viame side.


marySalvi (approved):
LGTM. Thank you for investigating and clarifying the labels.txt file usage.

PR #1098, opened 2021-12-28
Abandon ldc

subdavis:
While I still like LDC as an idea, and I would probably prefer to use it myself, everyone who gets started using this project seems confused by it.  It also protects the user from having to learn certain things about how docker-compose works, which is a drawback.

@erotemic recently made a PR https://github.com/Kitware/ldc/pull/2 that sorta forced a decision here: Keep supporting LDC as a clever 50 line shell script or just do something more familiar to [Girder 4](https://github.com/girder/cookiecutter-girder-4/tree/master/%7B%7B%20cookiecutter.project_slug%20%7D%7D) with an override file.

I think we should abandon ldc.

In order to make the transition nice and smooth, I recommend the following shell aliases.  Once these are in place, you shouldn't notice much difference from ldc

``` bash
alias dc="docker-compose";
alias ldc="docker-compose";
```

## Migration guide

The behavior of `ldc up -d` has been removed.  `ldc dev up -d` becomes the default because of the override file.  The thing I really liked about `ldc` was that it allowed you to explicitly and _conveniently_ differentiate between "run with the code bundled at build time" and "run with code mounted at runtime".  This distinction has never been as clear with vanilla docker compose, but it's probably better for users to just understand how it works than to have LDC protecting you.

```
# before (run with code bundled at build time)
ldc up -d
# after (sorta no longer needed)
dc -f docker-compose.yml up -d
# or, to use published pre-built images
dc pull && dc up -d

# before (run with code mounted at runtime)
ldc dev up
# after
dc up -d


# before (run a single service and mount code)
ldc dev up girder
# after
dc up -d girder
```


BryonLewis (approved):
pulled, built, ran integration test, ran some development testing and all seems to work well.  I just need to remind myself that ./server is hotload by default all the time so if I do an `up -d` and start messing around I can kill the container with errors during a save.

PR #1099, opened 2021-12-28
Arbitrary depth support for zip imports

subdavis:
* Changes a bit of code related to import dataset folder discovery to support arbitrary depths and cases where the zipfile is constructed in an unusual way.  Sometimes `filenames` doesn't list the directory as its own item, which was breaking stuff.  See this comment for a better description of what was broken: https://github.com/Kitware/dive/pull/1062#pullrequestreview-836459161
* Changes the integration test to parameterize the dataset rather than the user, so it's immediately clear what test case failed.
* Replaces the early return statement in `postprocess` because it makes things easier and you can leave off the "cleanup" routine in the celery task.  The server should be smart enough to detect a zip upload and not go around creating empty annotation files and aux folders when it knows that it's about to perform an import.

Now, even if there's an upload where a zip file contains multiple datasets at `a/b/c/d` and `a/1/2`, things should work out properly.

### DIFF
--- a/server/dive_server/crud_rpc.py
+++ b/server/dive_server/crud_rpc.py
@@ -374,7 +374,6 @@ def postprocess(
         Conversion of CSV annotations into track JSON
     """
     job_is_private = user.get(constants.UserPrivateQueueEnabledMarker, False)
-    auxiliary = crud.get_or_create_auxiliary_folder(dsFolder, user)
     isClone = fromMeta(dsFolder, constants.ForeignMediaIdMarker, None) is not None
     # add default confidence filter threshold to folder metadata
     dsFolder['meta'][constants.ConfidenceFiltersMarker] = {'default': 0.1}
@@ -413,6 +412,7 @@ def postprocess(
             )
             newjob.job[constants.JOBCONST_PRIVATE_QUEUE] = job_is_private
             Job().save(newjob.job)
+            return dsFolder
 
         # transcode VIDEO if necessary
         videoItems = Folder().childItems(
@@ -470,6 +470,7 @@ def postprocess(
             allFiles = [make_file_generator(item) for item in ymlItems]
             crud.saveTracks(dsFolder, meva.load_kpf_as_tracks(allFiles), user)
             ymlItems.rewind()
+            auxiliary = crud.get_or_create_auxiliary_folder(dsFolder, user)
             for item in ymlItems:
                 Item().move(item, auxiliary)
 
--- a/server/dive_tasks/tasks.py
+++ b/server/dive_tasks/tasks.py
@@ -622,9 +622,12 @@ def extract_zip(self: Task, folderId: str, itemId: str):
             "PUT",
             f"/item/{str(item['_id'])}?folderId={str(created_folder['_id'])}",
         )
-
+        # Only make subfolders if more than 1 discovered folder exists
+        make_subfolders = (
+            len(discovered_folders) - list(discovered_folders.values()).count('ignored')
+        ) > 1
         for folderName, folderType in discovered_folders.items():
-            subFolderName = folderName if len(discovered_folders) > 1 else ''
+            subFolderName = folderName if make_subfolders else ''
             if folderType == 'unstructured':
                 utils.upload_zipped_flat_media_files(
                     gc,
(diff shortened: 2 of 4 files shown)


server/dive_tasks/tasks.py
```diff
@@ -611,46 +623,23 @@ def extract_zip(self: Task, folderId: str, itemId: str):
...
-                gc.delete(f"folder/{aux_folders[0]['_id']}")
-            # remove metadata
-            metdata = [constants.TypeMarker, constants.FPSMarker, constants.DatasetMarker]
-            gc.sendRestRequest(
-                "DELETE",
-                f"folder/{folderId}/metadata",
-                json=metdata,
-            )
+        for folderName, folderType in discovered_folders.items():
+            subFolderName = folderName if len(discovered_folders) > 1 else ''
```
BryonLewis, line 627: The removal of the check for a single top level folder means that it imports anything from the exported Zip from the website as a subFolder instead of directly importing it in. I.E. - if you export a dataset as zip using the export all and then try importing it. Instead of treating it as the ./Import_name, it will treat it as ./Import_name/dataset_name meaning that it will always import the exported zips as a sub folder of the import. For multiple sub folders this makes sense but I'm not sure if we want this behavior for a single export zip dataset.  I realize this update now allows arbitrary depth of the imports, but at the cost of a smooth export/import procedure for the zip exports.  Just want to bring that up and see if it's worth it.

The idea would be to do a sum of the number of 'unstructured' and 'datatset' folders and if it is 1 the `subFolderName = ''` else it sets it to the current subfolder name?

server/dive_tasks/tasks.py
```diff
@@ -611,46 +623,23 @@ def extract_zip(self: Task, folderId: str, itemId: str):
...
-            )
-            if len(aux_folders) > 0:
-                gc.delete(f"folder/{aux_folders[0]['_id']}")
-            # remove metadata
-            metdata = [constants.TypeMarker, constants.FPSMarker, constants.DatasetMarker]
-            gc.sendRestRequest(
-                "DELETE",
-                f"folder/{folderId}/metadata",
-                json=metdata,
-            )
```
BryonLewis, line 656: The removal of this means that top level folders have residual metadata that stick around including `fps` and a `type: zip`.  The type may be fine as long as we don't ever use `type` for searching for data.  The `fps` is relative for creating a single dataset import but is a little weird if you have multiple zip files that are being uploaded.  Especially because if you set it to the video default rate the fps value will be -1.  This may be okay and intended though.

server/tests/integration/conftest.py
```diff
@@ -147,99 +147,97 @@
...
+            'name': 'testVideoZip',
+            'path': 'zipTestFiles/testVideoZip.zip',
+            'fps': 29.97002997002997,
+            'originalFps': 30000 / 1001,
+            'type': 'video',
+            'job_status': JobStatus.SUCCESS,
+        },
+        {
+            'name': 'singleDatasetImport',
+            'path': 'zipTestFiles/singleDatasetImport.zip',
```
BryonLewis, line 175: mistake on my part `SingleDatasetImport` not `singleDatasetImport`.  I should probably be more consistent in my zip file names (maybe the idea was to keep dataset imports capitalized in the beginning).

BryonLewis (changes requested):
Thanks for making this work with arbitrary levels.  I have a few questions and a comment based on using direct zip exports as imports without other datasets included.

server/dive_server/crud_rpc.py
```diff
@@ -413,6 +412,7 @@ def postprocess(
             )
             newjob.job[constants.JOBCONST_PRIVATE_QUEUE] = job_is_private
             Job().save(newjob.job)
+            return dsFolder
```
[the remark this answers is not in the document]
subdavis: Yes, see https://github.com/Kitware/dive/pull/1099#issue-1090108584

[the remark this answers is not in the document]
BryonLewis: sorry about that, I must of not internalized that before I started reading into the code.

server/dive_tasks/tasks.py
subdavis: Done.

server/dive_tasks/tasks.py
subdavis: Done.

server/tests/integration/conftest.py
subdavis: Done

BryonLewis (approved):

PR #1100, opened 2021-12-28
Resolve setuptools issue

subdavis:
Related issues

* https://github.com/PyCQA/isort/pull/1875
* https://github.com/pypa/virtualenv/issues/2257
* https://github.com/pre-commit/pre-commit/issues/2178

### DIFF
--- a/server/pyproject.toml
+++ b/server/pyproject.toml
@@ -81,5 +81,5 @@ rabbit_user_queues = "rabbitmq_user_queues:GirderPlugin"
 dive_tasks = "dive_tasks:DIVEPlugin"
 
 [build-system]
-requires = ["poetry-core>=1.0.0"]
+requires = ["poetry-core>=1.0.0", "setuptools"]
 build-backend = "poetry.core.masonry.api"


server/pyproject.toml
```diff
@@ -81,5 +81,5 @@ rabbit_user_queues = "rabbitmq_user_queues:GirderPlugin"
 dive_tasks = "dive_tasks:DIVEPlugin"
 
 [build-system]
-requires = ["poetry-core>=1.0.0"]
+requires = ["poetry-core>=1.0.0", "setuptools"]
```
waxlamp, line 84: Would it be better to specify a [compatible version](https://www.python.org/dev/peps/pep-0440/#compatible-release)?

waxlamp (approved):
Looks fine. Approving, despite also leaving a suggestion.

server/pyproject.toml
subdavis: All future versions of setuptools should be compatible. I don't know how I'd restrict this version, because I don't want to use an outdated version or prevent upgrade in the future.

PR #1104, opened 2022-01-03
Migrate track data to MongoDB

### LINKED ISSUE #306: [FEATURE] [Housekeeping] Move track data into Mongo
Right now, we store track json as a flat file in the asset store.

While it's quick to stream over the network, it's slow for update/delete operations because you have to load the whole thing into memory and parse it first.

What if we moved the track JSON into mongo, either through a custom Girder Model or directly in folder metadata.

This would also enable us to do much more powerful search/aggregation on the track data.  We could easily generate count summaries, lengths, and other statistics of the sort Matt has asked for.

We could keep doing what we're doing.  I don't care so much about the performance optimization, but I wonder if we could have better confidence in the integrity of the data if it lived in Mongo.

subdavis:
This PR keeps track of conflicts between main and the long-running feature branch.

This PR is made up of 5 sub-branches.

* #1066 
* #1067 
* #1068 
* #1069 
* #1071 

fixes #306


subdavis:
I think this is ready for review?

waxlamp:
> I think this is ready for review?

The review process will be a bit different here, I think.

@BryonLewis should lead the review process. If he is happy with all the constituent PRs, then in theory he should be ok with the concatenation thereof, but that's his determination.

I think the only thing remaining is to do a bunch of smoke testing locally to inspire confidence that everything is hanging together properly.

Thoughts on that? Anything I missed?

BryonCLewis:
I think we are waiting because Brandon may want to add in some more integration tests for the revisioning system and a few other things.  As soon as there is an update to the tests I'll take a final look.

BryonLewis (approved):
I looked over the last few Commits and it looks good to me so far.  I pulled ran some integration testing and did some other minor tests.

While giving approval I'm guessing we want to schedule the actual merge to main and deployment.

PR #1136, opened 2022-01-24
Fix #1124

### LINKED ISSUE #1124: Pad timestamps in output file formats (e.g. viame_csv) when necessary
DIVE is currently outputting some timestamps that look like   0:1:0.000, ideally the first and second numbers should be zero padded with at least one digit so it looks like 00:01:0.000 for consistency with other apps

Pad timestamps in first two numbers to 2 digits when necessary

subdavis:
fixes #1124

## Other minor changes

* makes upload dialog consistent with `dense` and `persistent-hint`
* Adds some useful hint description
* Formats job metadata with pretty indenting so the JSON is human readable.

### DIFF
--- a/client/platform/desktop/backend/serializers/viame.ts
+++ b/client/platform/desktop/backend/serializers/viame.ts
@@ -386,7 +386,7 @@ async function serialize(
             if (meta.type === 'image-sequence') {
               column2 = meta.originalImageFiles[feature.frame];
             } else if (meta.type === 'video') {
-              column2 = moment.utc(feature.frame / meta.fps).format('HH:mm:ss.SSS');
+              column2 = moment.utc((feature.frame / meta.fps) * 1000).format('HH:mm:ss.SSSSSS');
             }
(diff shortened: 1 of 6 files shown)


client/platform/desktop/backend/serializers/viame.ts
```diff
@@ -386,7 +386,7 @@ async function serialize(
             if (meta.type === 'image-sequence') {
               column2 = meta.originalImageFiles[feature.frame];
             } else if (meta.type === 'video') {
-              column2 = moment.utc(feature.frame / meta.fps).format('H:M:S.SSS');
+              column2 = moment.utc(feature.frame / meta.fps).format('HH:mm:ss.SSS');
```
subdavis, line 389: Notice the terrifying bug here.   `M` instead of `m`  indicates that until now, this was outputting the MONTH instead of the minute.

BryonLewis (approved):

PR #1151, opened 2022-02-03
Update documentation

subdavis:
fixes #1148 

blocked by #1147

PR #1153, opened 2022-02-07
Add extra comment about train execution time

mattdawkins:


subdavis (changes requested):
Should reference the appropriate documentation sections.

* Don't reference GCP directly, and instead just point to the deployment guide.
* We're moving to Discourse for support, so just referencing support in the docs would be better than saying the "help list" and not telling the user what that is.

mattdawkins:
I'll add a link to the deployment guide and the mailing list - these types of requests shouldn't be posted on a public discourse, as it will depend on the person doing the request for more resources whether or not we further support them and how soon.

subdavis:
rolled into #1151

PR #1155, opened 2022-02-07
No keyframe or interpolate defaults

subdavis:
This reconciles differences between web and desktop in the other direction:

* keyframe is not set by default and should _never_ be set, because all frames in a json export are always keyframes.
* interpolate is not set by default

Because these are repeated in every frame, even small datasets will have lots of useless information encoded in these.  keyframe=True is meaningless because interpolated frames are never exported: they only exist at runtime.  interpolate=True can be set, but the default is for them to be false and unset.

## Note

This does not change how the annotator saves annotations.  If an annotation is drawn, it will still have these useless values in the "upsert" payload and end up getting saved.  Scrubbing these useless values at save time (either client side or server side) is out of scope for this PR

### DIFF
--- a/client/platform/desktop/backend/serializers/nist.ts
+++ b/client/platform/desktop/backend/serializers/nist.ts
@@ -105,7 +105,6 @@ function loadObjects(
             track.begin = Math.min(track.begin, frameNum);
             track.features.push({
               frame: frameNum,
-              keyframe: true,
               bounds: adjustedBounds,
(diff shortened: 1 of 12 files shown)


BryonLewis (approved):
I'm good with this.  I just knew that anytime a user saves the data all the values get added back in from the client by default (at least the keyframes ones, not the `interpolate:false`).

Issue #1159, opened 2022-02-09
alternate strategies for compressing filenames in upload dialog


collinsr:
**Is your feature request related to a problem? If so, Please describe.**
The upload dialog sometimes chooses to hide parts of the filename which can make it difficult to check that I'm associating the right annotations with the video.

**Describe the solution you'd like**
Some way of seeing the full pathname / basename: perhaps displaying it in a smaller font size (similar to the size used for the 'Video file' and 'Annotation File (Optional)' headers)

**Additional context**

I've attached a screenshot showing the issue.

![dialog-filename-compression](https://user-images.githubusercontent.com/7387940/153268778-7a0ab319-9051-49d0-a11a-ade1db90ab1a.jpg)

PR #1162, opened 2022-02-10
Don't strip off too many periods from bulk uploaded filenames

### LINKED ISSUE #838: [BUG] Uploading videos in bulk with periods in title ignores periods
e.g. uploading 

test_10.00hz.mp4  test_30.00hz.mp4  test_60.00hz.mp4
test_29.75hz.mp4  test_59.94hz.mp4  test_9.5hz.mp4

results in a folder with the below not their full names

test_10 test_29 test_30 test_59 test_60 test_9

waxlamp:
This PR updates the regex that is used to strip off filename suffixes when multiple files are being bulk uploaded. The existing regex was too greedy and would strip off everything after the first period in the name; the new one does so only for the last one.

Closes #838.

### DIFF
--- a/client/platform/web-girder/constants.ts
+++ b/client/platform/web-girder/constants.ts
@@ -30,7 +30,10 @@ interface GirderMetadataStatic extends DatasetMetaMutable {
  */
 type GirderMetadata = DatasetMeta & GirderMetadataStatic;
 
+const fileSuffixRegex = /\.[^.]*$/;
+
 export {
+  fileSuffixRegex,
   GirderMetadataStatic,
   GirderMetadata,
 };
--- a/client/platform/web-girder/views/Upload.vue
+++ b/client/platform/web-girder/views/Upload.vue
@@ -113,7 +113,7 @@ export default defineComponent({
         createSubFolders,
         name:
           internalFiles.length > 1
-            ? defaultFilename.replace(/\..*/, '')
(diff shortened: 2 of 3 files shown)


BryonLewis (approved):
Thanks for fixing that.
Could be a separate PR but I noticed when Roddy sent that link about the weird truncation of longer filenames given the space we provided.
Could increase `truncate-length` on the `<v-file-input>` to like 75 or more given the space that is provided.
https://vuetifyjs.com/en/api/v-file-input/#props-truncate-length

waxlamp:
> Thanks for fixing that. Could be a separate PR but I noticed when Roddy sent that link about the weird truncation of longer filenames given the space we provided. Could increase `truncate-length` on the `<v-file-input>` to like 75 or more given the space that is provided. https://vuetifyjs.com/en/api/v-file-input/#props-truncate-length

Let's do that in a separate PR. For reference, we're talking about #1159.

waxlamp:
I forgot to mention it, but technically this regex (and the one it replaces) would strip off a lone period from the end of a filename (e.g., `hello.` -> `hello`). I don't think this every happens because logic previous to this step will flag files without a proper extension (maybe? Not sure about this), but the simple fix would be to use `+` instead of `*` in the regex (requiring at least one character after the period).

Let me know what you think and we can always make that change in a future PR.

PR #1171, opened 2022-02-15
Variation on MultiCam Media Controller

subdavis:
# Summary of changes

* Roughly 125 fewer lines
* The motivation here was to make Multicam support more fully integrated.  Instead of having different concepts for single and multi-cam code paths, this change **Removes AnnotationWrapper** and places its functionality inside `useMediaController`.  That composition function treats all cameras the same: `default` is not a special case.
    * It removes the need to do `{} as MediaController`, and instead throws a proper exception if the mediaController is ever attempted to be used before initialization.
    * This change backs out our abuse of child -> parent binding using `ref.value.mediaController`
* It also makes the "aggregate" media controller the primary way to interact with media.  Like the earlier version, it's reactive and recomputes when cameras are added.
* MediaController (the camera-specific one) extends AggregateMediaController.  There were several functions previously only available on Aggregate that should have been camera-specific, like resetMapDimensions.  LayerManager is camera-specific, and it gets passed a camera-specific mediaController, so it should only be calling camera-specific functions on MediaController.

## Description of logic

* Viewer initializes `useMediaController` (rather than a specific camera), which injects 2 static references for all children of Viewer.  `AgregateController` and `cameraInitializer`.
* Now, because aggregateController has been `provide()`ed one layer up, it's no longer necessary to pass through the template.
* Cameras are created as components, and they register themselves with the injected `cameraInitializer` function during their `setup()`.  They no longer have to worry about the internals of `useMediaController.data`, and are not given access to any state but their own.

## Drawbacks to this approach

* The arguments and return types to `cameraInitializer` had to be written excplicitly rather than inferred because of the provide/inject pattern.
* ImageAnnotator and VideoAnnotator now rely on function hoisting to pass references to `seek, play, pause, etc.` to the initializer before those function definitions have executed.  This isn't currently an issue because none of those functions are called except by human interaction, but it's worth noting.

# Other caveats

This is not ready to merge, but it is ready to at least discuss.

* I broke some CSS thing
* I also broke camera selection


BryonLewis (commented):
Just adding these here for reference.

client/src/components/annotators/useMediaController.ts
```diff
@@ -319,14 +297,14 @@ export function useMediaController() {
...
-          }
-        });
+      handleMouseMove(/*evt: MouseEvent*/) {
+        // const offsetX = evt.clientX + 10;
+        // const offsetY = evt.clientY - 25;
+        // window.requestAnimationFrame(() => {
+        //   if (imageCursorRef.value) {
+        //     imageCursorRef.value.style.transform = `translate(${offsetX}px, ${offsetY}px)`;
+        //   }
+        // });
```
BryonLewis, line 307: is this intended to be handled outside of the individual cameras?
There is some stuff that I probably need to do to get mouse-trap shortcuts working as well as proper behavior of being in a 'new' state and selecting a different camera before completing an annotation.

BryonLewis (approved):
Small comments, which I may try to address after this initial merging.  Making an internal list of things I need to update/fix like v-mousetrap and handling some weird switching of camera instances.

client/src/components/annotators/useMediaController.ts
subdavis: I couldn't figure out what this did, so I disabled it to see if anything broke.

It doesn't appear to be doing anything.  What is this for?

BryonLewis: This handles the icon for new items attached to the mouse.
image

Only in creation mode does this icon come up.

subdavis: OK, thanks.  Not sure why I didn't notice it.

client/dive-common/components/Viewer.vue
```diff
@@ -414,18 +412,25 @@ export default defineComponent({
 
     watch(datasetId, reloadAnnotations);
 
+    const controlsRef = ref();
+    const controlsHeight = ref(0);
+    const controlsCollapsed = ref(false);
+    function handleResize() {
+      controlsHeight.value = controlsRef.value.$el.clientHeight;
+      onResize();
```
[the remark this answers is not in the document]
subdavis: Let's fix this after merging to the main feature branch.

client/src/components/annotators/useMediaController.ts
subdavis: I'll fix this now, and try and resolve the CI failure.

PR #1194, opened 2022-03-04
Clone share

### LINKED ISSUE #1002: Sharing a clone should also share other required data automatically
```
WHEN a user tries to share a clone:
    IF the user is an OWNER of the original
        THEN the original should be shared READ-ONLY with the same target user as the clone
    ELSE the user should get an error message that they are not allowed to share a clone, since they have no permissions to share the source media.
```

marySalvi:
Closes #1002 

Creates new endpoint to download media. Endpoint checks for read access on the dataset, gets cloneroot and returns the media if found in the cloneroot.

### DIFF
--- a/client/src/components/annotators/ImageAnnotator.vue
+++ b/client/src/components/annotators/ImageAnnotator.vue
@@ -19,8 +19,7 @@ interface ImageDataItemInternal extends ImageDataItem {
 
 function loadImageFunc(imageDataItem: ImageDataItem, img: HTMLImageElement) {
   // eslint-disable-next-line no-param-reassign
-  img.src = '/api/v1/dive_dataset/61afbcacdbf9dc586bd7e27c/media/6196c871555dd52064119464/download';
-  // img.src = imageDataItem.url;
+  img.src = imageDataItem.url;
 }
 
 export default defineComponent({
--- a/client/src/components/annotators/ImageAnnotator.vue
+++ b/client/src/components/annotators/ImageAnnotator.vue
@@ -19,7 +19,8 @@ interface ImageDataItemInternal extends ImageDataItem {
 
 function loadImageFunc(imageDataItem: ImageDataItem, img: HTMLImageElement) {
   // eslint-disable-next-line no-param-reassign
-  img.src = imageDataItem.url;
+  img.src = '/api/v1/dive_dataset/61afbcacdbf9dc586bd7e27c/media/6196c871555dd52064119464/download';
+  // img.src = imageDataItem.url;
 }
 
 export default defineComponent({
--- a/server/dive_server/crud.py
+++ b/server/dive_server/crud.py
@@ -128,9 +128,7 @@ def getCloneRoot(owner: GirderModel, source_folder: GirderModel):
                 ),
                 code=404,
             )
-        print("1")
         verify_dataset(source_folder)
-        print("2")
         next_id = fromMeta(source_folder, constants.ForeignMediaIdMarker, False)
     return source_folder
 
--- a/server/dive_server/crud.py
+++ b/server/dive_server/crud.py
@@ -111,7 +111,7 @@ def verify_dataset(folder: GirderModel):
 def getCloneRoot(owner: GirderModel, source_folder: GirderModel):
     """Get the source media folder associated with a clone"""
     verify_dataset(source_folder)
-    next_id = fromMeta(source_folder, constants.ForeignMediaIdMarker, False)
+    next_id = source_folder.get(constants.ForeignMediaIdMarker, False)
     while next_id is not False:
         # Recurse through source folders to find the root, allowing clones of clones
         source_folder = Folder().load(
@@ -129,7 +129,7 @@ def getCloneRoot(owner: GirderModel, source_folder: GirderModel):
                 code=404,
             )
         verify_dataset(source_folder)
-        next_id = fromMeta(source_folder, constants.ForeignMediaIdMarker, False)
+        next_id = source_folder.get(constants.ForeignMediaIdMarker, False)
     return source_folder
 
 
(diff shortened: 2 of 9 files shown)


client/src/components/annotators/ImageAnnotator.vue
```diff
@@ -19,7 +19,8 @@ interface ImageDataItemInternal extends ImageDataItem {
 
 function loadImageFunc(imageDataItem: ImageDataItem, img: HTMLImageElement) {
   // eslint-disable-next-line no-param-reassign
-  img.src = imageDataItem.url;
+  img.src = '/api/v1/dive_dataset/61afbcacdbf9dc586bd7e27c/media/6196c871555dd52064119464/download';
+  // img.src = imageDataItem.url;
```
subdavis, line 23: BTW, you can avoid editing the client by returning the new URLS  from the `crud_dataset.get_media` function.  Instead of the existing girder API download URIs, you can return the new one you made.

subdavis:
Didn't look too closely at the code, but this is exactly the approach I expected.  Looks good!

subdavis:
Something you may have to take into account is that this has the potential to break Range requests, which are important for downloading video.  

https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests

client/src/components/annotators/ImageAnnotator.vue
marySalvi: great, that is the plan for wrapping this up. This line was leftover from a quick check and would change back on cleanup/finishing

server/dive_server/crud.py
```diff
@@ -118,6 +118,7 @@ def getCloneRoot(owner: GirderModel, source_folder: GirderModel):
             next_id,
             level=AccessType.READ,
             user=owner,
+            force=True,
```
subdavis, line 121: I think there's an issue here.   Now that we aren't enforcing permissions on the clone root, an attack has opened up.

If I know the identifier of a dataset I don't have access to (and have not been given a clone of), all I have to do is create a folder and set it up as if it were a clone of that folder. 

All I have to do is use the metadata editor and set `foreign_media_id`: https://viame.kitware.com/girder#user/5e56c7d778ed364cd0fd30a2/folder/606e0d89a1b1ab0602c28985

This is an obscure vulnerability, but not something we should allow.  My suggestion would be to migrate the foreign media id from metadata where users can edit it to a designated new property that cannot be set by the user.  That way, there's no way to forge a clone.

subdavis (changes requested):
Couple issues.  Let me know if you'd like to go over them in person.

server/dive_server/views_dataset.py
```diff
@@ -83,6 +87,34 @@ def create_dataset(self, cloneSource, parentFolder, name, revision):
...
+            force=True,
+        )
+    )
+    def download_media(self, folder, item):
+        root = crud.getCloneRoot(self.getCurrentUser(), folder)
+        if item["folderId"] == root["_id"]:
+            files = list(Item().childFiles(item))
+            if len(files) != 1:
+                raise RestException('Expected one file', code=400)
+            return File().download(files[0])
```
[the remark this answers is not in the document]
waxlamp: I was wondering about this. In [`file.py`](https://github.com/girder/girder/blob/master/girder/api/v1/file.py#L263-L278), the `File.download()` method grabs these bytes from the `Range` header. Would those be inherited from the `Range` header supplied to this endpoint?

server/dive_server/crud.py
waxlamp: This vulnerability already existed before, it just wasn't exploitable (for the same reason this PR is necessary), right?

Would it be ok to defer fixing this to a future effort? The migration seems like it would be a fairly heavy lift on its own.

subdavis: > This vulnerability already existed before

I don't think so.  Adding this `force=True` causes it.  Before, you could forge a clone, but your requests would fail because you didn't have access to the source.

Because this PR introduces an exploitable way to get around the permissions model, I don't think it can be merged, but we can discuss if I'm misunderstanding.

server/dive_server/views_dataset.py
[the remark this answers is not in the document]
subdavis: You should be able to basically copy-paste the code from the file linked above to here, using cherryPy to get the header value and pass to `File().download()`

server/dive_server/views_dataset.py
```diff
@@ -83,6 +87,34 @@ def create_dataset(self, cloneSource, parentFolder, name, revision):
...
+    @autoDescribeRoute(
+        Description("Download a media file")
+        .modelParam(
+            "id",
+            level=AccessType.READ,
+            **DatasetModelParam,
+        )
+        .modelParam(
+            "mediaId",
+            description="test",
```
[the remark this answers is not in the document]
marySalvi: fixed with 1202d4af38580ec147b027dbc37ab790f7a391c0

client/src/components/annotators/ImageAnnotator.vue
marySalvi: fixed in 17232d5104da398953d5baf748e56cfaa06183f9

server/dive_server/views_dataset.py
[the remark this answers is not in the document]
marySalvi: fixed by @subdavis  in 326c50e9c5c4f5d53542f85ff4bb7512f0b9a490

server/dive_server/crud.py
marySalvi: fixed by @subdavis  in 326c50e9c5c4f5d53542f85ff4bb7512f0b9a490

subdavis (approved):

PR #1195, opened 2022-03-04
Use zero padding on bounding box extensions

### LINKED ISSUE #1168: If head tail points are made within a bounding box, do not adjust the bounding box size
**Describe the bug**
If head tail points are made within a detection with an existing bounding box, do not adjust the bounding box size

**To Reproduce**
Make a box, make head/tail points within the box, sometimes the boxes get expanded or modified

**Related data**
Any

**Expected behavior**
Don't modify box size if points are created within the existing box's boundaries, 95% of time user is just adding or modifying points on an existing detection (creating a new detection without an existing box, in line mode, is a different case)

If a point is created outside of the box, adjust the box size to cover the point with no additional expansion ratios added to the bounding box that encompasses the points, this is a frequent occurrence (point added very close to the edge off the box and the box just needs to be updated to minimally include/end at the point, not the point + some large scale factor or additive amount)

waxlamp:
See #1168 for background.

This seems like the right approach to me, in general. If there's a need to have padding sometimes, but not other times, we should address that, possibly immediately (if changing this behavior is going to break someone else's workflow).

There's some uncertainty expressed by the codebase itself: a zero-padding vector was already in there, but commented out. This PR comments out the 10% box in favor of the zero box. We should probably figure out which one is the better choice, and whether to expose the padding vector to the frontend as a user setting.

Closes #1168.

### DIFF
--- a/client/dive-common/recipes/headtail.ts
+++ b/client/dive-common/recipes/headtail.ts
@@ -12,21 +12,22 @@ export const TailPointKey = 'tail';
 const EmptyResponse: UpdateResponse = { data: {}, union: [], unionWithoutBounds: [] };
 
 /* Standard 10% padding */
-const PaddingVector: [number, number][] = [
-  [-0.10, -0.10],
-  [-0.10, 0.10],
-  [1.10, -0.10],
-  [1.10, 0.10],
-  [-0.10, -0.10],
-];
-/* No padding */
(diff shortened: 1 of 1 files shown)


subdavis (changes requested):
edit: rereading

subdavis:
This change does not resolve the issue as written, it just changes the padding.

PR #1204, opened 2022-03-24
Clone key migration and download ranges

subdavis:
## Summary of changes

* move `meta.foreign_media_id` to `foreign_media_id`
* Migrate
* Expose new field from folder
* Respect range headers in media download.


marySalvi (approved):
Looks great. Confirmed range headers and migration of foreign id locally

PR #1237, opened 2022-05-04
Group Creation

subdavis:
Things are getting easier now!  

## Summary

* This pull request introduces the ability to create activity groups using multi-select.
* It also fixes the issue with group type editing to make the group style editor work.

https://user-images.githubusercontent.com/4214172/166807083-70cd0588-b5ff-4595-82aa-b1962b800566.mp4

part of #850

### DIFF
--- a/client/dive-common/components/EditorMenu.vue
+++ b/client/dive-common/components/EditorMenu.vue
@@ -42,6 +42,10 @@ export default Vue.extend({
       type: Boolean,
       default: false,
     },
+    groupEditActive: {
+      type: Boolean,
+      default: false,
+    },
     tailSettings: {
       type: Object as PropType<{ before: number; after: number }>,
       default: () => ({ before: 20, after: 10 }),
@@ -142,7 +146,10 @@ export default Vue.extend({
     mousetrap(): Mousetrap[] {
       return flatten(this.editButtons.map((b) => b.mousetrap || []));
     },
-    editingHeader(): {text: string; icon: string; color: string} {
+    editingHeader() {
+      if (this.groupEditActive) {
+        return { text: 'Group Edit Mode', icon: 'mdi-group', color: 'primary' };
+      }
       if (this.multiSelectActive) {
         return { text: 'Multi-select Mode', icon: 'mdi-call-merge', color: 'error' };
       }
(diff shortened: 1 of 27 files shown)


subdavis:
## Discussion

Some areas are still incomplete

* I haven't worked out the UI for editing participation ranges yet.  I think this might require customer feedback
* The logic to actually edit the activity group members when multi-select changes is not hooked up.  Save doesn't get triggered.  This is more to demonstrate the UI.
* I changed my mind about not being able to select multiple things at once.  While editing a group, it seems useful to be able to select an individual track in order to identify it uniquely among other tracks in the group to signify which frame ranges (or activity participangs) you want to change.
* Certain conveniences, like jumping to time stamp when group is selected and auto-scrolling the group list when a track is selected and stuff like that is not yet done.   These will be polish tasks.  Also, auto-switching to the track details tab when group edit mode is entered
* Deletion hasn't been thought about yet.  Probably want 2 different delete mechanisms: delete group only and delete group and all members.
* I don't like the difference between clicking the track number (triggers selection) and clicking group number (triggers editing).  I did this because "selection" of a group is useless.  I think that the group editor sidebar can function as everything: showing what tracks are in it, letting you see all the frame ranges and edit the type, etc.  Adding distinct selected and editing states for a group adds complexity for no reason IMO.  Also, since you can't "right click" a group, there's no way to enter editing state without clicking some button on the group list item.   Probably will end up making the ID number not clickable and adding a dedicated edit button for consistency.

BryonLewis (approved):
I think this can merge as is right now.  Everything seems to be okay.  The questions I have are minor and can probably be resolved in the next version.

I have a list of things that I think may want to be updated at some point in the future.   This is to just keep track of them.

- Selecting a single track when items overlap for selection is difficult when adding new groups.  A click will always sequentially select all trackIds if they overlap instead of one and require another click for another selection.
- Scrolling to the selected GroupId
- Tooltip hover in Group list should probably specify "GroupId" and not "TrackId".  (I like clicking on it selects the group
- Selecting a group should jump to the beginning of the group timeline if not already within the timeline, much like how track selection works
- There are some closely overlapping tracks and I believe that the Layer Tooltip (toggled on in the display settings like boxes/polygons) should probably include group information as well as track information
- I kind of feel while having a group selected the existing of the mode should include clicking on an empty area.  I get that it's a bit harder because after selecting a group it is automatically in edit mode.

subdavis:
Thanks Bryon, I'll address these in the next PR.

PR #1238, opened 2022-05-09
Complete Activity Group Support

### LINKED ISSUE #850: [FEATURE] Ability to create activity groups.

subdavis:
## Summary

* Makes membership ranges editable
* Adds button for setting range to current frame
* Fixes some shenanigans with imports.  Reminder that all code in `src/` needs to use relative imports in order to build the library correctly.
* Removes `TypeList.vue` because it was replaced by `FilterList.vue`
* Hooks up KPF support in desktop
* Fixes desktop icon for win/mac/linux by putting it in the standard location `build/icon.png`
* refactor `useVirtualScroll` so I can use it in both virtual-scroll-list components.

fixes #850


subdavis:
I think this is ready for review.  @BryonLewis PTAL.

client/src/components/TypeEditor.vue
```diff
@@ -3,13 +3,12 @@ import {
...
-
 import { usePrompt } from 'dive-common/vue-utilities/prompt-service';
 
+import BaseFilterControls from '../BaseFilterControls';
+import type Group from '../Group';
+import type StyleManager from '../StyleManager';
+import type Track from '../track';
 import { useReadOnlyMode } from '../provides';
 
 export default defineComponent({
```
BryonLewis, line 14: I still think this component needs some props to reference properly the differences between track data and group data.

Stuff like the confidence/label checkboxes don't work properly.
The concept of deleting a type doesn't make sense because there are no empty group types that can be deleted.

client/src/provides.ts
```diff
@@ -128,6 +131,8 @@ export interface Handler {
   groupAdd(): void;
   /* Put UI into group editing mode */
   groupEdit(id: AnnotationId | null): void;
+  /* Select next group in the list */
+  groupSelectNext(delta: number): void;
```
BryonLewis, line 135: I think I mentioned this elsewhere but I don't any actual usage of this provides function.

docs/DataFormats.md
```diff
@@ -10,13 +10,13 @@ DIVE Desktop and Web support a number of annotation and configuration formats.
 * DIVE Annotation JSON (default annotation format)
 * DIVE Configuration JSON
 * VIAME CSV
-* COCO and KWCOCO
-* KPF (KWIVER Packet Format) for MEVA
+* COCO and KWCOCO (web only)
+* KPF (KWIVER Packet Format) for MEVA (web only)
```
BryonLewis, line 14: should we advertise this when we don't have a user facing UI to load this.  Or should we explain how this is done?
Unless I missed something, I saw some export of .yml/KPF extensions but know actual usage.

docs/Mouse-Keyboard-Shortcuts.md
```diff
@@ -43,8 +43,9 @@ Most editing controls are available when a track or detection is selected.
 | ++i++ | toggle interpolation for the current range of the selected track |
 | ++m++ | enter merge mode for on selection |
 | ++shift+m++ | commit (finalize) merge for selected tracks. |
+| ++g++ | create new group including the selected track |
```
BryonLewis, line 46: should this be added to the keyboard/mouse shortcuts in the app?

BryonLewis (changes requested):
The UI features are really nice.  I like the range selector and how that works especially with taking current frame and allowing the multiple  frame ranges.

There are a few things that don't fit directly into code comments so I'm going to list them below:

- The dropdown for group type changes in the group list needs to be disabled in ReadOnly Mode
- I think there may still be a reason to view detailed ranges and information about a group while in ReadOnly mode.  The current implementation locks out viewing that info because selecting a group has the immediate effect of putting it into edit mode.  I think the act of editing ranges and selecting new tracks for a group should be the things locked out during edit mode.
- LOW PRIORITY QUESTION: Wondering if the Group Event Viewer should handle gaps like it does for interpolation when a group is selected?
- Within `useModeManager.ts` - `handleUnstageFromMerge` if you have a track and group selected with only the track as a member and delete the track you will get conflicts as it tries to remove a member from a group that has already been deleted.  The deletion of a track should probably begin with deselecting the group at the same time.
- Splitting of tracks within a group can result in some weird behavior.  It may be best to not split tracks that are in groups and require the track to be clean (not in a group) before splitting.
- Deletion of groups through the group type list needs to be clearer in the text about what it is doing.  Currently it's using the same text as deleting tracks of visible types when really you're deleting the selected/visible groups and not the tracks themselves.

client/dive-common/components/TrackDetailsPanel.vue
```diff
@@ -292,10 +295,47 @@ export default defineComponent({
...
+        </v-btn>
+        <v-btn
+          color="secondary"
+          class="mt-2"
+          :disabled="readOnlyMode"
+          depressed
+          block
+          x-small
+          @click="toggleMerge"
+        >
```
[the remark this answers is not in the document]
subdavis: The cancel button does the same thing as the escape key.  Otherwise there's no way to get out of editing group mode except to hit escape.

docs/DataFormats.md
subdavis: There is UI for it now.   It just requires you to put all your KPF into a single yml file.  There's actually nothing about KPF that requires 3 files, and you can concatenate all like this.

```
cat  2018-03-07.11-05-07.11-10-07.school.G339.*.yml > combined.yml
```

I'm not really willing to modify all this UI to accept multiple annotation files to support a format that's already deprecated, so this is my compromise.

client/src/components/TypeEditor.vue
BryonLewis: I meant the concept of having the Delete Type button in the `TypeEditor.vue` for Group Types (I know confusing).  I don't think this button will ever become active because there is no cause in which there will be 0 groups associated with a group type.  Where as with track types we can have empty types.  There is no way to add empty group types (I didn't see one yet).
image

client/dive-common/use/useModeManager.ts
```diff
@@ -633,6 +680,7 @@ export default function useModeManager({
       trackSeek: handleTrackClick,
       trackSelect: handleSelectTrack,
       trackSelectNext: handleSelectNext,
+      groupSelectNext: handleSelectNextGroup,
```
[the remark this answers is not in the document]
BryonLewis: did you consider pg up/down for the group list and sticking to regular up/down for the track list instead of relying on focus?

subdavis:
@BryonLewis I think I responded to everything except the below

> LOW PRIORITY QUESTION: Wondering if the Group Event Viewer should handle gaps like it does for interpolation when a group is selected?

IDK, maybe.  I want to get feedback from Kellie first.

BryonLewis (changes requested):
The `<Type-Picker>` in the <Group-Item> also needs the `read-only-mode` attribute applied to prevent changing group types when in readOnlyMode.

Other than that everything mentioned seems to be resolved.

BryonCLewis (approved):
LGTM, nice work on this whole thing and creating a UI to make sense of it.

BryonLewis (approved):
Let's approve this with the Right account.

PR #1241, opened 2022-05-11
Fix CSV parse

subdavis:
Image sequences with tracks and more than 256 images will fail to load annotations.

I was bamboozled by cpython: https://stackoverflow.com/questions/2239737/is-it-better-to-use-is-or-for-number-comparison-in-python


BryonLewis (approved):
Ahh, it's the similar gotchas to python floor division (`//`) where it does what you think except not really because it will return an integer as type float if one of the values is a float.  So it only returns an int if both are ints.
https://peps.python.org/pep-0238/#semantics-of-floor-division

PR #1242, opened 2022-05-11
Upgrade python 3.8, Upgrade python deps

subdavis:
* Upgrades to python 3.8 -- Recommend moving to 3.8.13 in your local virtual environment.  3.7 is no longer an active version, and will reach end of life in a year.
* Upgrades most dependencies.  It's been 5 months, and I like to do this about 2 times per year.

Integration and unit tests passing (aside from the failure caused by viame).


BryonLewis (approved):
built, ran the integration tests as best I could (besides pipelines).
Poetry installed and everything seem to be working.

PR #1303, opened 2022-10-14
Fix typo in import formats list (configuation -> configuration)

rgov:


rgov:
Maybe there's an ESLint plugin that can check for typos

BryonLewis (approved):

PR #1458, opened 2024-11-18
Add filter and transcode pipelines

jjnesbitt:
Exposes filter and transcode pipelines to the client. Results from these jobs are placed in the job work directory.

### DIFF
--- a/client/platform/desktop/backend/native/viame.ts
+++ b/client/platform/desktop/backend/native/viame.ts
@@ -59,7 +59,7 @@ async function runPipeline(
 
   //TODO: TEMPORARY FIX FOR DEMO PURPOSES
   let requiresInput = false;
-  if ((/utility_/g).test(pipeline.pipe)) {
+  if ((/utility_|filter_|transcode_/g).test(pipeline.pipe)) {
     requiresInput = true;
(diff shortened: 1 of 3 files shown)


client/platform/desktop/backend/native/viame.ts
```diff
@@ -125,6 +125,13 @@ async function runPipeline(
       command.push(`-s detector_writer:file_name="${detectorOutput}"`);
       command.push(`-s track_writer:file_name="${trackOutput}"`);
     }
+
+    // TODO: Investigate if any other types can use this.
+    if (runPipelineArgs.pipeline.type === 'filter') {
+      command.push(`-s kwa_writer:output_directory="${npath.join(jobWorkDir, 'output')}"`);
+      command.push('-s image_writer:file_name_template="frame%06d.jpg"');
+      command.push('-s image_writer:type="vxl"');
+    }
```
jjnesbitt, line 134: This is to accommodate some pipeline args that I noticed were missing. This afaik is not a complete list, but does allow for some pipelines like `filter_to_kwa` to be run.

BryonLewis (approved):

PR #1491, opened 2025-05-14
Add shared directories in the shared tab

Louis-Pagnier-KW:
In the "shared with me" tab, only datasets were shown.
Now, it also includes any shared directory to the user, so it can download and interact with shared files.

Previously, the user was not able to navigate to a shared directory that was in a private directory.
Now, in the share tab are included all shared directories whose parent is not accessible.

Example:
```
User1
├── Private
│   ├── sharedWithUser2                            (accessible directly)
│   │   └── sharedWithUser2SubDirectory            (accessible through parent)
│   └── stillPrivate                               (not shown)
│          └── alsoSharedWithUser2                 (accessible directly)
├── Public                                         (not visible, because public and not specifically shared)
└── PrivateButSharedWithUser2                      (accessible directly)

```


Louis-Pagnier-KW:
image
Shared directories are directly shown in a Girder Browser instead of a v-table.

The other possible implementation was to keep a table here, and on click route the user to "Browse Data" tab, but i had some problems with the integrated breadcrumb that can't query the current folder path and cause bugs.

Louis-Pagnier-KW:
I finally changed the implementation to keep only one browser as before, and the share tab only redirect to the "browse data" tab.
image

The main problem on the "Browse Data" tab was to have the bread crumb working with path whose parent is not accessible.
In the GirderFileManager component, there was no slot for a custom breadcrumb, as opposed to the DataBrowser component used by it.

So I had to copy the GirderFileManager implementation and update the bread crumb with a custom one that use a new endpoint which returns the path until hitting a non-readable folder, instead of returning 403.

server/dive_server/crud_override.py
```diff
@@ -0,0 +1,146 @@
...
+from girder.models.folder import Folder
+from girder.models.user import User
+from girder.constants import AccessType
+
+def get_folders_shared_with_me_query(
+    user: types.GirderUserModel,
+):
+    query = {
+        '$and': [
+            Folder().permissionClauses(user=user, level=AccessType.READ),
```
BryonLewis, line 16: The only thing I see in this PR that may cause a problem would be with the deployment and how it is used (on small deployments this may be an issue).

Admins while they have implicit access to everything you can still explicitly share folders and files with them.

The current .permissionClauses folder will return back a falsy {} meaning that admins won't see anything explicitly shared with them.

https://github.com/girder/girder/blob/614a8ebdcfe153accf3615e245ddb763208bc871/girder/models/model_base.py#L47-L50

You could check clone the user object and make sure that internal object doesn't see the user as an admin or you could check the response and see if it is an empty object to make it so Admin's can see what is shared with them.

BryonLewis (changes requested):
Thanks this is great.

One minor comment about admins not having data shared with them.  It may only depend on the deployment for how that could impact users but I assume you want to have items show up to people if it's been explicitly shared, including admins.

server/dive_server/crud_override.py
Louis-Pagnier-KW: Thank you, I effectively forgot to test with admin accounts.
The get_folders_shared_with_me_query code was almost the same as the permissionClauses method, so I just added the 'level' condition.

I have also included folders that are explicitly shared with groups that the user belongs to.

BryonLewis (approved):
:+1:

PR #1567, opened 2026-01-08
First-pass native video ffmpeg display setting to prevent transcoding, when selected

mattdawkins:
## Summary

Add a desktop-only setting that plays the source video natively (via on-demand frame extraction over HTTP) instead of forcing transcoding to a web-safe format up front. Eliminates a significant portion of import-time transcode for HEVC/HDR/odd-codec videos that the system video player can already display directly through ffmpeg.

## Changes vs main

- New backend: \`platform/desktop/backend/native/frameExtraction.ts\` — \`getVideoInfo\`, \`extractFrame\`, and \`prefetchFrames\` (each backed by ffmpeg/ffprobe), plus three new \`/api\` routes in \`server.ts\` (\`/video-info\`, \`/frame\`, \`/prefetch-frames\`).
- New \`NativeVideoAnnotator\` Vue component — desktop-only — that drives \`<canvas>\`-based playback off the \`/frame\` endpoint at the dataset's reported fps. Loaded via dynamic import so the web build doesn't pull desktop-only code.
- Settings UI in desktop \`Settings.vue\` and \`Recent.vue\` for toggling native vs transcoded playback per-dataset.
- Switched the rendering surface to \`<canvas>\` to avoid the brief black flash that \`<video>\`-element-driven swaps produced.
- Misc fixes folded in: ordering correction, runtime error guard, lint, test, removed unused await-in-loop.

### DIFF
--- a/client/platform/desktop/backend/native/frameExtraction.ts
+++ b/client/platform/desktop/backend/native/frameExtraction.ts
@@ -1,0 +1,243 @@
+/**
+ * Frame extraction service for native video playback.
+ * Extracts frames on-demand using FFmpeg without requiring
+ * full video transcoding.
+ */
+import npath from 'path';
+import { spawn } from 'child_process';
+import { getBinaryPath, spawnResult } from './utils';
+
+const ffmpegPath = getBinaryPath('ffmpeg-ffprobe-static/ffmpeg');
+const ffprobePath = getBinaryPath('ffmpeg-ffprobe-static/ffprobe');
+
+// LRU cache for extracted frames
+interface CachedFrame {
+  data: Buffer;
+  timestamp: number;
+}
+
+interface FrameCache {
+  frames: Map<number, CachedFrame>;
+  maxSize: number;
+}
+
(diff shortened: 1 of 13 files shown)


client/platform/desktop/backend/native/frameExtraction.ts
```diff
@@ -0,0 +1,257 @@
...
+async function processBatches(
+  batches: number[][],
+  videoPath: string,
+  fps: number,
+): Promise<void> {
+  if (batches.length === 0) return;
+  const [batch, ...remaining] = batches;
+  await Promise.all(batch.map((frame) => extractFrame(videoPath, frame, fps).catch(() => null)));
+  await processBatches(remaining, videoPath, fps);
+}
```
BryonLewis: I'd have to look into how this specific ffmpeg comand works (in extractFrame) but I'm guessing you probably wouldn't want to be spawning a new process for each frame because of the overhead.  There should be an option to be able to a range of frames in a single process.

PR #1574, opened 2026-01-19
Use newer viame runners for training and inference as opposed to kwiver call

mattdawkins:
## Summary

Use \`viame runner\` (a thin wrapper that pre-loads VIAME's environment) instead of bare \`kwiver runner\` for pipeline execution and ONNX export, matching the runner used by the rest of the desktop pipeline tooling.

## Changes vs main

- \`server/dive_tasks/tasks.py\`: pipeline runs (video and image-sequence paths) and \`export_trained_pipeline\` now invoke \`viame runner -p <pipe>\` instead of \`kwiver runner <pipe>\`.
- \`client/platform/desktop/backend/native/viame.ts\`, \`linux.ts\`, \`windows.ts\`: replace \`kwiver runner\` with \`viame runner\` and update \`-p\` syntax in the desktop equivalents.
- \`docs/Command-Line-Tools.md\`: doc reference updated to match.

\`viame runner\` is identical to \`kwiver runner\` from a pipeline-execution standpoint but already sources the VIAME setup script and resolves binary paths against the VIAME install, which removes a class of "binary not found" / partially-loaded-environment failures we hit on Windows and on minimal Linux installs.

### DIFF
--- a/client/platform/desktop/backend/native/linux.ts
+++ b/client/platform/desktop/backend/native/linux.ts
@@ -40,7 +40,7 @@ const DefaultSettings: Settings = {
 
 const ViameLinuxConstants = {
   setup: 'setup_viame.sh',
-  trainingExe: 'viame_train_detector',
+  trainingExe: 'viame',
   kwiverExe: 'kwiver',
   shell: '/bin/bash',
 };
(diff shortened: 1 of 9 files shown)


BryonLewis:
I've modified the system to simplify by removing some confusing key names like viameTrainingExe.  Instead of having that separate key for the training exe everything is reliant on the viame cli so I upated all places where it is reference and the documentation for this as well.

BryonLewis (approved):
Approving after I made some changes to remove the training vs the viame cli.  So now everything uses the viame cli instea dof having a separate kwiver and training executable

PR #1577, opened 2026-01-24
Add the option to have the type list on the bottom of the screen

mattdawkins:
## Summary

Adds an optional bottom-anchored track list with per-column visibility / sort, plus the related layout / scrubber / notes plumbing.

## Changes vs main

- AnnotationVisibilityMenu: toggle between no list, side list, and bottom list. Multi-cam top-button option exposed in the same menu.
- TrackListColumnSettings: choose which columns to show (type, confidence, attributes, start/end frames, notes); sort/filter on any column.
- Notes column with inline edit; ordering keys-aware.
- Bottom mode: right-side toggle between Track Details and Type Filters.
- FrameScrubber: clickable start/end frame labels; z-index fix for Advanced Tools tooltip.
- Server CSV serializer: parse and emit notes alongside attributes; \`models.py\` exposes the notes field.


BryonLewis:
Did some update Commits:

- Completely restructured to have SideBar and BottomBar instead of having it all in the Viewer.vue
- Move Tracks into a ./Tracks folder and have TrackList and TrackItems for both SideBar and BottomBar views
- Update the controls panel to have a subment when in the bottom bar viewer, also updated the no sidebar view to have controls properly aligned
- Updated support for notes to support coco import/export formats.

BryonLewis (approved):
I believe I've updated all failure conditions and cleaned up some code to be a bit clearer and concise.

Issue #365, opened 2020-09-30
[BUG] GeoJS polygon layer must draw twice


subdavis:
Two calls to `draw()` are necessary for the polygonLayer  to get it to pick up style functions applied to the layer.  This shouldn't happen, and we aren't sure what caused it.  The bug was discovered on #355 where an extra reactivity loop update was making the second call, hiding the bug.  That change eliminated the extra loop, causing this to appear.

>  It's always the second render loop.  If you disable the default updateLayers in LayerManager it won't draw anything in the beginning, on first update only text, then the second update it will do rects.  So that eliminates any weird timing issues with loading the data.

> The Polygon webGL render has RGB data for the color of lines.  Looks like I can grab out of the renderer in there a color that corresponds with the type color.

> I've traced the issue to createStyle  -- basically when the app loads, the rectangles don't get drawn.  If you touch basically anything, they appear.  checkbox, visibility controls, even clicking the right region of the screen to trigger selection.

waxlamp:
@subdavis, is this still valid?

If yes, could we fix this in the short term by just calling `draw()` twice, opening a new issue for the performance implications thereof? And also: is this a DIVE bug or a GeoJS bug?

subdavis:
We currently just call draw twice. Not sure what the status of the bug is in GeoJS.  Not sure if it rises to the level of importance to investigate.

waxlamp:
I'd consider this issue closed. Perhaps we should make sure there's a comment explaining why we call `draw()` twice, and perhaps we should ask David Manthey for his insight. But as far as actionability here, I think we're done. Re-open if I'm wrong.

PR #1578, opened 2026-01-24
Fix annotation layer initialization race condition

mattdawkins:
## Summary

Fix a race condition where DIVE annotation layers were created before the GeoJS viewer was ready, causing annotations not to render on first load (#365).

## Changes vs main

- `useMediaController` now exposes the existing `state.ready` flag, and `MediaController` declares `ready: Readonly<Ref<boolean>>` — true only after `initializeViewer()` completes.
- `LayerManager` defers all layer instantiation to `initializeLayers()`, called from a `watch` on `annotator.ready.value` (immediate).
- Event-listener wiring moved into `setupEventListeners()`, invoked from `initializeLayers()` once the layers exist.
- `updateLayers` / `updateAttributes` are guarded by a `layersInitialized` flag.
- The previous workaround that called `updateLayers()` twice during setup is removed.

### DIFF
--- a/client/src/components/LayerManager.vue
+++ b/client/src/components/LayerManager.vue
@@ -93,68 +93,28 @@ export default defineComponent({
     const frameNumberRef = annotator.frame;
     const flickNumberRef = annotator.flick;
 
-    const rectAnnotationLayer = new RectangleLayer({
-      annotator,
-      stateStyling: trackStyleManager.stateStyles,
-      typeStyling: typeStylingRef,
-    });
-    const overlapLayer = new OverlapLayer({
-      annotator,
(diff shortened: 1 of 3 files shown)


BryonLewis (commented):
This isn't a fix at all, a ton of rearranging code that doesn't do anything to address the reference issue.  Doesn't fix #365 at all because if I remove the second `updateLayers` call it does the same thing asif I removed the `[0,1].forEach(() => updateLayers());`

BryonLewis:
Merged with main and started testing.  This doesn't fix the issue and introduces more issues.  I'm closing (the current issue is documented and the fix is a minor adjustment to first time rendering so it isn't a big deal).

PR #1592, opened 2026-02-03
Add time filter persistence, text inputs, and overlay settings panel

mattdawkins:
## Summary

Persist time filters per dataset, expose them as text inputs in the overlay panel, and have pipelines respect the active filter range.

## Changes vs main

- Persist \`timeFilters\` on \`DatasetMetaMutable\`.
- Add overlay-panel text inputs for filter start/end frames.
- Pipeline run path (desktop + web): pass current filter range as \`frameRange\` and forward it to the backend.
- Backend / tasks: forward \`frame_range\` into the kwiver downsampler (\`start_frame\`, \`end_frame\`); for transcode/filter pipelines also set \`renumber_frames\` and \`adjust_timestamps\` so the output media is renumbered relative to the trimmed range.
- Desktop \`viame.ts\`: filter detector/track CSVs by frame range before ingest, and trim image lists by frame range for image-sequence inputs.


BryonLewis:
I've updated the system to use PipelineParams and that type has RunTimeParams like the frameRange and KwiverParams are params passed into kwiver.

@mattdawkins  - I've ran a simple fish detector on this and the output produces two files.  A detector_output.csv and a detector_output_filtered.csv, the filtered file typically has no tracks.  The detector_output has tracks but the frame Ids are reset to 0 and not the range where they shgould be.  I tried setting the `renumbered` setting to true (it was false before) so see if that adjusted frames and it doesn't seem to.

PR #1624, opened 2026-04-21
Add custom run pipeline settings support

Louis-Pagnier-KW:
This PR add the following improvements to the run pipeline menu:
- Add description when cursor hover a pipeline (based on @mattdawkins  work: [viame/master](https://github.com/Kitware/dive/tree/viame/master))
- Add pipeline input/output types when cursor hover a pipeline
- Add a popup to override default pipeline parameters.

image

image

(The goal would be to limit the number of customizable settings, to not overload the UI).

Here are all possible values for parameters types:
- int
- positive_int
- strictly_positive_int
- range_int
- float
- positive_float
- strictly_positive_float
- range_float
- path
- file
- folder
- string

Except for `path`, `file` and `folder` types, which are only icons, there is form validation to ensure user inputs are valid.

Currently, customizable parameters are specified in .pipe files using the `# DIVE_PARAMS["label", type, arg1, arg2, ...]` comment suffix. There are current discussions on whether we should use dedicated files for this or not. However, changing how we read these metadata should not impact a lot of code of this PR.

**Note 1**: Some work of this PR was already done by @mattdawkins on the [viame/master](https://github.com/Kitware/dive/tree/viame/master) branch:

- Description overlay
- Support for passing pipeline params to VIAME using "-s" arguments

**Note 2**: On the main branch of VIAME, this PR will have no effects.
To test the PR, you will have to compile VIAME web docker on the `dive-params` branch.

### DIFF
--- a/client/dive-common/components/PipelineParamsDialog.vue
+++ b/client/dive-common/components/PipelineParamsDialog.vue
@@ -1,0 +1,232 @@
+<script lang="ts">
+import { Pipe } from 'dive-common/apispec';
+import {
+  defineComponent,
+  PropType,
+  ref,
+  watch,
+} from 'vue';
+
+type ValidationRule = (value: string | number) => boolean | string;
+
+export default defineComponent({
+  name: 'PipelineParamsDialog',
+  props: {
+    value: { type: Boolean, required: true },
+    pipeline: { type: Object as PropType<Pipe | null>, default: null },
+    params: {
+      type: Object as PropType<Record<string, string>>,
+      default: () => ({}),
+    },
+  },
+  setup(props, { emit }) {
+    const localParams = ref<Record<string, string>>({ ...props.params });
+    const getIcon = (type: string) => {
+      switch (type) {
+        case 'bool': return 'mdi-toggle-switch-outline';
+        case 'int':
+        case 'positive_int':
+        case 'strictly_positive_int':
+        case 'range_int': return 'mdi-numeric';
+        case 'float':
+        case 'positive_float':
+        case 'strictly_positive_float':
+        case 'range_float': return 'mdi-decimal';
+        case 'folder':
+        case 'path': return 'mdi-folder-outline';
+        case 'file': return 'mdi-file-outline';
+        default: return 'mdi-alphabetical-variant';
+      }
+    };
+
+    watch(() => props.params, (newParams) => {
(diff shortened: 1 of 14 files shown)


client/dive-common/components/RunPipelineToast.vue
```diff
@@ -0,0 +1,89 @@
...
+        case 'FULL-SIZE-BBOX':
+          return 'mdi-square-outline';
+        case 'HEAD-TAIL':
+          return 'mdi-vector-line';
+        case 'MASK':
+          return 'mdi-vector-polygon';
+        case 'TRACK':
+          return 'mdi-gesture';
+        default:
+          return type;
```
BryonLewis, line 39: Wouldn't returning a type probably return a v-icon with some random text in it.  Should the default be empty without a icon, or should we have a true default icon?

client/dive-common/components/PipelineParamsDialog.vue
```diff
@@ -0,0 +1,231 @@
...
+    pipeline: { type: Object as PropType<Pipe | null>, default: null },
+    params: {
+      type: Object as PropType<Record<string, string>>,
+      default: () => ({}),
+    },
+  },
+  setup(props, { emit }) {
+    const localParams = ref<Record<string, string>>({ ...props.params });
+    const getIcon = (type: string) => {
+      switch (type) {
```
BryonLewis, line 25: any chance we could create Type specification for these pipeline input types?

BryonLewis (changes requested):
I haven't build a new container to test the web version or grabbed new pipelines for the desktop.  Looks mostly good, just some question about typing and the icons.

client/dive-common/components/RunPipelineToast.vue
Louis-Pagnier-KW: Yes you are right.  
I think it is better to have a default "unknown" icon.

client/dive-common/components/PipelineParamsDialog.vue
Louis-Pagnier-KW: Yes, just did it

BryonLewis (approved):
I'm good with this, nothing breaks if the underlying viame pipelines aren't updated and I could see some tooltips and other information on the electron version.

Thanks for working on this.

PR #1688, opened 2026-06-08
Add Dataset Info panel and metadata export

PaulHax:
Towards Kitware/dive#1585. Associate dataset-level metadata (e.g. `gfishsite_id`, station info) with a dataset, display it, and carry it out to export so it can be re-linked downstream.

- Adds a `datasetInfo` field to the dataset metadata model (server + client).
- New **Dataset Info** sidebar panel: manual key/value entry and display, round-trips on web and desktop, with save-error retry.
- Passes `datasetInfo` into the VIAME CSV `# metadata` header as one nested JSON entry.

image


BryonLewis:
This looks really good.

I would suggest updating some docs in the `./docs/DataFormats.md` to mention that the DIVE Configuration JSON now has a new key 'datasetInfo' that can contain key/value pairs, make a note under the VIAME CSV section that metadata inside the datasetInfo will be included.

Possibily create a new `./docs/UI-Dataset-Info.md` file explaining what the interface does, where data comes from and how to edit/modify it

Then update `Context sidebar (Web) to have a link to the new `./docs/UI-DatasetInfo.md`

PaulHax:
Added some docs and fixed a double vertical scrollbar issue.

BryonLewis (approved):
Thanks for finding the double scroll bar. I didn't think to add enough data for scrolling.
Looks good.

PR #1690, opened 2026-06-09
Percentile stretch Image Enhancement option

romleiaj:
# Feature: Percentile Stretch for 16-bit IR Imagery

## What it does

Adds a **Percentile Stretch** contrast option to the Image Enhancements panel. It lets
an annotator recover usable contrast from imagery whose real signal occupies only a
narrow slice of its value range — most importantly **16-bit thermal/IR TIFFs**, where
the interesting temperature differences are just a few hundred counts inside a
0–65,535 range and look nearly flat when naively squashed to 8-bit.

When enabled, the picture shown in the annotator is re-rendered so that the chosen
low/high percentile of pixel values maps to black/white, spreading the meaningful
range across the full 0–255 grayscale.

## Why it exists

DIVE transcodes imported images to 8-bit PNGs for display. For a 16-bit IR TIFF that
transcode throws away the dynamic range before the annotator ever sees it, so the
target (e.g. a warm seal against cold ice) is hard or impossible to see. This feature
reads the **original** TIFF on demand and stretches *it*, restoring the detail the
8-bit copy lost. The stretch is computed **per frame** so each image is optimally
scaled to its own content.

## Web vs. Desktop

| Capability | Web | Desktop |
|---|---|---|
| Brightness / contrast / saturation / sharpen (existing SVG filters) | ✅ | ✅ |
| **Percentile stretch (this feature)** | ❌ | ✅ |

The stretch is rendered by the **desktop (Electron) backend only** — it depends on a
new Express endpoint and Node-side TIFF decoding. No equivalent Girder/Python endpoint
was added, so on web the toggle has no backing service. **This feature is
desktop-only.**

## GUI elements added

In the existing **Image Enhancements** side panel (`ImageEnhancements.vue`):

- A **"Percentile Stretch"** on/off toggle switch.
- When on, two sliders appear:
  - **Low %** (0 → 1, 0.01 steps, default 1) — value mapped to black.
  - **High %** (99 → 100, 0.01 steps, default 99) — value mapped to white.
- Live numeric readouts (2 decimals) next to each slider.
- "Reset" clears the stretch along with the other enhancements.

Settings are **per camera** in multi-camera datasets and are **persisted** to dataset
metadata (debounced), so they restore on reload and follow the selected camera.

## How it works (high level)

**Frontend**
- `useImageEnhancements.ts` — extends the enhancement model with an optional
  `percentileStretch { lowPercentile, highPercentile }`, adds per-camera state, and
  factors the filter/isDefault math into shared helpers.
- `Viewer.vue` — keeps the original frame list (`rawImageData`) and, when stretch is
  active for a camera, rewrites each frame's `url` to point at the new
  `/media/display` endpoint (`toDisplayUrl`). Handles per-camera apply/save debouncing,
  camera switching, and reload cleanup.
- `ImageAnnotator.vue` — when the image URL is swapped for the *same* camera (a stretch
  remap), it rebuilds the image cache and redraws **in place** instead of recreating the
  geojs map, so annotation/calibration layers attached by other components aren't
  orphaned.

**Desktop backend**
- `media/displayProcessing.ts` *(new)* — the core. Decodes a TIFF band with `geotiff`,
  computes the per-frame pixel values at the low/high percentiles (histogram for
  integer data, sort for signed-16/float), linearly stretches to 0–255, and encodes a
  grayscale PNG. Results are cached by `(path, low, high, mtime)` with an LRU cap, and
  concurrent requests for the same key share one decode.
- `backend/server.ts` — new `GET /api/media/display?id&frame&low&high` route. Resolves
  the original source file, stretches TIFFs, and **passes non-TIFF originals through
  unchanged** (stretch is a visual no-op rather than an error for 8-bit EO images).
- `native/common.ts` — `getDisplayImagePath()` maps `(dataset id, frame index)` back to
  the **original** pre-transcode file (via `originalImageFiles`), which is what makes
  stretching the real 16-bit data possible.
- `tiles/geotiffTiles.ts` — refactor only: shared `normalizeToU8` / `encodePngRgba`
  helpers moved into `displayProcessing.ts` (no behavior change).

## Files touched (feature only)

| File | Role |
|---|---|
| `client/src/components/ImageEnhancements.vue` | Toggle + sliders UI |
| `client/src/use/useImageEnhancements.ts` | Enhancement model, per-camera state, helpers |
| `client/dive-common/components/Viewer.vue` | Display-URL remap, per-camera persistence |
| `client/src/components/annotators/ImageAnnotator.vue` | In-place redraw on URL swap |
| `client/dive-common/apispec.ts` | Type additions |
| `client/platform/desktop/backend/media/displayProcessing.ts` | **New** — TIFF decode + stretch + PNG |
| `client/platform/desktop/backend/server.ts` | New `/media/display` endpoint |
| `client/platform/desktop/backend/native/common.ts` | Resolve original source image path |
| `client/platform/desktop/backend/tiles/geotiffTiles.ts` | Refactor: share helpers |

### DIFF
--- a/client/dive-common/apispec.ts
+++ b/client/dive-common/apispec.ts
@@ -298,6 +298,12 @@ interface Api {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   getTileURL?(itemId: string, x: number, y: number, level: number, query: Record<string, any>):
    string;
+  getTileHistogram?(itemId: string, options?: {
+    bins?: number;
+    frame?: number;
+    width?: number;
+    height?: number;
+  }): Promise<unknown>;
   importAnnotationFile(id: string, path: string, file?: File,
     additive?: boolean, additivePrepend?: string, set?: string): Promise<boolean | string[]>;
   // Desktop-only calibration persistence functions
--- a/client/dive-common/apispec.ts
+++ b/client/dive-common/apispec.ts
@@ -134,9 +134,11 @@ export interface MultiCamImportFolderArgs {
   sourceList: Record<string, {
     sourcePath: string;
     trackFile: string;
+    /** Per-camera media type when cameras differ (e.g. EO JPG + IR TIFF on web). */
+    type?: 'image-sequence' | 'video' | 'large-image';
   }>; // path/track file per camera
   calibrationFile?: string; // NPZ calibation matrix file
-  type: 'image-sequence' | 'video';
+  type: 'image-sequence' | 'video' | 'large-image';
 }
 
 export interface MultiCamImportKeywordArgs {
--- a/client/dive-common/apispec.ts
+++ b/client/dive-common/apispec.ts
@@ -7,7 +7,7 @@ import { TrackData } from 'vue-media-annotator/track';
 import { Attribute } from 'vue-media-annotator/use/AttributeTypes';
 import { CustomStyle } from 'vue-media-annotator/StyleManager';
 import { AttributeTrackFilter } from 'vue-media-annotator/AttributeTrackFilterControls';
-import { ImageEnhancements } from 'vue-media-annotator/use/useImageEnhancements';
+import { ImageEnhancements, PercentileStretch } from 'vue-media-annotator/use/useImageEnhancements';
 
 type DatasetType = 'image-sequence' | 'video' | 'multi' | 'large-image';
 type MultiTrackRecord = Record<string, TrackData>;
@@ -526,4 +526,5 @@ export {
   TrainingConfigs,
   MultiCamMedia,
   MediaImportResponse,
+  PercentileStretch,
 };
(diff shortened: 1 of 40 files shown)


BryonLewis:
I've made some updates to the code:

- Disabled the percentile scaling on data that isn't greater than 8 bit.  There was no sense in having it there for standard images
- Minor change to the spacing around the percentage values to prevent overflowing onto the next line
- I've added a small histogram display to the viewer for each image so the distribution can be seen
- Support for the web version through girder-large-images tools for scaling and histogram generation
- Also updated the documentation for image enhancements

Below is a small video demonstrating the changes:

https://github.com/user-attachments/assets/6fe8b17f-deab-495c-b12a-7f63e1acf87a

romleiaj:
Looks great! Yeah I really like the histogram there to better see where the tails lie - awesome that it's in web too.

BryonLewis:
Another small update that makes it easier to use the 'parent folder' option for multicamera image-sequenences on both web and desktop.  This also needs to handle that on web the image-sequence for IR images are made into 'large-image' datasets to utilize the girder-large-image functionality for histograms and percentage scaling.

BryonLewis (approved):
I did some updates to make it a bit more compatible by disabling the functionality if the data isn't greater than 8-bit.  Added a histogram, implemented support for web, improved the parent folder importing of datasets to make it easier to just point it towards a folder for importing.

PR #1703, opened 2026-06-17
Fix single-dataset dive_json export emitting tracks as a list

PaulHax:
The single-dataset `dive_json` export built `tracks` as a **list**, while every other producer/consumer treats it as a **dict keyed by id** (`get_annotations`, the multicam path, the `DIVEAnnotationSchema` type, and the client's `MultiTrackRecord`).

The result: a single-cam `.dive.json` export can't be re-imported — `serializers/dive.py` calls `tracks.items()` and raises `AttributeError: 'list' object has no attribute 'items'`.


BryonLewis (approved):
Thanks, this is something I've been meaning to fix but just never got around to it.

PR #1704, opened 2026-06-17
Round-trip datasetInfo through COCO/KWCOCO and VIAME CSV

PaulHax:
## Round-trip `datasetInfo` through COCO/KWCOCO and VIAME CSV (#1585)

Builds on the Dataset Info panel (#1688). The dataset-level `datasetInfo` block (cruise id, station id, lat/lon, a `gfishsite_id` for re-linking to an external DB, …) now travels with the annotations on **export** and is restored on **import** — for both VIAME CSV and COCO/KWCOCO, on both the server and the desktop client.

### Interface changes

**Wire format (serialized keys)**
- **COCO** — written to `info.dive_dataset_info`, advertised in `info.dive_extensions`. Restored on import. Omitted when empty.
- **VIAME CSV** — added to the `# metadata` line as one nested JSON entry keyed `dataset_info` (renamed from `datasetInfo:` in #1688). Omitted when empty.

```
# metadata,fps: 23.976,"dataset_info: {""gfishsite_id"": ""2024TXN012"", ""year"": ""2024""}", ...
info.dive_dataset_info = { "gfishsite_id": "2024TXN012", "year": "2024", ... }
```

The in-memory field stays `datasetInfo` (camelCase); the snake_case wire keys exist only at each serializer boundary.

### Behavior
- **Import merge** (mirrored server + desktop): **Overwrite** (the default) replaces the block wholesale; **additive** merges per-key, imported values winning. A file carrying no `datasetInfo` never touches the existing block.
- **Fix:** the importer only matched a capitalized `Fps:` and silently dropped the frame rate from real VIAME CSVs — metadata keys are now parsed case-insensitively.
- Unified the single-dataset and multicam COCO export paths: `views_annotation.export` now delegates to `crud_dataset._coco_json_export_text` instead of duplicating the track-filtering / image-naming logic.

### Tests & docs
Round-trip, empty-omission, and merge-direction tests on both platforms, plus the case-insensitive `fps` regression test. Updated `docs/DataFormats.md` and `docs/UI-DatasetInfo.md`.

### DIFF
--- a/server/dive_server/crud_rpc.py
+++ b/server/dive_server/crud_rpc.py
@@ -610,6 +610,16 @@ def _get_data_by_type(
     return None, None
 
 
+def merge_imported_dataset_info(existing: dict, imported: dict) -> dict:
+    """Per-key merge of an imported ``datasetInfo`` block over the existing one.
+
+    Imported values win on collision; keys the imported file did not carry are kept
+    from ``existing`` so a re-import never clobbers prior station metadata. Inputs are
+    not mutated.
+    """
+    return {**existing, **imported}
+
+
 def process_items(
     folder: types.GirderModel,
     user: types.GirderUserModel,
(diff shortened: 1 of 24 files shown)


server/dive_server/crud_rpc.py
BryonLewis, line 605: I think you added ther capability of DatasetInfo to be overwritten or merged with existing.  Should this be applied to the DIVE Configuration imports as well.  You can import a dive configuration file with datasetInfo as well from the interface.

PaulHax: Good check.  Looks like DIVE Configuration is not a separate datasetInfo merge implementation. It is parsed into the same `meta` object as CSV/COCO imports, and the shared import-finalization path resolves datasetInfo before saving metadata. 

I added new tests which lock in the merge logic for configuration files.

BryonLewis (approved):

PR #1715, opened 2026-06-24
fix: remove play icon in run pipeline menu

Louis-Pagnier-KW:
This small MR remove the play icon from pipeline run menu (when there was no dive params to setup).
It was misleading and users could believe it would expand a submenu on the right when hovering.

I also replaced the icon when there is dive params by a simple and smaller cog.

image


BryonLewis (approved):
👍

PR #1728, opened 2026-06-30
Fix interactive service launch on system-python builds and surface failures

mattdawkins:
## Summary

The unified interactive service (segmentation + stereo) failed to launch on VIAME builds that use the system Python interpreter rather than a bundled one, and several failure modes degraded silently instead of being reported to the user. This addresses both.

## Launch fixes

- **`bin/python` fallback (`linux.ts`)** — `getViamePythonExe()` hard-coded `<viamePath>/bin/python`, which only exists in bundled binary installs. From-source builds against system Python have no such interpreter (they rely on `PYTHONPATH` set by `setup_viame.sh`). It now uses the bundled interpreter when present and otherwise falls back to `python` on PATH. Windows is unchanged.
- **`-I` → `-s` (`interactive.ts`)** — the service was spawned with `python -I`, but isolated mode implies `-E`, which discards `PYTHONPATH`, so `viame` could not be imported (`No module named 'viame'`). Switched to `-s` (ignore the per-user site dir, but keep `PYTHONPATH`). The original intent of `-I` (avoid a stray cwd entry shadowing packages) is already covered because the process cwd is forced to `viamePath`, which has no shadowing modules.

## Error surfacing

- **Stereo enable launch failures (`interactive.ts`, `api.ts`, `ViewerLoader.vue`)** — the stereo service watcher auto-runs on dataset load (not user-initiated) and previously degraded *all* failures silently so that uncalibrated datasets wouldn't pop a dialog. That also swallowed genuine service-launch failures. `enable()` now tags infrastructure failures with `launchFailed`, and the watcher always surfaces those in a dialog, while benign per-dataset failures (e.g. missing calibration) still degrade quietly.

## Notes

- Stereo-segmentation match failures on the second camera (returning `success: false` with a clear message rather than an empty polygon) are handled in the corresponding `viame.core` Python service and are not part of this client diff.
- Build artifacts under `client/.electron/` are gitignored and not included.

### DIFF
--- a/client/platform/desktop/backend/native/interactive.ts
+++ b/client/platform/desktop/backend/native/interactive.ts
@@ -201,12 +201,15 @@ export class InteractiveServiceManager extends EventEmitter {
       const segConfig = npath.join(pipelines, 'interactive_segmenter_default.conf');
       const stereoConfig = npath.join(pipelines, 'interactive_stereo_default.conf');
 
-      // -I: isolated mode — do not prepend the process cwd to sys.path. Without
-      // this, a sibling folder named "coverage" (e.g. Vitest HTML output under
-      // client/coverage/) shadows Python's coverage package, breaks numba import,
-      // and prevents ocv_watershed from registering during plugin load.
+      // -s: ignore the per-user site-packages dir so a stray package in
+      // ~/.local (e.g. a conflicting "coverage" that breaks numba import and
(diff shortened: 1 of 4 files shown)


BryonLewis (approved):
Tested locally, mostly looks good to me.

PR #1730, opened 2026-06-30
add delete popup in calibration dialog

Louis-Pagnier-KW:
Just add a simple delete popup before deleting a calibration when users click the "delete" button:

image


BryonLewis (approved):
👍

PR #1731, opened 2026-06-30
Fix bottom-sidebar timeline controls layout

mattdawkins:
Restore the previous/next frame buttons, the selected-camera filename, and the synchronize-cameras button to the bottom-sidebar controls bar, which had been hidden behind a popup menu. Keep the count-settings, seek, and camera-control buttons on a single line with the filename below, and let the timeline shrink to fit the panel so its x-axis labels are no longer clipped.


BryonLewis:
Prior to my commits:
image

image


I've made commits to adjust the height/area for the line to make it so it didn't cut off the frame outline

Also made another update to truncate the beginning of filenames when they are too long.  I figure the end is more important because of the ordering of the files.  It will also provide a hover tooltip giving the full name in the bottom sidebar view

image

image

BryonLewis (approved):

PR #1732, opened 2026-06-30
Desktop auto-save cadence + unsaved-changes exit guard

mattdawkins:
## Summary

Two related desktop annotation-safety improvements:

### Auto-save cadence
- Default auto-save interval lowered from **60s to 15s** so annotation work is persisted more frequently when auto-save is enabled.
- Shorter default applies on **desktop only**; web keeps the 60s default to reduce server churn.
- Auto-save remains **disabled by default** (unchanged); this only affects the default delay.

### Unsaved-changes guard on desktop exit
Closing the desktop window with unsaved annotations now prompts instead of silently discarding work.

- The main process intercepts window close while a viewer with unsaved changes is open and defers the decision to the renderer's close guard.
- On unsaved changes the user gets a native **three-button** prompt:
  - **Save and Exit** — saves, then closes
  - **Exit Without Saving** — closes immediately
  - **Cancel** — keeps the app open
- With no pending changes (or no viewer open) the app closes with no prompt.
- If the save fails, the app stays open (the viewer surfaces the save error).

## Testing
- Verified the desktop app builds and runs.
- Open a dataset, make an edit, then close the window: confirm the three-button prompt appears and each option behaves correctly; confirm no prompt when there are no pending changes.

### DIFF
--- a/client/dive-common/store/settings.ts
+++ b/client/dive-common/store/settings.ts
@@ -154,7 +154,7 @@ const defaultSettings: AnnotationSettings = {
   },
   autoSaveSettings: {
     enabled: false, // Disabled by default for backward compatibility
-    delaySeconds: 60,
+    delaySeconds: 15,
   },
   stereoSettings: {
     clearLengthOnCameraFileLoad: true,
--- a/client/dive-common/store/settings.ts
+++ b/client/dive-common/store/settings.ts
@@ -154,7 +154,8 @@ const defaultSettings: AnnotationSettings = {
   },
   autoSaveSettings: {
     enabled: false, // Disabled by default for backward compatibility
-    delaySeconds: 15,
+    // Shorter default delay on desktop; web keeps the longer default to reduce server churn.
+    delaySeconds: isDesktopRuntime() ? 15 : 60,
   },
   stereoSettings: {
     clearLengthOnCameraFileLoad: true,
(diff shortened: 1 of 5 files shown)


BryonLewis (commented):
I don't think I want to lower the default settings any further.  Mostly because on web we create a revision each time save the dataset.  So this could make the web revision history quite noisy.  I initally settles on 60 seconds because that should be enough time for most people to get done with at least a single annotation.  On desktop this obviously doesn't matter as much but on web it could be annoying having all these saves.

Now the user can set the time down to 15 and they want and that is persistant.  So if a user really wants it set to 15 they can do that manually.  I think that is probably enough to satisfy this.

mattdawkins:
Are abrupt shutdowns covered in both web and desktop?

A user exits browser, or closes desktop using windows right click->close

BryonLewis:
For web it will pop open a system level dialog saying you have unsaved changes on either navigating away, closing the tab, or closing the browser:

image


Desktop doesn't, maybe that is something that can be added to this PR.

BryonLewis:
I've updated this so that the Desktop now has the same protections as the web and will warn/ask the user if they close or navigate away with pending saves.

BryonLewis (dismissed):
- Desktop only defaults to 15s (lower amount because no revisions on desktop)
- Added Desktop guards so it warns when navigating away or closing the application that there are unsaved annotations.

BryonLewis (approved):
👍

Issue #1725, opened 2026-06-26
Per-frame Metadata New Feature


PaulHax:
# Per-frame Metadata New Feature

Follow-on to #1585. This proposes how to attach **per-frame metadata** to a DIVE dataset: show
capture-time telemetry (one record per image, for example timestamp, latitude/longitude, water depth,
and altitude) in a playhead-synced panel. Targeting web (Python/girder) and desktop
(Electron/TypeScript).

Feedback welcome.

## Summary

Per-frame telemetry is observed, read-only data that describes the imagery, much like resolution or
frame rate. We propose treating it as a **property of the media**:

- read it from its source file when a dataset opens,
- show it in a panel synced to the playhead,
- and persist nothing derived.

The source file the user already has next to the imagery is the single source of truth. Everything
else is an in-memory, read-time projection of it. This keeps the feature small: a read-only viewer
with no new storage, no import step, and no data that can drift out of sync.

### To setup dataset with metadata

1. Make a delimited text file (`.txt`/`.csv`) with a header row and one row per image.
2. Give it a column of image file names — that column is what joins each row to its frame.
3. Drop the file next to the imagery (or, for multicam, at the parent root or in each camera folder).

## Non-goals

| Item                                           | Why                                                                                                                                                                                                                                                                                                                                                                                                                     |
| ---------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| Editing telemetry/frame-metadata                              | Telemetry is observed, not authored.                                                                                                                                                                                                                                                                                                                                                                                    |
| Whole-frame classification and per-frame notes | These are editable, authored labels, a different kind of data from read-only observed telemetry. They are annotation-shaped (a whole-frame classification is a detection with a class and confidence; a note is a per-frame attribute) and, if built, belong in the annotation system rather than this media-property channel. Folding them in would mix editable authored data into a read-only, write-nothing design. |

Deferred features that the design is built to accommodate (export, embedded sources, charting,
training integration, and out-of-folder source selection) are described under Future work.

## Technical design


### Approach

| Decision      | Choice                                                        | Why                                                                                                                                                                                                                                             |
| ------------- | ------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
[...]

PR #1733, opened 2026-07-01
Add read-time per-frame metadata

PaulHax:
## Summary

- Implements v1 per-frame metadata as read-only media telemetry for image-sequence datasets.
- Reads a co-located `.txt` or `.csv` sidecar at view time.
- Matches rows to frames by image filename.
- Shows the active frame's values in the Frame Metadata section of the Media Metadata panel.

The sidecar file remains the source of truth. This PR does not import telemetry into annotations, does not write a derived `frame_metadata.json`, and does not add frame metadata to exports.

## How To Use

1. Create a `.txt` or `.csv` file with a header row and one row per image.
2. Include at least one column containing image filenames. The column can have any name; DIVE finds it by matching the values against the dataset images.
3. Include at least one metadata column beyond the filename column, for example timestamp, latitude, longitude, depth, altitude, or other telemetry fields.
4. Place the file next to the imagery:
   - Single-camera image sequence: put the sidecar in the dataset folder beside the images.
   - Multicam image sequence: put one shared sidecar at the multicam parent folder, or put one sidecar inside each camera folder.
5. Open the dataset and the Media Metadata panel. The Frame Metadata section updates as the playhead moves.

Example:

```text
image_file timestamp latitude longitude water_depth altitude
img_0001.tif 15:40:56 46.575870 -124.603094 192.80 2.78
img_0002.tif 15:41:04 46.575912 -124.603080 193.10 2.70
```

Each metadata row must include the image filename it describes. Rows that do not match an image are ignored. The metadata file does not need a special name.

For multicam, a shared sidecar can use one filename column per camera:

```text
port_image starboard_image timestamp latitude longitude water_depth
port_0001.tif starboard_0001.tif 15:40:56 46.575870 -124.603094 192.80
port_0002.tif starboard_0002.tif 15:41:04 46.575912 -124.603080 193.10
```

### Source Contract

- **Delimited sidecars:** Added Python and TypeScript parsers for `.txt` and `.csv` sidecars with a header row and comma, tab, or whitespace delimiter sniffing.
- **Filename joins:** Match metadata rows to image frames by filename value, with filename normalization matching DIVE's image-key behavior.
- **Raw display values:** Preserve source field order through parsing, serving, and display, and keep all values as raw strings.

### Source Discovery

- **Co-located files:** Read candidate sidecars from the dataset folder, multicam parent folder, or camera child folder.
- **Safety checks:** Reject VIAME annotation CSVs, ignore unrelated text files and bare image lists, and skip ambiguous matches instead of guessing.

### Read Contract

- **Web endpoint:** Added `GET /dive_dataset/:id/frame_metadata?startFrame=<n>&endFrame=<n>` with inclusive, non-negative frame-window bounds.
- **Desktop parity:** Added matching desktop `loadFrameMetadata` support so web and desktop return the same `{ cameras: { cameraName: { frame: values } } }` shape.
- **No source state:** Return an empty `cameras` map when no usable frame metadata source is present.

### Camera Routing

- **Single camera:** Return single-camera records under the `singleCam` camera key.
- **Multicam:** Route parent-level and per-camera sidecars by matching each camera's own image filenames, so shared files with columns such as `port_image` and `starboard_image` bind correctly.
- **Collisions:** If two distinct records target the same camera/frame, omit that frame rather than resolving by precedence.

### Client Display

- **Windowed cache:** Added a bounded client cache around the playhead so the UI fetches frame metadata windows instead of holding the full dataset in client state.
- **Stale response guard:** Ignore older frame-metadata responses when a newer request has already been issued.
- **Panel placement:** Display the active frame's values in the existing Media Metadata panel, where the Frame Metadata section follows the playhead and active camera.

### Persistence and Export

- **Read-only:** Keep frame metadata out of annotation, attribute, and dataset metadata stores.
- **No persistence/export:** Do not import telemetry into annotations, write a derived `frame_metadata.json`, maintain a field registry, or add frame metadata to exports.
- **Docs:** Added documentation for the source contract, architecture, and UI behavior.

## V1 Boundaries

- Supported: image-sequence datasets with co-located `.txt` or `.csv` sidecars.
- Not included: editing frame metadata, importing it into annotations, exporting it, embedded EXIF/KLV, video telemetry, charting, training integration, manual out-of-folder source selection, or server-side parsed-source caching.

## Related Issues

- https://github.com/Kitware/dive/issues/1725 - direct per-frame metadata feature.
- https://github.com/Kitware/dive/issues/1585 - broader metadata/export request.


PaulHax:
superceded by #1741

PR #1734, opened 2026-07-01
Surface pipeline job failures to the user

mattdawkins:
When a pipeline/training job process exits non-zero (and was not cancelled), scan its captured output for lines beginning with `ERROR:` and show them in a dialog. Processes emit a single `ERROR: <cause>` line to report the real failure reason (e.g. a GPU out-of-memory during SAM3 model load), so the user sees the cause without opening the job log. General mechanism — any pipeline that prints `ERROR:` and exits non-zero surfaces it.


BryonLewis (approved):
👍

PR #1737, opened 2026-07-02
Multicam: draw the selected detection on other cameras without switching first

mattdawkins:
## Motivation

In multicam/stereo manual annotation, after drawing a line (or rectangle) on one camera, drawing it on the other camera required left-clicking that camera (switching the selected camera, which also dropped edit mode) and re-entering creation mode. For stereo measurement workflows this adds several clicks per detection.

## Change

While in edit mode, any non-selected camera that is **missing the selected track's geometry at the current frame** now keeps a live creation cursor. Drawing there commits to that camera **under the same track id** — so a line drawn on the left camera can immediately be placed on the right camera with two more clicks.

This generalizes the existing "seamless multicam creation" mechanism (which handled brand-new tracks with no geometry anywhere) to the continuation case.

## Implementation

**`LayerManager.vue`**
- New `cameraAwaitingGeometry()` helper — true when, in edit mode, the selected track has no geometry on *this* camera at this frame (or doesn't exist on it). Complements `isCreatingNewDetection()`.
- `updateLayers`: the non-selected-camera creation-cursor branch now also fires for this case.
- `update:geojson` routing: for the continuation case, the same-id track is created on the drawn-on camera **before** `selectCamera` is called. Ordering matters: since the track then exists, `selectCamera` skips its `trackEdit` toggle — which would call `finalizeInProgress()` and discard the in-progress 1-point line replica, breaking the 2-click line completion. The brand-new-track path (`trackAdd()`) is unchanged.

**`Viewer.vue`**
- `changeCamera` no longer consumes a left-mousedown on a camera awaiting geometry (that click is the start of a draw there). Right-click camera switching (`mouseup.right`) and the camera dropdown are unaffected. Escape exits edit mode, restoring normal left-click switching.

Point/segmentation mode is excluded from all of the above — segmentation has its own cross-camera machinery (stereo warp).

Side benefit: when the second camera's line completes, `onStereoAnnotationComplete` fires as usual, so with "update lengths when modified" enabled the stereo length is computed immediately once both sides have lines.

## Verification

- `eslint` clean on both files.
- `yarn test`: 197 passed / 2 failed — the failures are identical on the pristine tree (Electron install + missing `File` global in the node test env), unrelated to this change.

### DIFF
--- a/client/dive-common/components/AnnotationVisibilityMenu.vue
+++ b/client/dive-common/components/AnnotationVisibilityMenu.vue
@@ -159,7 +159,10 @@ export default defineComponent({
 </script>
 
 <template>
-  <span class="toolbar-group-host">
+  <span
+    class="toolbar-group-host"
+    :class="{ 'toolbar-group-host--expanded': isExpanded }"
+  >
     <!-- Dropdown mode when collapsed -->
     <v-menu
       v-if="!isExpanded"
--- a/client/dive-common/components/EditorMenu.vue
+++ b/client/dive-common/components/EditorMenu.vue
@@ -437,7 +437,10 @@ export default defineComponent({
         </div>
       </div>
       <!-- Collapsed mode for edit buttons -->
-      <span class="toolbar-group-host">
+      <span
+        class="toolbar-group-host"
+        :class="{ 'toolbar-group-host--expanded': isEditButtonsExpanded }"
+      >
         <v-menu
           v-if="!isEditButtonsExpanded"
           :key="editButtonsMenuKey"
@@ -555,10 +558,6 @@ export default defineComponent({
       </span>
       <!-- Segmentation Reset button -->
       <template v-if="activeSegmentationRecipe && editingMode === 'Point'">
-        <v-divider
-          vertical
-          class="mx-2"
-        />
         <v-btn
           color="error"
           class="mx-1"
@@ -772,5 +771,14 @@ export default defineComponent({
 .mode-button{
   border: 1px solid grey;
   min-width: 36px;
+}
+
+/*
+ * Keep the segmentation reset divider from stretching to the full toolbar
+ * height (the flex row can be tall when the edit-types group is expanded).
+ */
+.segmentation-divider {
+  align-self: center;
+  max-height: 28px;
 }
 </style>
--- a/client/dive-common/components/EditorMenu.vue
+++ b/client/dive-common/components/EditorMenu.vue
@@ -451,13 +451,14 @@ export default defineComponent({
             <v-btn
               v-bind="attrs"
               :disabled="!editingMode || activeEditButton?.loading"
+              :loading="!!activeEditButton?.loading"
               :color="activeEditButton?.active ? editingHeader.color : ''"
               class="mx-1 mode-button toolbar-group-activator"
               small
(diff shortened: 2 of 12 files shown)


mattdawkins:
Added a second commit extending the continuation mechanism to **point-click segmentation**:

- **Prompt-point markers no longer render on every camera.** The green/red click markers were drawn by each camera's layer manager with no camera filter, so a click also showed its marker on the other camera at the same pixel location — easily mistaken for an unwarped auto-created point (it appeared regardless of the auto-compute setting or calibration). They now render only on the selected camera.
- **One-click segmentation on the other camera.** Point mode was previously excluded from cross-camera continuation. Now a click on the other camera segments there under the same track id. "Awaiting geometry" for Point mode means *no segmentation polygon on that camera yet*, so box-only detections accept a click. A new `segmentationFinalizePending` handler (invoked on any camera switch) locks in the source camera's pending mask and clears the recipe's accumulated prompt points — they belong to the source camera's image and must not leak into the other camera's prediction. Unlike right-click confirm, it does not deselect, so the flow continues uninterrupted.

Line/rectangle/polygon continuation behavior is unchanged.

mattdawkins:
Rebased on top of `main` (picks up #1736 stereo auto-enable, #1735, and the deployment-docs commit) and added a third commit with fixes from interactive testing:

- **Edit mode on both cameras.** Edit handles are now live on every camera showing the selected track, so a stereo detection can be adjusted on either camera without selecting it first. The mousedown that grabs a handle switches the selected camera (edit mode preserved) without `preventDefault` killing the drag; the edit-layer reset is deferred while the button is held; and a routing fallback commits an edit from a non-selected camera to that camera's own track.
- **Single-press right-click deselect.** A right-click while editing finalizes and deselects the detection in one press (matching single-camera behavior), instead of merely switching cameras and leaving it selected until a second right-click. Also fixes `finalizeInProgress` emitting the garbage GeoJS trace for a LineString still in creation mode (the real feature lives in `shapeInProgress`), which could commit a bogus line.
- **Stereo startup.** Entering a sequence with a stereo feature on now shows the persistent, annotation-blocking loading dialog until the service is ready; enable failures are surfaced instead of degrading silently; and if a measurement is requested while the service is wanted but not running, the handler starts the enable itself and waits — so drawing a line always forces the interactive stereo service up rather than silently producing no measurement.

BryonLewis:
I've made some updates to resolve some minor issues and some other fixes:

- Toolbar button alignment:  I didn't like the way the collapsed vs uncollapsed additional icons (like the delete buttons or reset in segmentation) didn't align properly so I've updated it to align with the collapsed version better.  Also there was a divider in segmentation mode that was too large (stretched above and below the toolbar line) so I removed it
- The segmentation reation used to rotate the magic marker/wand icon.  I changed it to use  a standard spinner instead of  having that icon rotate during the segmentation creation process.
- Multicamera creation/editing Adjustments:
    - Polygon creation didn't work across cameras so it has been udpated so you can just click and start creating the new polygon on the other camera
    - Updated some editing issues with bboxes where the editing bbox would snap back when selecting an editing handle on another camera

BryonLewis (approved):
Tested and made some updates to fix minor issues and some nagging UI/UX issues.

PR #1741, opened 2026-07-06
Add read-time frame metadata sidecars

### LINKED ISSUE #1725: Per-frame Metadata New Feature

PaulHax:
Adds read-only per-frame metadata for image-sequence datasets using a sidecar file placed next to
the imagery. The user-facing CSV filename is `frame-metadata.csv`; `frame-metadata.txt` is supported
for delimited text, and the existing snake_case spellings are accepted by the shared name predicate.

image

Shows the Frame Metadata section in the Dataset Info panel, synced to the active camera and playhead.
The Dataset Info UI is split into collapsible sections so frame metadata and editable dataset
metadata stay visually separate.

## Upload classification change

This PR includes the upload packaging changes needed to keep frame metadata files from being
dropped.

Before this change, web upload classification treated any `.csv` as an annotation candidate. A
`frame-metadata.csv` file could be selected as the annotation file, fail VIAME parsing later, or be
dropped when another CSV won the single annotation slot. Multicam folder upload had a related issue:
server validation could accept side files, but the browser uploaded only media plus an explicit track
file, so camera-folder sidecars could disappear.

The new path makes server validation authoritative. `dive_utils/frame_metadata.py` defines the fixed
sidecar basenames; `crud_dataset.validate_files()` partitions those names before generic CSV handling
and returns:

- semantic file roles (`media`, `annotations`, `datasetConfig`,
  `frameMetadata`)
- an `upload` list
- UI-visible `ignored` reasons

image

The web upload code maps `validation.upload` back to the selected `File` objects instead of
re-classifying them. `crud_rpc._get_data_by_type()` then short-circuits declared sidecars before
annotation parsing and leaves them in the dataset folder.

For clones, the source-list endpoint also follows the dataset clone root. Multicam soft clones keep
shared parent sidecars visible to each cloned camera.

### Desktop behavior

The Electron backend scans the dataset media folders for
declared sidecar basenames, resolves them with the same shared TypeScript parser, and rejects an
explicit annotation import of `frame-metadata.csv` before parsing.

## What changed

- A no-parse server endpoint lists frame metadata source items per camera:
  `GET /dive_dataset/:id/frame_metadata_sources`.
- The browser downloads listed sidecar text and runs the shared TypeScript resolver in the renderer.
- The desktop backend discovers sidecars on disk, reads them, and runs the same shared TypeScript
  resolver before returning the compact resolved payload over IPC.
- The Dataset Info panel was split into reusable sections; Frame Metadata displays in its own
  collapsible section with source details in the section header tooltip.
- User docs describe the supported filenames, delimited text format, placement rules, and limits.

## Design notes

- Frame metadata is like input media, not annotations.
- The sidecar file remains the source of truth. DIVE does not persist a derived
  `frame_metadata.json` or write these values into annotation stores.
- Python never parses frame metadata files. The only mirrored logic is the fixed-name predicate,
  pinned by a shared accepted/rejected fixture used by both Python and TypeScript tests.
- Rows join to frames by filename value, not row order. For multicam, one shared parent
  `frame-metadata.csv` can contain camera-specific image columns; each camera resolves against the
  column that matches its own frames.
- Sidecar precedence is camera folder, camera clone root, multicam parent folder, then parent clone
  root.
- Parser keeps values as strings and preserves source column order.

## Manual test

[okeanos-frame-metadata-upload.zip](https://github.com/user-attachments/files/29867555/okeanos-frame-metadata-upload.zip)

[okeanos-frame-metadata-nonmatching.zip](https://github.com/user-attachments/files/29867754/okeanos-frame-metadata-nonmatching.zip)

Use this fixture to test the web upload path. Upload the zip if using DIVE's zip upload flow, or
unzip it and upload the folder contents as an image sequence.

The zip contains one folder:

- 16 NOAA Okeanos image frames
- `frame-metadata.csv` with matching filename rows

1. Upload the zip file as a new dataset.
2. Confirm the upload dialog classifies `frame-metadata.csv` as frame metadata, not annotations.
3. Open the uploaded dataset in the viewer.
4. Open Dataset Info and confirm the Frame Metadata section shows values for the active frame and
   the section header source tooltip names `frame-metadata.csv`.
5. Step or scrub through the image sequence and confirm the displayed metadata updates with the
   playhead.
6. Confirm no annotation tracks were created from `frame-metadata.csv`.
7. Upload an image sequence with a non-matching `frame-metadata.csv`; confirm Dataset Info reports
   that a source exists but no row matches the current frame.
8. Try a video or large-image upload with `frame-metadata.csv`; confirm it appears in the ignored
   list with the unsupported-media reason rather than being uploaded as annotations.

## Automated coverage

- Python naming parity and upload classification tests.
- Server source-list endpoint tests for single-camera and multicamera datasets.
- Import-path tests that keep declared sidecars in place and out of annotation processing.
- TypeScript parser, resolver, and naming conformance tests.
- Dataset Info panel and shared frame metadata composable tests.
- Web upload package and multicamera registry tests.
- Desktop discovery/read-path/import-guard tests.
- Multicam clone coverage for shared parent frame metadata sidecars.

## Follow up work

- Support frame metadata txt/csv for videos by aligning on frame index or timestamp column.
- Support arbirarly named (or located) metadata files
- Extract metadata from images (EXIF) and videos (KLV)

closes https://github.com/Kitware/dive/issues/1725


PaulHax:
The "Upload classification change" was the wrong approach.  During upload users can declare specific files as a second step after picking an image set.  Also the "frame-metadata.csv" reserved name is not usable.  Working up a refactor of this that will rebase on https://github.com/Kitware/dive/pull/1764

PaulHax:
superseded by #1809

PR #1742, opened 2026-07-07
Add option to infer frame index from filename

romleiaj:
# Frame Alignment Feature (`dev/frame-alignment`)

## What it does

Multi-camera datasets used to require every camera to have the **same number of
frames**, aligned by position (frame 5 = frame 5). That breaks for capture systems
like **KAMERA** that drop frames independently, so frame 5 of one camera isn't the
same instant as frame 5 of another.

This branch aligns cameras by **when each frame was captured** instead of by
position:

1. Reads a capture timestamp from each image's filename (e.g. KAMERA's
   `..._20240407_130757.206341_ir.tif`).
2. Builds one shared timeline, grouping frames with the **same** capture
   timestamp into the same slot.
3. Allows cameras to have different frame counts; a camera with no frame at a slot
   shows *"No frame at this instant"* rather than a stale image.

If timestamps aren't available (or it's a single-camera dataset), it falls back to
the old positional behavior — **existing datasets are unaffected**.

## What a user sees

| Element | Where | Purpose |
|---|---|---|
| **"Infer frame index from filename"** checkbox | Import dialog | Opt-in; allows unequal frame counts. Off by default (old equal-length check still applies). |
| **"No frame at this instant"** overlay | Camera pane | Shown when a camera has no frame at the current slot. |
| **Gap indicator bar** | Under the scrubber | Marks timeline slots where a camera is missing a frame. |

## How it works

- **Timestamp parsing** — one shared module (`client/dive-common/frameTimestamp.ts`)
  used by both platforms: desktop parses local media on load, web parses after
  fetching from girder. The server does not parse timestamps.
- **Timeline** — `client/dive-common/alignedTimeline.ts` (pure, unit-tested) turns
  per-camera timestamped frames into global slots, and computes the gap slots.
- **Playback** — the viewer hands the media controllers an `AlignedFrameResolver`
  that maps a global slot to each camera's local frame (or "no frame"), keeping all
  panes in sync during seek/scrub.
- **Import** — both the client validation and the server (`create_multicam`) no
  longer reject cameras with differing frame counts.

## Note

The KAMERA `YYYYMMDD_HHMMSS.ffffff` convention is confirmed against real sample data
(`data/test_data`) and locked by a test. Cameras must currently share an **exact**
timestamp to align (tolerance `0`); this can be widened later if real captures show
sub-second jitter. The epoch-based fallback patterns are likewise best-effort for
other capture systems.


BryonLewis (approved):
Tested parent folder and multicam import with uneven frames.  Seemed to work properly for all of them after mergning with the other updated dev/sealtk branch.

Added some minor documentation.

PR #1749, opened 2026-07-09
Add keypoint GUI for manual calibration

romleiaj:


BryonLewis:
Most of the changes are related to getting a large-image camera to align with images.  This is because large-image is used for the images that have higher bit-depth and need to be rescaled like IR images.  So a lot of my changes were fixing/adjusting and adding functions to handle  this

- Large-image overview frameTexture — Tiled large-image panes have no img/video quad for Align View / registration ghosts. LargeImageAnnotator now composites a capped overview canvas and publishes it as frameTexture so warps/ghosts have a texture source.
- geojsWarpQuadsForImage crop remapping — Overview canvases are smaller than native frame size. Crop rects now scale into texture pixels while corners stay in native/map space, fixing massively wrong Align / ghost scaling.
- CameraImage shared type (incl. canvas) — Texture type moved to cameraImage.ts and allows HTMLCanvasElement, so image/video/large-image share one warp input model.
- Align View navigation for large-image — On Align-on: unlock GeoJS clamp, fit reference native frame, mirror that FOV via zoomAndCenterFromBounds (not zoom-0 unitsPerPixel). Without this, tiled maps stay on a native tile fit and look stranded / clamped after warps switch to reference coords.
- Multicam clamp unlock for tiled panes — resetMapDimensions used to skip clamp unlock when isMap=true, so Align snaps were silently rejected on large-image. Multicam now unlocks clamp for all panes.
- Viewer registers linked navigation in setup() — useAlignedNavigation / useRegistrationNavigation moved out of async loadData() so onBeforeUnmount attaches correctly and pan/zoom listeners tear down (fixes Vue instance warnings / leaks).
- GeoJS map / WebGL lifecycle hardening — Exit previous maps on re-init/clear; guard annotation redraw/disable after context loss (“Too many active WebGL contexts”) so orphaned maps don’t kill redraws.
- Large-image seek blanks missing frames — seek(undefined) clears tiles/frameTexture like image/video annotators, for aligned-timeline slots with no frame on that camera.
-

BryonLewis (approved):
went through and tested, then did several updates for web/girder support mostly related to large-image being part of the transformation/alignment process when it wasn't before.

Also slightly modified the alignment toggle button to provide a button link to Camera Registration so user's don't ahve to find it in the dropdown box

PR #1751, opened 2026-07-13
Fix rollback leaving stale tombstones that re-delete restored annotations

mattdawkins:
## Problem

Annotations are lazily deleted: a track removed at revision `R` keeps its document and is tombstoned with `rev_deleted = R`. Restoring it means clearing that tombstone.

`rollback()` built one query keyed on `REVISION_CREATED` and used it for **both** the removal and the un-delete:

```python
listQuery = {DATASET: dsId, REVISION_CREATED: {'$gt': revision}}
updateQuery = {'$unset': {REVISION_DELETED: ""}}
TrackItem().removeWithQuery(listQuery)
TrackItem().update(listQuery, updateQuery)   # <-- wrong selector
```

As an un-delete selector that query is wrong: it only ever matches documents `removeWithQuery` just dropped, so **tombstones are never cleared**. The function's own comment states the intent it doesn't implement — *"erase deletions for anything deleted after revision"*.

## Why this is worse than it looks

The bug is **invisible immediately after the rollback**. A track tombstoned `rev_deleted = R` is still visible while the head sits at `R-1`, because `BaseItem.list()` treats `rev_deleted > head` as live. The rollback appears to work.

But `save_annotations()` computes `new_revision = latest + 1`, so **the next save of any kind re-issues revision `R`**. The stale tombstone reactivates and the restored track disappears — permanently, since every later head is `>= R`.

A user who rolls back a bad delete and then makes one unrelated edit silently loses the restored data.

## Fix

Select records to restore by `REVISION_DELETED`; removal still keys off `REVISION_CREATED`.

## Tests

`server/tests/test_annotation_rollback.py` covers both the query shape and the end-to-end property that a restored track **survives a subsequent save**. Note the latter matters: a test that asserts only immediately after the rollback passes even against the buggy code.

Both tests fail on `main` and pass with this change.

### DIFF
--- a/server/dive_server/crud_annotation.py
+++ b/server/dive_server/crud_annotation.py
@@ -140,12 +140,18 @@ def rollback(dsFolder: types.GirderModel, revision: int):
     # And erase deletions for anything deleted after revision
     dsId = dsFolder['_id']
     RevisionLogItem().removeWithQuery({DATASET: dsId, REVISION: {'$gt': revision}})
-    listQuery = {DATASET: dsId, REVISION_CREATED: {'$gt': revision}}
+    removeQuery = {DATASET: dsId, REVISION_CREATED: {'$gt': revision}}
+    # Deletion is lazy, so restoring a record means clearing its rev_deleted
+    # tombstone.  Those records must be selected by REVISION_DELETED: keying this
+    # off REVISION_CREATED only ever matches records removeWithQuery just dropped,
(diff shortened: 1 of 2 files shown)


BryonLewis (approved):
LGTM. Correct fix for the stale-tombstone case, and the post-save assertion is the important one. Note this only affects POST dive_annotation/rollback (scripts/tests), not the readonly revision inspector.  I.E there is no UI that hits this endpoint outside some integration tests.

PR #1753, opened 2026-07-13
Write (note) columns on CSV export so notes survive a round trip

mattdawkins:
## Problem

The server CSV exporter (`export_tracks_as_csv`) emits `(atr)`, `(trk-atr)`, `(poly)`, `(hole)` and `(kp)` columns — but never `(note)`.

This is a one-sided format gap:

- its **own importer** parses `(note)` (`_parse_row`, `viame.py:234-237`)
- the `Feature` model carries the field (`models.py:54`)
- the **desktop TypeScript serializer already writes it** (`viame.ts:790-792`)

So detection notes are dropped by every server-side export, and an export → import cycle — which is how DIVE moves annotations through pipelines and downloads — silently loses them. The two exporters disagree on the same format.

Before this change, a feature carrying notes exported as:

```
0,1.png,0,884,510,1219,737,0.9,-1,fish,0.9
```

— the notes are simply gone.

## Fix

Append `(note)` columns after the geometry columns, matching the column ordering the TypeScript serializer uses so both exporters produce identical rows.

## Tests

Adds an export → import round-trip test asserting notes survive. It fails on `main` and passes with this change; the existing 23 serializer tests are unaffected.

### DIFF
--- a/client/src/components/Tracks/sidebar/SideBarTrackItemView.vue
+++ b/client/src/components/Tracks/sidebar/SideBarTrackItemView.vue
@@ -179,7 +179,8 @@ export default defineComponent({
       <v-tooltip
         open-delay="200"
         bottom
-        max-width="320"
+        max-width="360"
+        content-class="sidebar-notes-tooltip"
       >
         <template #activator="{ on, attrs }">
           <v-btn
@@ -187,7 +188,7 @@ export default defineComponent({
             small
             icon
             class="ma-0"
-            :color="hasNotes ? 'warning' : 'grey darken-1'"
+            :color="hasNotes ? 'white' : 'grey darken-1'"
             v-on="on"
             @click="openNotesDialog"
           >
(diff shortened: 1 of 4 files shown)


BryonLewis:
Make sure that is also exported in the TeackJSON format to ensure parity.  Check the kwcoco export and see if that can be added in as well.

BryonLewis:
While reviewing I remembered that we didn't have note editing/viewing in the standard mode, I added an icon and a dialog for editing notes to the track.
image

image

image

BryonLewis (approved):
Doubled checked and TrackJSON/COCO have the notes export/import options.
Updated the docs for data formats to include the notes information.
Noticed that only the bottom view have a 'notes' editor.  Added to the standard view a Notes icon that will be highlighted when a note exists.  When hovering it will display the text of the note and then when clicked a dialog will appear for editing the note.

PR #1754, opened 2026-07-13
Support launching DIVE Desktop on a dataset from the command line

mattdawkins:
## Summary

Adds `--import` / `--annotations` / `--name` so DIVE Desktop can be started directly on media plus an annotation file, skipping the import wizard:

```bash
dive-desktop --import input_list.txt --annotations detections.csv --name "Sea Lions"
```

`--import` accepts anything the wizard does (an image-sequence directory, an image-list text file, or a video); `--annotations` accepts a VIAME CSV or DIVE JSON.

This came out of wanting to review detector output over an image list without clicking through the import wizard each time.

## Behavior

The import runs through the same backend calls as the wizard (`beginMediaImport` → `finalizeMediaImport` → `dataFileImport`), so the result is an ordinary dataset — it lands in the recents list and can be reopened later, exactly as if it had been imported by hand. Media that needs transcoding is converted first, and the viewer opens when the conversion job completes.

## Two things worth a look in review

**Renderer pulls, main process does not push.** The renderer asks for the pending dataset (`desktop:cli-open-pending`) once it is mounted. Settings are owned by the renderer and handed to the background process at startup, and the import needs them (for `dataPath`), so the request waits on that handoff — otherwise the backend throws `settings has not been initialized`. Pulling also avoids racing the listener registration against an import that finishes before the window is ready.

**Second-instance forwarding.** A second launch carrying `--import` now opens that dataset in the running window instead of being dropped by the single-instance lock. The args are parsed from the raw `process.argv` and handed over via `requestSingleInstanceLock(additionalData)`, because the argv Chromium forwards to `second-instance` hoists switches ahead of positionals and thereby detaches `--import` from its value:

```
["dive-desktop","--import","--annotations","--name", ...switches..., "/path/list.txt","/path/x.csv","Sea Lions"]
```

Window creation is now guarded on actually holding the lock. Previously a losing instance raced its own `app.quit()`, and with the "already running" dialog no longer blocking that race it would build a window mid-teardown and report a spurious "application failed to start".

## Testing

- New unit tests for the argument parser (6 cases, including `=` syntax, short flags, relative-path resolution, and not swallowing a following flag as a value).
- Built with `build:electron:dir` and exercised end to end on a 3-image list with a VIAME CSV of 570 polygon detections: dataset imports, viewer opens, annotations and polygons render, dataset appears in recents. Second-instance forwarding verified with a separate CSV opening into the already-running window.
- `platform/` suite passes apart from two `web-girder/multicamFileRegistry.spec.ts` failures that reproduce on a clean `main` and are unrelated to this change.

### DIFF
--- a/client/platform/desktop/README.md
+++ b/client/platform/desktop/README.md
@@ -71,9 +71,32 @@ Relative paths are resolved against the working directory. For example, to revie
 dive-desktop --import input_list.txt --annotations detections.csv --name "Sea Lions"
 ```
 
-The dataset is imported through the same backend calls as the wizard (`beginMediaImport` → `finalizeMediaImport` → `dataFileImport`), so it is a normal dataset: it is added to the recents list and can be reopened from the dataset list later. Media that requires transcoding is converted first, and the viewer opens when the conversion job completes.
+### Multi-camera and stereo
+
+Name each camera with a repeated `--camera` instead of using `--import`:
+
+```bash
+dive-desktop --camera left=/data/left --camera right=/data/right \
+             --annotations left=/data/left.csv --annotations right=/data/right.csv \
+             --calibration /data/calibration_matrices.npz
+```
+
+* **`--camera`, `-c`** — `<name>=<media>`, repeated once per camera (two or more). Each media path is the same set of things `--import` accepts. Only the first `=` separates, so Windows paths survive. Flag order is the display order.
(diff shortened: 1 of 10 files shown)


BryonLewis (approved):
I've tested locally for a few formats and made some updates:

- Added coumentation into the docs for the system
- Fixed some minor issues with the second-instance
- prevented auto navigating away from a dataset if there are unsaved annotations (follows the other navigation guards)
- If a dataset requires transcoding it can't open instantly, I provide a dialog message to the user indicating that it is transcoding and will open the dataset when complete

PR #1756, opened 2026-07-14
Infer the KAMERA format in multicam and batch multicam import

romleiaj:
KAMERA writes each swathe to a flat view folder (left_view/center_view/ right_view) holding up to three modalities side by side, distinguished by a filename suffix (*_rgb / *_ir / *_uv), plus sidecar metadata files. Both multicam import modes now detect this layout and create one dataset per view folder with one camera per modality present.

Cameras sharing a folder are selected by a new optional per-camera glob on MultiCamImportFolderArgs.sourceList, applied by findImagesInFolder on desktop and by staged-file filtering in the web upload path. Batch scan treats flat KAMERA folders as self-describing collects (skipping the shared camera-set validation) and defaults dataset names to <flight>_<view>; the parent-folder import auto-enables filename-based frame alignment since KAMERA modalities drop frames independently.


BryonLewis (approved):
Tested with an example dataset and everything seems to work properly for importing on both web and desktop.l

PR #1760, opened 2026-07-15
Suppression regions: hide and un-count detections beneath them

mattdawkins:
Adds a "Suppression Region Type" setting (`typeSettings.suppressionType`, default `Suppressed`, empty disables) selectable in the type-settings panel.

Any detection whose geometry lies >=50% under an annotation of that type on a frame is:
- hidden from every annotation layer, and
- excluded from the per-frame type counts and the frame-filtered track list.

Overlap uses each side's polygon when it has one and its bounding box otherwise, estimated by point sampling, so concave warped field-of-view polygons are handled without adding a polygon-clipping dependency.

Motivation: multi-camera surveys (e.g. sea-lion) emit "Suppressed" field-of-view regions where another camera already covers the ground; this lets a reviewer read de-duplicated counts off the interface.

### DIFF
--- a/client/dive-common/components/BottomPanel.vue
+++ b/client/dive-common/components/BottomPanel.vue
@@ -154,6 +154,7 @@ export default defineComponent({
             :show-empty-types="clientSettings.typeSettings.showEmptyTypes"
             :height="130"
             :width="300"
+            :header-height="50"
             :style-manager="trackStyleManager"
             :filter-controls="trackFilters"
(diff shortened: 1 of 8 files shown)


BryonLewis (approved):
Adjusted some styling a bit to better fit the new option
Added an indicator in the type list to which type is being supressed with a tooltip explaining it
fixed a small issue with the size of the type list in the bottom bar view.

PR #1762, opened 2026-07-16
Resolve annotation sets in setup so the Import menu cannot throw

mattdawkins:
## Problem

`ImportAnnotations` calls `useAnnotationSets()` — and therefore `inject()` — from *inside* the `sets` computed:

```js
const sets = computed(() => {
  const data = useAnnotationSets();   // inject() in a computed getter
  const temp = cloneDeep(data.value);
  temp.push('default');
  return temp;
});
```

`inject()` resolves against the current component instance. Outside `setup()`, Vue 2.7 only sets that during render (`core/instance/render.ts` calls `setCurrentInstance(vm)` around `_render`). So the getter works when the template evaluates it, and throws for every other caller:

```
Missing provided object for symbol Symbol(annotationSets): must provideAnnotator()
```

The Import menu's contents are lazy — `v-menu` doesn't render them until opened — so the first read of `sets` already happens off the render path. Any future watcher, effect, or programmatic read hits the same throw. It survives today only by accident of *when* Vue happens to evaluate the computed, which is not a property this code controls.

This was found while debugging a separate issue: reading `sets` from a debugger (i.e. outside a render pass) threw the error above on unmodified `main`.

## Fix

Hoist the lookup into `setup()` and close over the resulting ref. Behaviour is unchanged: clone the provided sets, append `default`, leave the provided array untouched.

## Test

Adds `importAnnotationsSets.spec.ts`, which mounts the component with the menu closed — reproducing the off-render read — and asserts `sets` resolves.

Verified the test is meaningful rather than vacuous: against the previous code all three cases fail with

```
AssertionError: expected [Function] to not throw an error but
'Error: Missing provided object for symbol Symbol(annotationSets)…' was thrown
```

## Verification

- `npm test` — 609 passing (main's suite plus the 3 new cases)
- `npx eslint` on both changed files — clean
- `npm run build:web` — clean

## Note for reviewers

`npm test` requires Node ≥20 since #1758 bumped vitest to 4 (`engines: ^20 || ^22 || >=24`); it fails to start on Node 18 with `ERR_REQUIRE_ESM`. `client/README.md` still says "Requires Node 18+" — stale, but out of scope here.

There is a second, separate defect in this file that I deliberately did **not** touch: `const currentSet = ref(defaultSet || 'default')` (line ~143) is missing `.value`. `defaultSet` is a ref, so it's always truthy and `ref()` passes an existing ref straight through — `currentSet` therefore aliases the injected `annotationSet` ref rather than copying its value, and the `|| 'default'` fallback never applies. Fixing it would change behaviour (the `Annotation Set` combobox is gated on `currentSet !== ''`, so it would start rendering where it currently doesn't), so it wants its own PR and a decision about intent.

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -8,7 +8,7 @@ This directory contains the code for:
 
 ## Development
 
-Requires Node 18+.
+Requires Node 20+.
 
 ``` bash
 # install dependencies
--- a/client/dive-common/components/ImportAnnotations.vue
+++ b/client/dive-common/components/ImportAnnotations.vue
@@ -143,7 +143,8 @@ export default defineComponent({
       return temp;
     });
     const defaultSet = useAnnotationSet();
-    const currentSet = ref(defaultSet || 'default');
+    // Local copy (not an alias of the injected ref); '' means default.
+    const currentSet = ref(defaultSet.value || 'default');
     const { prompt } = usePrompt();
     const processing = ref(false);
     const menuOpen = ref(false);
@@ -475,7 +476,6 @@ export default defineComponent({
               </v-btn>
             </v-row>
             <v-row
-              v-if="currentSet !== ''"
               class="mt-3"
               dense
             >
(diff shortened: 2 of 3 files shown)


BryonLewis (approved):
fixed the other issue referenced in the PR description and updated some documentation about node versions.

PR #1765, opened 2026-07-17
Out of Order Annotation FPS Fix

BryonLewis:
FPS metaddata in CSV's were importing previously but if the conversion process happened quick enough it would overwrite the FPS in the metadata to be the video FPS.  This could be see in video imports when using the assetStore importing

This PR:

- Updates the assetStore script to make it so the fps in the metadata for viame-csv is a subset of the video's fps for testing purposes
- the `convert_video` and `convert_images` now check the DIVE Datasets FPS value before setting it to the video's native FPS
-

### DIFF
--- a/samples/scripts/assetStoreImport/README.md
+++ b/samples/scripts/assetStoreImport/README.md
@@ -15,9 +15,10 @@ The script generateSampleData.py creates a folder structure containing:
 
 - Videos: MP4 format, H.264 codec, random duration (5–30 seconds), 1280x720 resolution.
 - Image Sequences: Extracted frames from temporary videos, stored as sequential JPGs.
-- Annotations: Each video or image sequence is accompanied by either a DIVE track JSON (`.json`) or VIAME CSV (`.csv`) file describing moving or scaling geometric shapes (rectangle, star, circle, diamond) per frame. The format is chosen at random per dataset.
+- Annotations: Each video or image sequence is accompanied by a DIVE track JSON (`.json`), VIAME CSV (`.csv`), or COCO JSON file describing moving or scaling geometric shapes (rectangle, star, circle, diamond) per frame. The format is chosen at random per dataset.
     - Videos: the annotation file has the same basename as the video, with extension `.json` or `.csv`
     - Image Sequences: any `.json` or `.csv` file in the same folder as the frames is imported as annotations
+    - **VIAME CSV videos**: annotation FPS is a random even subsample of the video FPS (30 → e.g. 1, 5, 10, 15, 30). The CSV `# metadata` `fps` field and frame count match that rate. Filenames include `_annfps{N}` (e.g. `reef_annfps10.mp4` / `reef_annfps10.csv`) so you can verify import respects CSV FPS.
 
 Usage
 ```bash
@@ -29,7 +30,14 @@ uv run --script generateSampleData.py
 - --folders (-f): Number of top-level folders (default: 3)  
 - --max-depth (-d): Maximum subfolder depth (default: 2)  
 - --videos (-v): Maximum videos per folder (default: 2)  
-- --total (-t): Total number of datasets (videos or image sequences) to create (default: 10)  
+- --total (-t): Total number of datasets (videos or image sequences) to create (default: 10)
+- --annotation-formats: Comma-separated formats (`dive-json`, `viame-csv`, `coco-json`). Default mixes `coco-json,viame-csv`. Use `viame-csv` alone to focus on CSV FPS import testing.
+
+Example focused on CSV FPS testing:
+```bash
+uv run --script generateSampleData.py --annotation-formats viame-csv --total 8
+uv run --script minIOConfig.py
+```
 
 The script will randomly generate videos or image sequences with annotations inside the output directory.  The output directory defaults to ./sample
 
@@ -90,7 +98,7 @@ Example Output
     - Check the http://localhost:8010/girder#jobs page to see that importing jobs have completed
 10.  Finally you can go to the regular DIVE interface (http://localhost:8010) click on the globe in the breadcrumb bar at the top of your user directory and navigate to your collection
     - You should be able to open DIVE Datasets and they should have random annotations in them.
-    - Video Datasets should have an annotation speed relative to the video and image-sequences should default to 1FPS
+    - Video datasets with VIAME CSV should keep the annotation FPS from the CSV `# metadata` line (see `_annfps{N}` in the filename). Other videos use video FPS. Image sequences should default to 1FPS.
 
 
 Notes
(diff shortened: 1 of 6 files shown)


kelseylmartin:
@brendatran 
Thank you so much Bryon. Has this update been pushed? I just tried running a 5 fps videos and got the same result (attached) where it's still defaulting to the video fps:
image

BryonLewis:
yeah it has been pushed, you may need to rebuild images or pull images again to resolve it.  I want to confirm these are items being imported through the assetStore importing?

kelseylmartin:
Yes, I'm importing through the AssetStore. I've reimported and am still getting the same result. It appears that the metadata FPS information is not in the csv for our tracker model output. I thought that this had been resolved but it's possible that we need to redocker our models?
@csbrown-noaa @brendatran May have more insight.

csbrown-noaa:
@kelseylmartin are you doing this in GCP?  It will take some time to pull the new image into the cloud environment.  
I think we pushed the new viame-vertex-docker but not the new DIVE.

BrendaTran:
I have not pulled the fix yet. Will give you an update here.

kelseylmartin:
Hi Bryon, the FPS info is now appearing in the csv but it's still not being applied to Dive via the girder and I'm getting the same result as above. Attached is the csv that I'm using.
[2023-NCD-005_1.csv](https://github.com/user-attachments/files/30242567/2023-NCD-005_1.csv)

BrendaTran:
Hi @BryonLewis, the FPS  appears in the csv but it's not correctly applying it via the girder import.

csv_fps

https://github.com/user-attachments/assets/e1cb60e1-d3d1-449c-affa-7e4c0918edfc

[bigfish.csv](https://github.com/user-attachments/files/30242623/bigfish.csv)

BryonLewis:
This didn't properly run the conversion process.  It still has the MarkForPostProcess set to `true` meaning it either didn't kick off the batch processing at the end of import or something else.

I don't know about this specific server and how it is configured but it may be related to #1766 where if the worker isn't within the same docker network or doesn't have acces to the girder api url it may not run the localworker jobs properly.

csbrown-noaa:
Is any of that actually needed here?  The annotation fps is right there in the csv file....  The contents of the video file doesn't seem to be important...

BryonLewis:
If you don't have an FPS set in the CSV I need to set the fps to something reasonable and I can't set that without knowing the fps of the video itself.  I also want to avoid an annotation setting the FPS to some rate that is higher than the video fps because then the frame counts would be completely off.  I have a video explaining in more detail what is happening after an asset store import and how it relates to the conversion process.

https://github.com/user-attachments/assets/a9bb9c67-5ee1-4613-994e-3a163f82d680

csbrown-noaa:
> If you don't have an FPS set in the CSV I need to set the fps to something reasonable and I can't set that without knowing the fps of the video itself. 

-- sure, but, we do have the FPS set in the CSV.  So, while fallback behavior is important, it isn't applicable here.

> I also want to avoid an annotation setting the FPS to some rate that is higher than the video fps because then the frame counts would be completely off.

-- I understand the sentiment here, but if the fps of the annotations is not reliable, then the whole thing is 💩'd anyway, so /shrug



To clarify, our DIVE instance is not backed by the hundreds of GPUs we would need to meticulously read and muck about with our petabytes of video files, and we have intentionally avoided doing this.  Our DIVE is backed by a simple CPU that cannot possibly hope to handle the compute loads of ffmpegging every single video we need to review.  We are handling outside of the VIAME/DIVE ecosystem the hard work of getting the videos prepped for streaming.  They're all h.26x'd and ready to go.

Do you have other suggestions for how we can work around this?  This seemingly minor issue has completely broken our bridge between running the predictions with VIAME CLI in our CloudBatch scaled infra and then displaying the results in our lightweight DIVE deployment.

BryonLewis:
It's not a heavy task (less than a second if the file has already been converted) but I do understand that when dealing with 10's of thousands of videos it could be annoying to have to ffprobe each one to get information about the video.

I could create an admin configuration setting that will prevent launching jobs on asset store imports and will just assume all of the videos/files imported are safe for running in DIVE.  I obviously then wouldn't know the video fps, but I could force it to just use any related annotation FPS value.  It just means that you could potentially have broken videos if they haven't been converted properly.  DIVE will attempt to show them and you'll get an error in the annotation viewer.

csbrown-noaa:
> It's not a heavy task (less than a second if the file has already been converted)
 
 Hmmmm..... will do some digging and discussion with the team.  OTOH, I think we may actually be trying to avoid that our specific DIVE deployment ever even tries to convert any video files, since a scientist could inadvertently DOS the server by simply trying to load in a handful of videos if they hadn't gotten their conversion pre-processing just right - would be better to just fail hard and fast.  Will come back to this discussion after we formulate a plan.  Thanks @BryonLewis for all of the insights.

BrendaTran:
@BryonLewis, I noticed something changes during Girder import process. After importing the above bigfish.mp4 and bigfish.csv to Girder, if I wait a bit until the transcoded file gets generated, I saw Girder changed fps from -1 to 30 and the original fps showed 30. However, our imported csv has fps 20. Still Girder doesn't match FPS from the imported csv

image

BryonLewis:
I did merge another PR: https://github.com/Kitware/dive/pull/1792

I found this when testing with much larger CSV files.  There was a condition where the conversion would happen quicker than processing the CSV file so it the CSV could clobber the conversion results which would cause it to not be registered as a valid DIVE dataset.

I don't believe that is the issue you're having though because even in my testing with the bigfish.csv (I couldn't download or see the bigfish.mp4 but created a sample one) it would still properly set the fps to 20 and not 30 or the video FPS.

It's going to take a bit to rebuild the latest containers but I would make sure that you have the latest dockerhub kitware/viame-worker:cpu and kitware/viame-worker:gpu as well as kitware/viame-web:latest.  That if you aren't building the containers yourself locally with the source code.

If you are using the docker images from docker hub try running: `docker images kitware/viame-worker` and provide the ID of the images to confirm they are a more recent one.
May also be usefule to know `docker images kitware/viame-web` to know the web container Id as well.

BrendaTran:
Hi @BryonLewis, confirmed that csv fps is working.

image

kelseylmartin:
worked for me as well. Thank you Bryon!

BryonLewis:
No problem, sorry about the lengthy back and rroth and the confusion. Thaks for having patience.  It's a bit headspinning jumping between the different functionality of DIVE/VIAME (going from stereoscopic/multicam dev to the server infrastructure/devOps and then following the ingestion logic between Desktop/Web for the User Interface vs Server Side Object storage imports.

PR #1769, opened 2026-07-20
Add the functionality to automatically register images in the "Camera Registration" tab

romleiaj:
Uses MINIMA-LoFTR under the hood to compute features and match points - a model tuned to work across modalities (EO/IR)


romleiaj:
See https://github.com/VIAME/VIAME/pull/272

romleiaj:
# Branch summary: `dev/auto-image-registration`

  **What it adds:** an "Auto Register" button in the Camera Registration panel that computes a cross-modality (e.g. EO↔IR) homography between two cameras automatically, using a deep feature matcher (MINIMA-LoFTR) hosted in the VIAME **desktop** interactive service. ~550 added lines over 14 commits.

  ## How it workds

  - **Interactive service / IPC** (`interactive.ts`, `ipcService.ts`): a new `register_images` wire command sends two image paths (plus optional RANSAC/matching thresholds) to the Python service, which returns a 3×3 homography, the inlier correspondences, and quality stats. The matcher model loads lazily on first use and stays resident. A companion `register-images-available` handler probes whether the matcher weights exist in the VIAME install; without them the button is hidden entirely. The boundary is typed end to end: both the IPC handler and the service method use the shared `AutoRegisterRequest`/`AutoRegisterResponse` types, with a local interface for the service's snake_case wire payload.
  - **API contract** (`apispec.ts`): new `AutoRegisterRequest`/`AutoRegisterResponse` types, including machine-readable failure codes (`insufficient_matches`, `low_confidence`, `degenerate_homography`).
  - **Platform bridge** (`useAutoRegister.ts`, `Viewer.vue`, `ViewerLoader.vue`): a provide/inject service keeps the shared UI platform-agnostic — desktop's `ViewerLoader` passes a handler prop that resolves each camera's image path for the current frame and calls the service; on web nothing is provided, so the panel shows no button. Along the way, the stereo-only image-path-getter setup in `ViewerLoader` was refactored into a shared `populateMultiCamImagePathGetters`, because Auto Register works on **any** multi-camera dataset, not just stereoscopic ones (this does touch stereo code).
  - **UI + store** (`RegistrationTools.vue`, `CameraRegistrationStore.ts`): the button runs the matcher on the current frame's pair, confirms before replacing any existing points/transform, then injects the returned inliers as **ordinary correspondence points** — so the user can inspect, drag, or delete them exactly like hand-picked ones — switches the pair to a homography fit, and refits. Point editing is auto-enabled so the injected points are immediately visible. Results surface as a success summary (point count + match consensus) or an actionable error.
  - **Copy changes:** the feature was renamed Auto Align → Auto Register (UI, IPC, and wire names all updated), and the "Pick points" toggle became "Edit points." The button reads "Auto Register" with the current-frame qualifier in a tooltip. The rig status now says "N/M cameras **ready**" instead of "registered" (the count includes the reference camera, which is ready by definition but was never registered to anything) — this touches the panel summary, the Align View tooltip, and the import menu hint.

  ## Tests

  `CameraRegistrationStore.spec.ts` gains coverage for `applyAutoRegistration` (point injection, homography switch, source non-restamping), and `alignedViewTooltip.spec.ts` is updated for the "cameras ready" wording. The service/IPC layer and Vue wiring are untested — the Python matcher side lives in the separate VIAME/service repo, not this diff.

romleiaj:
Close in favor of #1854

PR #1774, opened 2026-07-21
Import menu: clearer annotation-set label, no overlap with Import button

mattdawkins:
- Rename the Import menu's set combobox label to `Annotation Set Name`
- Increase the row's top margin so the floating outline label no longer overlaps the Import button


BryonLewis (approved):
LGTM

PR #1780, opened 2026-07-21
Pass DIVE camera registrations to 2-cam/3-cam VIAME pipelines

romleiaj:
Multicam pipelines previously received no transform information from DIVE; the old h5 workflow relied on paths hardcoded inside the pipes. Desktop and web now hand the dataset's camera registration to the pipeline's warp processes, mirroring the stereo calibration flow:

- One standard <camera>_to_<reference>_registration.json per non-reference camera is written into the job work dir; warpN (matching the pipeline camera position) receives its file via -s warpN:transformation_file=..., with the pair and direction pinned through the reader's from_camera/to_camera config since a pair may be stored in either orientation.
- Pipeline camera order is the registration reference first (the pipes warp everything onto camera 1's frame), then display order. Which detector a pipe runs on which input is the pipe's documented contract, not something DIVE infers.
- Only pairs registering a camera directly onto the reference are supported; pairs between two non-reference cameras are explicitly unsupported for pipelines and never reach the job. Cameras with points-only (unfitted) pairs get no settings.
- Web sends the registration as a multicam_registration job param built from dataset meta; the worker writes the files.

Requires VIAME's dive transform_2d_io reader and warp_detections process.


romleiaj:
Paired with https://github.com/VIAME/VIAME/pull/278.

romleiaj:
# Pass DIVE camera registrations to 2-cam/3-cam VIAME pipes

Multicam pipes never got any alignment info from DIVE — the transform paths were
hard-coded in the pipe files. Worse, which of your cameras fed which pipe input
was guessed from display order, so a thermal detector could quietly end up
running on the wrong camera.

This sends the dataset's registration to the pipe and turns the camera order
into an explicit step you confirm before the run.

## What's new

**Assign cameras dialog.** Running a 2-cam or 3-cam pipe now opens a dialog with
one row per pipe input, prefilled with DIVE's best guess. Pipes say what they
want with a `# Camera Order: EO, UV, IR` header; datasets learn which camera is
which from the camera and image names at import. You can swap rows around, and
empty or duplicate slots block Run. Confirming saves the answer back to the
dataset so next time is already filled in.

**Registration hand-off.** DIVE writes one registration file per camera into the
job directory and points the pipe's warp steps at it. The first camera in the
run order is what everything else warps onto; camera pairs that were never
registered are skipped. Web works the same way — the job carries the
registration and the worker writes the files. (Before this, the worker crashed
outright on v2 datasets.)

**Blocked before it fails.** DIVE now knows which inputs a pipe warps, so if a
camera has no registration onto the first one, the dialog says so per row and
refuses to start — "Register X → Y in the Camera Registration tab first" —
instead of letting the pipe die on a missing file. Desktop and web both do this.

## Testing

Needs VIAME from `main` and a registered multicam dataset.
`utility_warp_annotations_2-cam.pipe` is a model-free test pipe.

- **Assign dialog** — run a 2-cam pipe with a `# Camera Order:` header. Slots
  prefill by role, swapping works, empty/duplicate blocks Run, roles get saved.
- **Hand-off** — with IR registered onto EO, run. The job dir should have
  `<IR>_to_<EO>_registration.json` (`"version": 2`) and the command should have
  `-s warp2:transformation_file=…`.
- **Missing registration** — clear the IR→EO registration and run. The dialog
  blocks with "Register IR → EO … first"; the API returns the same.
- **Web** — same hand-off on web-girder; the worker writes the file and the pipe
  finishes.
- **No header** — run a pipe with no `# Camera Order:`. You should get bare
  input1/input2 rows in the dataset's stored order.

## Notes

Standalone on `main`, rebased onto be70fba7. Uses the v2 registration format
from `dev/auto-register-pipeline`, which should land around the same time.
Companion VIAME branch: `dev/dive-registration-json`.

The old automatic header-matching path is gone — the dialog always supplies the
order now. `docs/Pipeline-Import-Export.md` covers the header and the dialog.

21 files, +1539/-10. Commits 0adb257c through 6769ed72.

BryonLewis (approved):
I've made some modifications:

- If there is no cameraRoles I have the dialog fall back to using the camera order by default (with the option to change it).  For things that aren't EO/UV/IR it was a bit annoying to choose a camera order if one is already set by my import.  It will still use the cameraRoles if they are found but will fall back to default displaying the displayOrder for the cameras
- The cameraRoles weren't properly being saved to metadata in the desktop like they were in the Web version so updated that
- there was some differences in the fitted registration calculation for server vs desktop
- Updated some docs to reflect the role/name matching and enure that role matching will override the default display order for the dialog
- Some better warning for a mismatch saving of roles

PR #1781, opened 2026-07-21
Fix DIVE config import not sticking on multicam datasets (desktop)

mattdawkins:
## Problem

Importing a DIVE configuration file (custom type colors, confidence thresholds, ...) works in single-camera mode but silently does nothing in multi-camera mode: neither the thresholds nor the colors apply or persist.

## Cause

In multi-camera mode the desktop viewer hands the importer the camera-qualified dataset id (`<base>/<camera>`), so `dataFileImport` wrote the config into that camera's `meta.json`. The viewer, however, only reads `confidenceFilters` / `customTypeStyling` (and the other mutable config keys) from the **base** dataset's metadata, so the imported values were never seen.

## Fix

When an import targets a camera of a multicam dataset and produces mutable-config keys, `dataFileImport` now also merges those keys into the base dataset's `meta.json`, using the same overwrite/additive `datasetInfo` semantics as the per-camera write. Since `reloadAnnotations` re-reads the base metadata after import, the imported config now applies immediately as well.

Includes a regression test covering the camera-targeted config import.

### DIFF
--- a/client/dive-common/apispec.ts
+++ b/client/dive-common/apispec.ts
@@ -226,6 +226,21 @@ interface DatasetMetaMutable {
   error?: string;
 }
 const DatasetMetaMutableKeys = ['attributes', 'confidenceFilters', 'timeFilters', 'imageEnhancements', 'customTypeStyling', 'customGroupStyling', 'attributeTrackFilters', 'datasetInfo', 'cameraHomographies', 'cameraCorrespondences', 'cameraTransformTypes', 'cameraRegistrationSource'];
+/**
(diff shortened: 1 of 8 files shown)


BryonLewis (approved):
Updated this PR to:
- Be more explicit about the merging process for the configurations.  I didn't like that camera config files could overwrite or merge imageEnhancements (those should be set per camera)  or other fields that are technically metadataMutable keys.
- Duplicated the desktop functionality of merging configs to the web version.

PR #1782, opened 2026-07-21
Fill only the polygon, not the rectangle, when a detection has both

mattdawkins:
## Problem

With the per-type "Fill" style enabled, a detection that has a polygon gets both its polygon *and* its bounding rectangle filled, shading the whole box area.

## Fix

When a detection has polygon geometry, the fill setting now applies solely to the polygon; the bounding rectangle stays unfilled. Detections with only a rectangle keep the existing fill behavior. The rectangle layer already computed a per-detection `hasPoly` flag, which the fill style function now respects.


BryonLewis (approved):
Updated to make it so that if it has another polygon but the polygon visibility is off it will apply the fill on the rectangle.  Did this because we don't have a concept of fill for different annotation types (just one fill checkbox) so if you turn off polygon visibility it makes sense that the rectangle has the fill enabled afterwards.

PR #1783, opened 2026-07-21
Trim a detection's polygon when its bounding box is shrunk over it

mattdawkins:
## Problem

Resizing a detection's bounding box leaves its polygon untouched, so shrinking the box past the polygon leaves polygon geometry sticking out beyond the box.

## Fix

Box edits now clip any polygon on the detection to the new bounds using a small Sutherland–Hodgman implementation (`clipPolygonToBounds` in `src/utils.ts`, no new dependency):

- Polygons crossing the new box edge are trimmed to fit.
- A polygon falling entirely outside the new box is removed.
- Holes are clipped by the same rule; holes left with fewer than 3 points are dropped.
- Clipping happens against the rounded bounds that `setFeature` stores, so the trimmed polygon is consistent with the persisted box.
- Rotated boxes are exempt: their stored bounds are the unrotated envelope, so an axis-aligned clip would cut the wrong region.

## Redundant full-box polygons are dropped

A polygon identical to the detection's bounding box carries no information beyond the box and makes editing harder, so (second commit):

- When a box shrink clips a polygon down to exactly the new box, the polygon is **removed** rather than kept as a redundant rectangle.
- On load, `Track.fromJSON` drops any polygon equal to its feature's bounds (extra collinear vertices and either winding still count as equal; polygons with holes or any interior structure are kept, and other geometry like head/tail points is untouched).

The equality test (`polygonEqualsBounds`) requires a hole-free polygon whose extent matches the bounds and whose area fills the box — a simple polygon with the area of its bounding box must be that rectangle.

Unit tests cover containment, single/multi-edge clipping, slanted-edge intersection points, full removal, hole handling, box-equality detection (including clip artifacts and near-miss shapes), and the load-time cleanup.

### DIFF
--- a/client/dive-common/use/useModeManager.spec.ts
+++ b/client/dive-common/use/useModeManager.spec.ts
@@ -13,6 +13,7 @@ import { IDENTITY3 } from 'vue-media-annotator/alignedView/alignedView';
 import type { Matrix3 } from 'vue-media-annotator/alignedView/homography';
 import type { AggregateMediaController } from 'vue-media-annotator/components/annotators/mediaControllerType';
 import type { AnnotationId } from 'vue-media-annotator/BaseAnnotation';
+import { ROTATION_ATTRIBUTE_NAME } from 'vue-media-annotator/utils';
 import useModeManager from './useModeManager';
 
 function translation(tx: number, ty: number): Matrix3 {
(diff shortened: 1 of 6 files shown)


BryonLewis (approved):
minor modification to avoid clipping any annotations that have been rotated.

PR #1784, opened 2026-07-21
Allow right-click edit entry when box display is turned off

mattdawkins:
## Problem

With box display toggled off, right-clicking a detection to enter edit mode stops working (e.g. finalize a polygon, turn off box display, then right-click another detection — nothing happens until boxes are re-enabled).

## Cause

Turning off box display called `rectAnnotationLayer.disable()`, which empties the layer's feature data. That layer's rectangles are the click target covering every detection, and geojs only fires feature mouse events on existing features — so a detection without a polygon had nothing left to right-click.

## Fix

When box display is off, the rectangle layer now keeps its features as **invisible hit targets**: nothing is drawn (no stroke, no fill, no rotation arrow), left-clicks and hover tooltips ignore them, but right-click still enters edit mode. Editing is now reachable regardless of which display types are visible, and a duplicate emit from an overlapping polygon is already deduped by the click handler's same-tick guard.


BryonLewis (dismissed):
LGTM, did some testing and it resolves the bbox off issues.  May have to revisit some stuff in the future regarding overlapping.

PR #1785, opened 2026-07-21
Display detections flagged with a suppression attribute as suppressed

mattdawkins:
## Behavior

Suppression regions (#1760) hide detections lying under a region. A detection can also be marked suppressed by an *attribute* — an attribute named after the suppression type ("Suppressed" by default) set true by a pipeline or user — while sitting outside any region. Previously such a detection displayed as its normal type and counted normally, with nothing indicating its suppressed status.

A detection carrying the suppression attribute (on the track, or on the detection at the current frame) is now treated as a display-only special case:

- **Stays visible** (region suppression is unchanged: detections under a region are still hidden).
- **Real type preserved**: canvas label, hover tooltip, and details panel all keep the actual type (editable). The attribute only marks display/counts, never the stored type.
- **Eye-off tag**: canvas text labels and hover tooltips can show an `mdi-eye-off` icon next to the type (Annotation Visibility → "Show suppressed tags").
- **Configurable styling** (Annotation Visibility → Suppression): dashed outline, outline opacity, fill color, and fill opacity. Disabled styling still leaves the detection visible with its normal type color.
- **Doesn't count toward its own type**: excluded from the type list's dataset totals and per-frame counts. Not credited to the suppression type either, and never appears as a composite type-list category.

## Also included

- The suppression-region overlap threshold is configurable (Type Settings → "Suppression Overlap %"), defaulting to **99%**.
- Recognized attribute values: `true`, nonzero numbers, and `'true'` / `'yes'` / `'on'` / `'1'` in any case. On interpolated frames the check falls back to the previous keyframe's detection attributes.
- Unit tests cover attribute parsing, threshold normalization and behavior.

### DIFF
--- a/client/dive-common/components/AnnotationVisibilityMenu.vue
+++ b/client/dive-common/components/AnnotationVisibilityMenu.vue
@@ -295,6 +295,7 @@ export default defineComponent({
                   >
                     mdi-eye-off
                   </v-icon>
+                  <span class="ml-1 text--secondary">(labels & tooltips)</span>
                 </span>
               </template>
             </v-checkbox>
@@ -511,6 +512,7 @@ export default defineComponent({
                   >
                     mdi-eye-off
                   </v-icon>
+                  <span class="ml-1 text--secondary">(labels & tooltips)</span>
                 </span>
               </template>
             </v-checkbox>
@@ -627,9 +629,9 @@ export default defineComponent({
               </template>
               <span>
                 Detections flagged with a suppression attribute stay visible
-                but can be drawn with a dashed outline, custom fill, and
-                opacity so they are easy to distinguish from normal detections.
-                Enable this to apply those styles.
+                with their real type. When enabled, they can be drawn with a
+                dashed outline, custom fill, and opacity so they are easy to
+                distinguish from normal detections.
               </span>
             </v-tooltip>
           </div>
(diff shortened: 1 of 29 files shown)


BryonLewis:
## My changes on `feat/attribute-suppression`

Follow-up work on attribute suppression display, visibility controls, and docs (on top of the earlier region/attribute suppression work).

### Configurable suppression visibility & styling
- Added annotator preferences for **Show suppressed tags** and **Suppression display** (dashed outline, outline/fill opacity, optional fill color).
- Wired these into the Visibility menu (editing bar) and applied them in rectangle/polygon/text layers.
- Attribute-suppressed detections can be visually distinguished without changing how region suppression still hides covered detections.

### Keep real type + clearer suppressed cues
- Attribute-suppressed detections **keep their real type** for coloring and labels (no longer shown as `Type - SuppressionType`).
- Show an **eye-off** glyph on canvas labels and hover tooltips when suppressed tags are enabled.
- Type-list suppression tooltip now shows the **configured overlap threshold** and clarifies attribute behavior.
- Updated settings/help copy to match region vs attribute suppression.

### Documentation
- Added [`docs/UI-Suppression.md`](docs/UI-Suppression.md) covering region suppression, attribute suppression, and visibility/styling.
- Cross-linked from Overview, Annotation View, Attributes, Type List, Track List, Navigation/Editing Bar, and `mkdocs.yml`.

### Commits
- `bf86a765` — configurable settings for suppression visibility
- `b03143b2` — Keep attribute-suppressed detections on their real type with eye-off cues
- `5ab0644c` — Document suppression regions, attributes, and visibility controls

BryonLewis (approved):
Modified it to simplify the view of supression and also add configuration to how suppressed tags are rendered.  Updated documentation as well for the new settings/configuration.

PR #1786, opened 2026-07-21
Desktop: support multicam filter pipelines (filter_*_2-cam / 3-cam)

mattdawkins:
## Problem

Pipeline categories come from the pipe filename, and a `2-cam`/`3-cam` suffix overrides the prefix — so a multicam *filter* pipeline (e.g. VIAME's new `filter_register_frames_2-cam.pipe` and `filter_compute_tile_mosaics_2-cam.pipe`, which write registered/blacked-out images and per-timestep tile mosaics) is typed `2-cam`, never matches the `type === 'filter'` checks, and consequently: its image writers get no output-directory prefix, its (nonexistent) CSVs are ingested, and no new dataset is created from the output images.

## Fix

Desktop `runPipeline` now recognizes filter/transcode output handling from the pipe filename prefix as well as the resolved type:

- `createsNewDataset` / `isFilterPipe` are true for `filter_*.pipe` regardless of the `-cam` category, enabling output-dir creation, CSV-ingest skip, image-writer prefixing, frame renumbering, and new-dataset import.
- The output prefix is additionally passed to `image_writer2` / `image_writer3`, the per-camera writers multicam filter pipes use (extra `-s` keys for absent processes are ignored by the kwiver runner).

Web/girder pipeline running is untouched.

### DIFF
--- a/client/dive-common/apispec.ts
+++ b/client/dive-common/apispec.ts
@@ -79,6 +79,8 @@ interface PipelineRuntimeParams {
 interface PipelineParams {
   kwiverParams?: Record<string, string>;
   runtimeParams?: PipelineRuntimeParams;
+  /** Desktop filter/transcode pipelines: name for the newly created dataset. */
+  outputDatasetName?: string;
 }
 
 interface Pipe {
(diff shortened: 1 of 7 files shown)


BryonLewis (approved):
Basic running can be tested (with a failing job because of lack of official build).

I've added to the PR the ability for the interface to provide a new Dataset name when running these multicam filter pipelines.

PR #1787, opened 2026-07-22
Stereo Import menu: hide registration section, align Warp to All, interactive-stereo warp

mattdawkins:
Two related Import-menu changes for stereo datasets:

## 1. Hide the Import Registration section on stereo datasets

Per-camera registration transform import is meant for (non-stereo) multicam rigs; stereo rigs use the calibration (camera file) import instead. `registrationSupported` now also requires the dataset subtype not be `stereo`.

## 2. "Warp to All" alignment + interactive-stereo behavior

- The "Warp to All" checkbox now lines up with the "Overwrite" checkbox instead of sitting partially below it.
- On **stereo** datasets the option now operates through the interactive stereo service: when interactive stereo is enabled and a camera calibration is present, checking it warps every imported detection from the active camera to the other camera — head/tail lines transfer as lines (also computing the stereo measurement), polygons as polygons, plain boxes as boxes — skipping any frame the other camera already has, then saving. The per-annotation transfer machinery is reused with the auto-compute preference bypassed (the user asked for the warp explicitly).
- When interactive stereo is off or no calibration is loaded, the checkbox is **greyed out** with a hint explaining what's missing.
- **Multi-camera (non-stereo)** datasets keep the existing registration-based warp, gated on rig registration as before.

### DIFF
--- a/client/dive-common/components/ImportAnnotations.vue
+++ b/client/dive-common/components/ImportAnnotations.vue
@@ -529,22 +529,38 @@ export default defineComponent({
                 </template>
               </v-combobox>
             </v-row>
-            <v-row align="start">
+            <v-row
+              class="flex-nowrap"
+              align="center"
+            >
               <v-checkbox
                 :input-value="!additive"
                 label="Overwrite"
                 class="mt-2"
+                dense
+                hide-details
                 @change="additive = !$event"
               />
-              <v-checkbox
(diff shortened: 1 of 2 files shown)


BryonLewis (approved):
Minor changes for await/async stuff to wait properly instead of happening after dataset has already opened

PR #1788, opened 2026-07-22
Let clicks inside a box but outside its polygon hit nested annotations

mattdawkins:
## Problem

Right-clicking a small annotation that sits inside a larger annotation's bounding box sometimes edited the larger annotation instead. In particular, when the larger annotation has both a polygon and a rectangle and the click lands **inside its rectangle but outside its polygon** — directly on the smaller annotation — the larger one could still win the click (feature click ordering is closest-border, and the big box's edge can be nearer).

## Fix

While polygons are displayed, a detection that has a polygon now owns only its polygon shape for click targeting; its rectangle is treated as just the envelope:

- In `RectangleLayer`'s feature click handler, a click (left or right) that falls inside the detection's box but outside **every** one of its polygons is ignored by that feature, letting the click fall through to whatever is actually under the cursor — e.g. the nested smaller annotation.
- The detection's polygon rings are carried on the layer data (`polyRings`, native coords, transformed through the display transform for the containment test so the aligned/warped view works too).

Detections without polygons, and all behavior while polygon display is off (`drawingOther` false), are unchanged.

### DIFF
--- a/client/dive-common/apispec.ts
+++ b/client/dive-common/apispec.ts
@@ -79,6 +79,8 @@ interface PipelineRuntimeParams {
 interface PipelineParams {
   kwiverParams?: Record<string, string>;
   runtimeParams?: PipelineRuntimeParams;
+  /** Desktop filter/transcode pipelines: name for the newly created dataset. */
+  outputDatasetName?: string;
 }
 
 interface Pipe {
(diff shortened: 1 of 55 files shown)


BryonLewis (dismissed):
mistaken approval

BryonLewis (approved):
adjusted some logic changes to simplify and make sure that the closest border logic still works

PR #1789, opened 2026-07-22
Parse filename-like attribute values as full strings, not leading digits

mattdawkins:
## Problem

A detection attribute whose value is a filename like `0123ABC456` (numbers, letters, numbers — e.g. the `source_image` attributes the sea-lion suppressor writes) displayed as just `123`: only the first digit group survived import.

## Cause

- **Desktop CSV parser** (`_deduceType` in `serializers/viame.ts`) used `parseFloat`, which parses a leading numeric prefix (`parseFloat('0123ABC456') === 123`), replacing the string with a number.
- **Desktop attribute-definition inference** (`attributeProcessor.ts`) used the same `parseFloat` check, so such attributes were also *typed* as numbers.
- **Server parser** (`dive_utils/serializers/viame.py`) had the adjacent Python gotcha: `float()` accepts underscore digit separators (`float('20240624_120000') == 20240624120000.0`) and `inf`/`nan` spellings, corrupting underscore-separated filename stems the same way.

## Fix

- Desktop `_deduceType` uses `Number()`, which converts only fully numeric strings (with an explicit empty/whitespace guard, since `Number('') === 0`).
- Desktop `attributeProcessor` infers `number` only for fully numeric values.
- Server value parsing and type inference both require a strictly numeric string (regex full-match) before converting.

Real numbers (`12.5`, `-3`, `1e5`) and booleans still convert exactly as before. Regression test parses `(atr)` values with filename-like, underscore-separated, numeric, and boolean content and checks both values and inferred datatypes.

Note: values already truncated in previously saved annotations can't be recovered — re-importing the source CSV restores them.

### DIFF
--- a/client/platform/desktop/backend/native/attributeProcessor.ts
+++ b/client/platform/desktop/backend/native/attributeProcessor.ts
@@ -47,7 +47,11 @@ function processTrackAttributes(tracks: TrackData[]):
             lowCount = val;
           }
           values.push(key);
-          if (attributeType === 'number' && Number.isNaN(parseFloat(key))) {
+          // Full-string numeric check: parseFloat would accept the leading
+          // digits of filename-like values ('0123ABC456') and mistype the
+          // attribute as a number.
+          if (attributeType === 'number'
+            && (key.trim() === '' || Number.isNaN(Number(key)))) {
             attributeType = 'boolean';
           }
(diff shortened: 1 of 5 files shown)


BryonLewis (approved):
only added some minor additonal tests.

PR #1793, opened 2026-07-23
Speed up suppression-region computation during region edits

mattdawkins:
## Problem

Resizing a large suppression region (especially one carrying a polygon) on sequences with many detections lagged during the drag and briefly hung when releasing the corner handle.

Three compounding costs:
1. Every intermediate drag event bumps the pending-save counter, which re-ran the **dataset-wide** fully-suppressed computation (every keyframe of every track, each doing 16×16 point sampling against the region polygons) once per mouse move.
2. The per-frame suppressed set was computed **three times** per update — once each for the canvas (LayerManager), the per-frame type counts (FilterList), and the track list.
3. Every candidate detection on the frame paid the full point-sampling cost even when it obviously couldn't reach the coverage threshold.

## Fix

- **BBox prefilter**: `getSuppressedTrackIds` first computes a cheap bbox-overlap upper bound on the covered fraction; candidates that cannot possibly reach the threshold skip the point sampling entirely.
- **Per-revision memoization**: results are cached per store+frame keyed on an edit revision (the pending-save counter), passed by all three consumers — one computation per frame per edit instead of three.
- **Debounced dataset totals**: the fully-suppressed set behind the type-list totals recomputes on a 500 ms debounce instead of synchronously per save-counter bump, so corner drags stay responsive and the totals settle right after the drag ends. Suppression-type/camera changes still recompute immediately.

Unit tests cover the prefilter behavior (covered vs. barely-touching vs. distant candidates) and the memoization/revision semantics.

### DIFF
--- a/client/src/components/FilterList.vue
+++ b/client/src/components/FilterList.vue
@@ -179,7 +179,11 @@ export default defineComponent({
         });
         const hasRegionAt = (f: number) => regionRanges.some(([b, e]) => f >= b && f <= e);
         const suppressedAt = (f: number) => getSuppressedTrackIds(
-          store, f, suppType, suppThreshold, { revision: editRevision },
+          store,
+          f,
+          suppType,
+          suppThreshold,
+          { revision: editRevision },
         );
         store.annotationMap.forEach((annotation) => {
           const track = annotation as Track;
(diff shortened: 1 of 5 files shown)


BryonLewis (dismissed):
Updated to fix a few minor issues:
- pendingSavecount was declared multiple times in LayerManager.vue
- Have FilterList use each regions tracks span so interpolated frames also count
- a few other minor changes

PR #1794, opened 2026-07-23
Order recognized rig cameras STAR/CENTER/PORT on CLI import

mattdawkins:
## Summary

A multi-camera CLI import (`dive-desktop --camera NAME=path ...`) displayed cameras in the exact order the `--camera` flags were passed. The folder-import wizard, by contrast, already canonicalizes a recognized rig to **STAR / CENTER / PORT** via `sortSubfolderCameraNames` (`PREFERRED_SUBFOLDER_ORDER`). So the same rig could show up in a different camera order depending on which import path was used.

This applies that same canonical ordering in the CLI path: when the `--camera` set contains all of STAR/CENTER/PORT (case-insensitive), the returned `cameraOrder` is reordered to STAR/CENTER/PORT. Any other camera set (e.g. `left`/`right`, or a partial rig) keeps the flag order verbatim, so existing behavior is unchanged for everything that isn't a recognized rig.

## Changes
- `cliImport.ts`: import `PREFERRED_SUBFOLDER_ORDER` / `sortSubfolderCameraNames` from the existing `multicamSubfolderLayout` module (backend-safe — it only pulls `fileVideoTypes` from `dive-common/constants`, already imported here) and apply the canonical order to `cameraOrder`; `defaultDisplay` falls back to the first canonical camera.
- `cliImport.spec.ts`: 4 new cases — rig reordering is order-independent, `defaultDisplay` fallback for a rig, and non-rig sets stay in flag order.

## Test plan
- `vitest run platform/desktop/backend/cliImport.spec.ts` → 26 passed
- `eslint` on both files → clean

### DIFF
--- a/client/platform/desktop/backend/cliImport.spec.ts
+++ b/client/platform/desktop/backend/cliImport.spec.ts
@@ -111,7 +111,7 @@ describe('parseCliArgs multi-camera', () => {
     expect(args?.metadataPath).toEqual('/data/flight_log.csv');
   });
 
-  it('defaults defaultDisplay to left, else the first camera', () => {
+  it('defaults defaultDisplay via wizard heuristics (left / center / EO / first)', () => {
     expect(parseCliArgs([
       'app', '--camera', 'right=/d/r', '--camera', 'left=/d/l',
     ])?.defaultDisplay).toEqual('left');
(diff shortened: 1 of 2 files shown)


mattdawkins:
Maybe this should be abandoned if it's making it so that it ignores the CLI passed in ordering. Not sure what it was thinking here.

BryonLewis:
I think it's fine to have this to just make it eaiser.  I only added the functionality for our other orderings we use EO/UV/IR and Left/Right stereo.

BryonLewis (approved):
I think this is fine, it just makes it consistent if the user isn't.

I've added additional ordering for EO/UV/IR and left/right stereoscopic as well.

PR #1796, opened 2026-07-23
Shared color scope: reuse type/track colors across datasets

mattdawkins:
## Summary

Type/track/group color and style overrides were stored only in the metadata of the dataset they were set on (`customTypeStyling` / `customGroupStyling`). The same class could therefore render a different color in every sequence, and a color a user picked never followed them to the next dataset.

This adds a **Type color scope** user setting (User Settings dialog) with two modes:

- **Shared across all data** (default) — one set of color/style overrides is reused across every dataset.
  - **Desktop** (no user accounts): stored once per data directory in `global_style_settings.json`, shared across all sequences.
  - **Web**: scoped to the current user/browser via `localStorage`, consistent with how the web client already persists user preferences (`clientSettings`).
  - On dataset open, the shared styles **overlay** the dataset's own styling (shared wins on conflicts); any style the user edits is **mirrored back** to the shared store; and colors that ride in with a dataset **seed** the shared store (without overwriting existing shared choices) so imported colors propagate to future sequences.
- **Per dataset** — the original behavior; colors are saved only with the dataset they were set on.

The default is shared, so out of the box a color set once is reused everywhere. Choosing *Per dataset* reproduces today's exact behavior (the shared paths become no-ops).

## Changes

- `dive-common/store/settings.ts` — new `typeSettings.colorScope: 'shared' | 'dataset'` (default `'shared'`; hydrates onto existing stored settings).
- `dive-common/components/UserSettingsDialog.vue` — a "Type color scope" select.
- `src/StyleManager.ts` — optional `onStyleEdit` hook, fired on `updateTypeStyle` (genuine edits only, not on dataset load).
- `dive-common/components/Viewer.vue` — loads/overlays/seeds the shared store on dataset open and persists edits (debounced).
- `dive-common/apispec.ts` — new `GlobalStyleSettings` type and optional `loadGlobalStyleSettings` / `saveGlobalStyleSettings` Api methods.
- **Desktop**: `backend/native/globalStyles.ts` (file read/write under `settings.dataPath`, degrades to empty on missing/corrupt), IPC handlers, and frontend api wrappers.
- **Web**: `localStorage`-backed implementation in `web-girder/api/dataset.service.ts`, wired into `provideApi`.

## Test plan

- `vitest run platform/desktop/backend/native/globalStyles.spec.ts` → 6 passed (round-trip, missing dir, corrupt file, normalization).
- `vitest run src/StyleManager.spec.ts` → passed (optional hook is additive).
- `eslint` on all changed files → clean.
- `electron-vite build` (desktop backend + renderer) and `vite build` (web) → both succeed.

## Notes

- Web persistence is per-user-per-browser (localStorage), not cross-device — there is no per-user server settings blob today, and adding one would need a new Girder route. This keeps the change client-only; a server-backed per-user store can be a follow-up.
- Group styling is included alongside type styling for completeness.


BryonLewis (approved):
I've made some changes:
- Made it so that when unused types that aren't part of the dataset don't show up in the list from other datasets.  I.E types from another dataset only show up in unused if they have been explicitly added before to the dataset
- Created a way to edit the saved styles and remove/add styles to a user's add styles
- Linked up what dataset a style came from and created a filtering tool to make it easier to see the styles and remove batches of them
- updated documentation to reference this type styling system

PR #1807, opened 2026-07-27
Unify metadata attachments across web and desktop

PaulHax:
Second of a four-PR stack, on top of #1806. 

### Summary

One optional metadata attachment per dataset scope, consistent across web and desktop.

Single-camera datasets get one shared attachment; multicamera datasets get one shared parent
attachment plus one camera-local attachment per camera, camera-local taking precedence. Image
sequences and videos are supported from the start. JSON, TXT, and CSV stay opaque pipeline inputs
here — row parsing is the next PR.

Carried through: web and desktop creation-time selection; headless and RPC ingestion; single-camera
and multicamera imports; clone-root lookup; export/import round trips; pipeline discovery and opt-in
injection; and a normalized `GET /dive_dataset/:id/frame_metadata_sources`.

Reserved `frame-metadata.*` and `frame_metadata.*` names are recognized identically by Python and
TypeScript, pinned by a shared conformance fixture, and never sent through annotation parsing.

The read path — the source endpoint, the web and desktop loaders, and `loadFrameMetadata` on the
`Api` interface — lands here with no production consumer, because it *is* the attachment contract
and the next PR is its first caller. Review it against `test_frame_metadata_sources.py`, not a UI.

Two changes outside the attachment story ride along:

- **One annotation CSV per sweep.** `process_items` imported every CSV, last write winning. It now
  imports the oldest and names the rest in a warning, so the headless sweep enforces the rule #1806
  enforces interactively. The sweep is already rewritten here to pre-scan for reserved names.
- **Metadata-item download path.** `_inject_dataset_metadata_file` reconstructed `<dir>/<item name>`,
  wrong when girder_client sanitizes or nests the name — in the nested case `exists()` passes and a
  *directory* is bound into the KWIVER setting. It now locates what actually landed.

The write-only `metadataFile` item marker and its helper are dropped; nothing ever read it, and the
folder's `metadataFileItemId` is the only attachment locator. Items in existing datasets keep the
stale key harmlessly.

### Reviewer focus

- Attachment identity is stored once per scope and stays inside the owning dataset/clone root.
- Camera-local attachments override shared ones without hiding shared data from other cameras.
- Attachment bytes stay opaque to the server.
- Pipeline injection happens only for pipelines declaring a metadata-file key.
- Clone/export/import preserve original names and create no annotation tracks.
- Replacing an attachment deletes the superseded file only when DIVE stored it. A reserved-name file
  is the user's own content, uploaded with the media and merely discovered — and `process_items`
  records what it discovers in the same marker, so ownership is decided by the reserved-name
  predicate, not the marker's presence.
- A camera's attachment is the optional third argument of #1806's camera package. Unlike the
  annotation pick it is removed from the package rather than swapped into it — it is uploaded and
  declared separately once the camera dataset exists — and a folder file it displaces is still
  reported.
- Metadata selections use opaque per-selection ids, so same-named attachments picked from different
  directories for different cameras do not collide.

### Manual test

Uses `okeanos-frame-metadata-upload`, `okeanos-video-explicit-frame`, and `okeanos-video-opaque-json`
from `okeanos-frame-metadata-fixtures.zip`, attached to #1806 — CC0 NOAA Okeanos Explorer media with
synthetic metadata and annotations. Select all files in one directory per test.

1. Select all files; confirm the reserved name takes the Metadata File slot, not the annotation slot.
   In `okeanos-frame-metadata-upload` both slots fill at once: `annotations.csv` is the
   annotation pick and `frame-metadata.csv` the attachment, from a single selection of two CSVs.
2. Create and open: expected media, and the only tracks are the two from the VIAME CSV — the
   attachment contributes none.
3. `GET /api/v1/dive_dataset/<folder-id>/frame_metadata_sources` — one `shared` object with an
   `itemId`, the original name, and an empty `cameras` object.
4. Clone the image dataset; the same request against the clone still resolves through the clone root.
5. For the JSON case, the endpoint reports `frame_metadata.json` — stored for pipelines, not
   interpreted, and the underscore spelling is discovered exactly like the hyphen one.

Desktop smoke test: import a media directory, choose its sidecar in Metadata File,
export the archive, reopen, and confirm the attachment name survives.


BryonLewis:
Assetstore video metadata sidecars.  This is to account for assetStore Importing with metadaata, for image sequences it is fine to use the existing frame_metadata.csv or variations for videos because they are in a flat list a different choice had to me bade.  Thi is matching the video file name with a variable of metadata appended to it.

Import: Matches {videoStem}[_-]metadata.{csv|json|txt}, moves it into the video folder as frame_metadata.{ext}, and links it as the dataset metadata attachment. On reimport, any existing reserved-name frame metadata in that folder is replaced and the folder markers are retargeted.

The sample data generation script for assetStore import was updated to include generating fake metadataframe data.

Sample data: generateSampleData now emits those paired sidecars for videos, plus reserved frame[_-]metadata.{ext} files inside image-sequence folders, so MinIO/assetstore import can exercise the path end-to-end.

BryonLewis (dismissed):
I think this looks good now.

I've added the support for the assetStore for the metadata format.  This works by default for images but for video the files need to be matched and moved around.  There is an update that handles that as well as an update to the sample generation tool that makes it so the system will generate samples for testing purposes.

Did some small fixes to some minor race conditionsa and added a visualization indicator on the metadata item for debugging purposes.

Also updated documentation to add information about assetStoreImporting and how it will work

PR #1814, opened 2026-07-28
Let a pipe declare which config keys receive the calibration file

mattdawkins:
Stereo calibration was bound to two hardcoded keys, `measurer:calibration_file` and `calibration_reader:file`. Pipes whose calibration consumer is neither of those had no way to receive it.

Routing it through a `$CONFIG{global:...}` indirection does not work: kwiver expands macros at parse time with backward references only (`bakery_base.cxx:189-211`), while `-s` blocks are appended last (`pipeline_builder.cxx:131-173`), so the override never reaches the process.

- New `# Calibration Keys: <k> [k...]` header naming the consuming process keys directly, parsed on both server (`pipeline_discovery.py`) and desktop (`common.ts`)
- `append_stereo_calibration_kwiver_settings` emits one `-s` per declared key; unset keeps the two conventional keys, so existing pipes are unaffected
- Header added to the description stop-condition on both parsers so it doesn't bleed into descriptions

Paired with VIAME/VIAME#dev/measurement-calibration-passdown, which declares the header on the measurement and ifremer stereo pipes.

Tests added for declared-keys emission, default fallback, and header parsing.

### DIFF
--- a/client/dive-common/apispec.ts
+++ b/client/dive-common/apispec.ts
@@ -70,6 +70,14 @@ interface PipeMetadata {
    * it gets the first camera's list. Parsed from a `# Image List Keys:` header.
    */
   imageListKeys?: string[];
+  /**
+   * KWIVER config keys (e.g. "depth_map:computer:ocv_stereo_disparity:calibration_file")
+   * that the dataset's stereo calibration file is bound to at run time. Parsed from a
+   * `# Calibration Keys: <k> [k...]` header. Pipes whose calibration consumer is not the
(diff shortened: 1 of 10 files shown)


BryonLewis:
Create a new issue to make sure that measurement pipelines that creates images on desktop will create a new dataset like the filter/transcode pipelines do.  Currently it has to be either a filter or transcode pipeline, but this would be a special case for stereoscopic.

BryonLewis (dismissed):
Minor Documentation and unifiying the Static values for calibration keys.

PR #1825, opened 2026-08-05
Desktop: resume interrupted training runs

mattdawkins:
- Desktop training jobs that were interrupted (crash, cancel, closing DIVE) or exited early can now be continued if their intermediate files are still in `DIVE_Jobs`
- The Training page gains an "Interrupted training runs" section with Resume and Discard actions; resume re-launches `viame train` in the original working directory with `--continue`, discard deletes the run's working directory
- Training jobs now record their final exit status into `dive_job_manifest.json` so interrupted runs are detected across app restarts; legacy successful runs are excluded via their emptied `category_models` folder
- Requires a VIAME build whose train tool supports `--continue`; on older builds the resumed command fails with an unknown-option error in the job log

### DIFF
--- a/client/platform/desktop/backend/ipcService.ts
+++ b/client/platform/desktop/backend/ipcService.ts
@@ -331,6 +331,10 @@ export default function register() {
     };
     return currentPlatform.train(settings.get(), args, updater);
   });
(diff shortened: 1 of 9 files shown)


BryonLewis (approved):
Small change to handle the difference when cancelling a job vs it ending for another reason.

PR #1826, opened 2026-08-05
Frame label mode: rapid whole-frame event labeling

mattdawkins:
- New sidebar panel defines up to 9 frame labels; while the mode is enabled, keys `1`-`9` label all frames from the current frame forward with the corresponding label until the next labeled event (or end of video), and `0` ends the open segment — designed for one-pass labeling during playback
- Pressing a key at an existing segment's start relabels it in place; pressing inside a segment of a different label truncates it and starts the new one, so events can be inserted or corrected after the first pass
- Segments are ordinary full-frame tracks (begin/end keyframes, interpolated), so the type list, styling, confidence filters, event timeline, CSV export, and frame classifier training all work unchanged
- Label list persists in client settings; panel shows the label active at the current frame
- Core logic is a standalone composable (`useFrameLabelMode`) with unit tests

### DIFF
--- a/client/dive-common/apispec.ts
+++ b/client/dive-common/apispec.ts
@@ -245,9 +245,11 @@ interface DatasetConfigMutable {
   cameraTransformTypes?: CameraTransformTypes;
   /** Producer provenance of the camera registration (see RegistrationSource). */
   cameraRegistrationSource?: RegistrationSource | null;
+  /** Frame label mode label set defined by the dataset, in hotkey order. */
+  frameLabels?: string[];
   error?: string;
 }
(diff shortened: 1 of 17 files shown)


BryonLewis (commented):
There are a few things I need to get down into text about this functionality:

- Adding a new item to Left sidebar is fine but the label types push down the track list and don't properly resize the scrolling area
- "Frame label mode" isn't clear that it is used for full frame labelling and that it will generate full frame tracks from now until the next change time
- We have a type list already this is adding a secondary type list and now we have types displayed in two areas
- right now it only works in the standard sidebar configuration the layout with all items at the bottom doesn't work.

Alternative Proposal:
- In the type filter settings there is an option for full frame labelling
- This brings up a dialog where users can select out of the existing type list or add new types to set keyboard shortcuts from 0-9
- The type list now is reordered with types 1-9 and an indication of the shortcut key for each type
- once frame labelling has been configured in the setting there is an indicator on the top of the type list that allows toggling frame labelling on/off
- During frame labelling mode, the edit types goes away (because keyboard shortcuts for them don't work) and the info area where we indicate what can be done is swapped to give basic instructions about the frame labelling system.
- Update track text info that any full frame tracks place the label inside the upper right corner instead of outside for easier viewability
- the frame labelling settings should be stored with the types in the config for the dataset so it can be portable

PR #1828, opened 2026-08-06
Keep the pipeline menu's warning and gear icons in fixed columns

mattdawkins:
Long pipeline names pushed the gear (and warning) icon out of the row box instead of the name giving way, so icons landed in different columns per row.

- Name is its own flex child that ellipsizes instead of shoving the icons.
- Warning and gear each get a fixed slot, reserved for every row in a category when any pipeline in that category needs it — so the warning always sits directly left of the gear, and categories with neither icon reserve nothing.
- `scrollbar-gutter: stable` on the submenu list: without it the 10px scrollbar is carved out of the width the menu already sized itself to, and the widest rows overflow by exactly that.

Verified in headless Chrome against the real Vuetify DOM: warnings previously split across two columns (x=619.6 / 675.6), now all at 693.6 with gears at 715.6, holding even when the menu is narrow enough to truncate names.

### DIFF
--- a/client/dive-common/components/RunPipelineMenu.vue
+++ b/client/dive-common/components/RunPipelineMenu.vue
@@ -186,6 +186,19 @@ export default defineComponent({
       props.excludePipelineTerms,
     ));
 
+    /* Icon slots are reserved per category so the columns line up across rows. */
+    function categoryPipes(pipeType: string) {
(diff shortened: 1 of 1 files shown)


mattdawkins:
Goal is to prevent this

example2

BryonLewis (approved):

PR #1830, opened 2026-08-07
Allow deleting corrupt or incomplete datasets

mattdawkins:
A dataset whose import died before `dataset.json` was written can never be deleted from the Recents list — `delete-dataset` fails with `missing metadata json file <path>`.

- `deleteDataset()` no longer runs `getValidatedProjectDir()`; it removes the project directory whatever state it's in, loads the dataset json only opportunistically for the DIVE_Jobs_Output cleanup, returns true when the directory is already gone, and rejects ids resolving outside DIVE_Projects.
- `getValidatedProjectDir()` checks the base path before `ensureDirSync(aux)`, so failed lookups stop creating stub `<id>/auxiliary/` dirs — which is what made the ghost entries reappear after manual cleanup.

Tests updated for the new behavior, 3 added.


BryonLewis:
I think I just just autosync on launch ofthe electorn version, this would make sure that broken things are removed and it just suggest to run autosync if you try to open/delete a non-existing item

mattdawkins:
I'm not sure 'sync to delete' is readily known to users (or me)

BryonLewis (approved):

PR #1832, opened 2026-08-07
Regroup desktop import buttons

mattdawkins:
Desktop home screen import buttons only; web-girder upload is unchanged.

- `Open Tiled GeoTIFF / TIFF` moves from a top-level button into the `Open Image Sequence` dropdown, keeping its guidance as a tooltip on the menu item
- `Bulk Import` gains an arrow dropdown with `Single Camera` / `Multi-Camera`
- `MultiCam Batch` moves out of the Image Sequence dropdown into that new `Multi-Camera` entry


BryonLewis (approved):
Minor changes and we may want a future PR that reconfigures the Web version of the input arguments

PR #1836, opened 2026-08-10
Make right-click recenter opt-in

mattdawkins:
Right-click in the annotator recenters the view on the clicked point. That has
been on unconditionally since #1747 (`AlignedImageLayer`), including in single
camera, non-aligned datasets. It fights right-click's primary role as the
context/edit action: right-clicking a detection to edit it also jumps the view.

Gate it on a User Settings toggle, **defaulting off**. The behavior is
unchanged for anyone who turns it on.

- `navigationSettings.rightClickRecenter` in `clientSettings` (persisted, default `false`)
- `useLayerManagerAlignedView` ANDs it with the existing "not while editing geometry" guard
- Toggle in the User Settings dialog

Tests cover all three states: no recenter by default, recenter when enabled,
and left-click ignored when enabled.

### DIFF
--- a/client/dive-common/components/UserSettingsDialog.vue
+++ b/client/dive-common/components/UserSettingsDialog.vue
@@ -39,14 +39,6 @@ export default defineComponent({
           persistent-hint
         />
         <v-switch
-          v-model="clientSettings.navigationSettings.rightClickRecenter"
-          color="primary"
-          class="my-0 mt-3"
-          label="Right-click recenters the view"
-          hint="Right-clicking the image centers the view on that point. Off by default; right-click still opens edit actions either way."
-          persistent-hint
-        />
-        <v-switch
           v-model="clientSettings.autoSaveSettings.enabled"
           color="primary"
           class="my-0 mt-3"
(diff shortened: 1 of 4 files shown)


BryonLewis:
This is an issue indicated by that PR, but it should probablly default to the track locking indicator and not some new setting.  The current lock icon is used to center when selecting tracks and it should probably use that.

mattdawkins:
Good call — done in 14cb2b82. Dropped the new setting entirely and gated the recenter on `annotatorPreferences.lockedCamera.enabled`, so the lock icon ("center camera on selected track") is now the single control for "the camera may move itself." Still off by default, and still suppressed while creating/editing geometry where right-click means "remove last point."

One thing worth a sanity check from you, since it's a behavior change rather than just a relocation: the two are now coupled in both directions. Enabling the lock to get track-following also turns right-click recenter on, and anyone who wants right-click-to-center has to accept the camera jumping on every track selection. That seems like the intent — they're both "auto-centering" — but if you'd rather they stay independent, the alternative is a second entry in the lock icon's popover (next to Transition / Multiply Bounds) rather than a User Settings switch. Say the word and I'll move it there.

Tests cover the three states: no recenter with the lock off, recenter with it on, left-click ignored either way. Worth noting the aligned-view spec's viewer mock had no `center` at all before this, which is how the behavior shipped on-by-default in #1747 without anything catching it.

BryonLewis (approved):
one small change to suppress multiple component warnings in the testing for useLayerManagerAlignedView

PR #1843, opened 2026-08-12
Add a TypeScript type check gate for the client

PaulHax:
# Add a TypeScript type check gate for the client

## Why

The client had no TypeScript type check anywhere. CI ran `lint`, `lint:templates`, `test`, and two
vite builds, and vite/esbuild strips types without checking them. Vue SFC templates were therefore
completely unchecked — including the prop/method seams where a changed type silently becomes a
runtime `TypeError` on click.

## What

- Bump `typescript` to `~5.6.3` and add `vue-tsc@~2.1.10`. The old `~4.3.5` pin (2021) cannot parse
  modern `.d.ts` files in `node_modules`, failing with hundreds of syntax errors before it reaches
  any project code; `skipLibCheck` does not help, since those are parse errors.
- Add `client/tsconfig.typecheck.json`, extending `tsconfig.json` with `noEmit`, `skipLibCheck`,
  `isolatedModules`, `vueCompilerOptions: { target: 2.7 }`, and `types: ["vitest/globals", "node"]`.
- Add `npm run typecheck` and a `Typecheck` step on the `web` matrix leg of CI.
- Fix every error the gate reports, so it starts green.

The gate covers the whole client — `src/`, `dive-common/`, and `platform/`. No Vue 3 migration is
involved; `vue-tsc` checks these Vue 2.7 SFCs directly.

`isolatedModules` is on because a re-exported type in a value `export {}` block type-checks fine but
breaks the rollup build. That happened during this work, so the gate now catches it instead of CI.
Four modules had their export blocks split into value and `export type` blocks as a result.

## Bugs this surfaced

Most of the ~400 errors were honest-typing chores, but the check exposed real defects:

- `ViewerLoader.vue` (desktop) called `npath.join`/`npath.isAbsolute` in `getImagePathForFrame`,
  but `npath` does not exist in that module — the file's own `joinPath`/`isAbsolutePath` helpers are
  used everywhere else. That path threw `ReferenceError` whenever it was reached.
- `RunPipelineMenu.vue` passed the user's configured pipeline parameters into the
  `outputDatasetNameById` argument instead of `kwiverParams`, so parameters entered in the Configure
  dialog were silently dropped on every parameterized pipeline run. Now threaded through
  `kwiverParams`, the channel both the desktop and web-girder backends actually read.
- `AnnotationTags.vue` had `v-for="(set, index) in set"` — iterating an identifier that does not
  exist. Only `sets` is in scope.
- `AdminJobs.vue` called `formatStatus(item.status)` against a two-parameter function, leaving
  `updated` undefined, so `moment(undefined)` rendered the current time as every job's update time.
- `GroupItem.vue` passed `group.getType()` — a `[string, number]` confidence pair — to
  `TypePicker`'s `value` prop, which takes the type name `string`. `TrackDetailsPanel.vue` had the
  same defect.
- `GroupList.vue` passed a narrow `SortedAnnotation` copy as `GroupItem`'s `group` prop, but
  `GroupItem` calls `setType` and reads `memberIds`, neither of which exists on that copy. This is
  the same projection-vs-full-object class of bug the gate exists to catch.
- `useMediaController.ts` built its per-camera controllers without `resizing`, `alignedGapSlots`, or
  `seekCameraFrame`, all of which `MediaController` declares.
- `useMulticamEditRouting.ts` imported `'../../trackStore'`, which does not exist. The import
  resolved only because nothing checked it.
- `multiCamImport.ts` accepted `type: 'large-image'` and produced a camera record no downstream
  branch handles, collecting no media. It now rejects that combination explicitly.
- `AttributeNumberValueColors.vue` compared `item.key.toString() === val` with a numeric `val`, so
  the "values need to be unique" rule could never fire.

## Notes for review

- `SideBarTrackItemView.vue`'s `track` prop was `{ type: Object }`, i.e. `any`, so nothing in that
  template was checked. It is now `PropType<Track>`, which is what makes the gate cover the
  component where `TypeError: _vm.track.canSplit is not a function` originally shipped.
- Several large clusters came from one loose declaration each, fixed once at the source rather than
  at each use site: `DesktopBridge.invoke`/`.on` in `desktop-preload.d.ts` are now generic; the
  desktop `api.ts` routes ~55 IPC calls through one typed `invoke<T>` wrapper; the multi-camera
  import `ctx` prop declared `required: true` without `as const`, which defeated Vue's required-prop
  narrowing across eight components; `BottomPanel.vue`'s props were bare `Object`/`Function`.
- Validation rules in several components moved from inline template arrays into the script.
  `vue-tsc` accepts type annotations inside template expressions, but `vue-template-compiler` does
  not — annotating them in place type-checks and then throws at runtime.
- `@types/archiver` replaces an untyped import rather than a hand-written declaration.
- No error was silenced with `any`, `@ts-ignore`, `@ts-expect-error`, or a non-null assertion. One
  double cast remains, `girderRest as unknown as AxiosInstance` in `multicamFileRegistry.ts`: Girder's
  `RestClient` genuinely lacks members that `UploadManager`'s constructor requires, and that
  constructor is in `node_modules`.

## Verification

- `npm run typecheck` passes, and fails as intended when a bogus method call is added to a typed
  object in a template — verified against `track.canSplit` in `SideBarTrackItemView.vue`, the exact
  expression that produced the original crash.
- Client gate passes: unit tests, lint, lint:templates, build:web, build:electron.
- Server lint and mypy fail on this branch exactly as they do on `main`; no server file is touched.

`@typescript-eslint` 6.x officially supports TypeScript below 5.4 and is now paired with 5.6.
`npm run lint` and `npm run lint:templates` both pass clean, with no version warning, because the
config does not use type-aware linting. Upgrading the eslint stack is left as a separate decision.

### DIFF
--- a/.github/workflows/blank.yml
+++ b/.github/workflows/blank.yml
@@ -36,6 +36,10 @@ jobs:
       if: ${{ matrix.target == 'web' }}
       run: npm run lint:templates
     -
+      name: Typecheck
+      if: ${{ matrix.target == 'web' }}
+      run: npm run typecheck
+    -
       name: Run tests
       if: ${{ matrix.target == 'web' }}
       run: npm test -- --coverage
--- a/client/dive-common/apispec.ts
+++ b/client/dive-common/apispec.ts
@@ -8,7 +8,7 @@ import { Attribute } from 'vue-media-annotator/use/AttributeTypes';
 import { CustomStyle } from 'vue-media-annotator/StyleManager';
 import { AttributeTrackFilter } from 'vue-media-annotator/AttributeTrackFilterControls';
 import { ImageEnhancements } from 'vue-media-annotator/use/useImageEnhancements';
-import {
+import type {
   CameraHomographies, CameraCorrespondences, CameraTransformTypes, RegistrationSource,
 } from 'vue-media-annotator/alignedView/CameraRegistrationStore';
 import type { PercentileStretch } from 'vue-media-annotator/use/useImageEnhancements';
@@ -659,13 +659,16 @@ export {
 };
 
 export {
+  DatasetConfigMutableKeys,
+  MulticamSharedMutableKeys,
+};
+
+export type {
   AnnotationSchema,
   Api,
   DatasetConfig,
   DatasetConfigMutable,
-  DatasetConfigMutableKeys,
   DatasetInfoFields,
-  MulticamSharedMutableKeys,
   GlobalStyleSettings,
   DatasetType,
   DiveParam,
@@ -694,4 +697,4 @@ export {
   MediaImportResponse,
 };
 
-export type { PercentileStretch };
+export type { PercentileStretch, CameraCorrespondences };
--- a/client/dive-common/components/Attributes/AttributeEditor.vue
+++ b/client/dive-common/components/Attributes/AttributeEditor.vue
@@ -138,7 +138,7 @@ export default defineComponent({
       (v: string) => !isReservedAttributeName(v, baseSettings.belongs)
         || `Reserved name. ${RESERVED_ATTRIBUTES[baseSettings.belongs].join(', ')} are not allowed.`,
     ]);
-    const isValidNumberString = (v: string) => !Number.isNaN(parseFloat(v));
+    const isValidNumberString = (v: string) => !Number.isNaN(parseFloat(v)) || 'Number is required';
     const rangeMinRules = computed(() => [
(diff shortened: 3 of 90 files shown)


BryonLewis (approved):
small adjustment for attribute non numeric values.

Any other cleanups and updates can be done in subsquent PRs.  I want to get this merged so I can take a look at the ONNX PR.

PR #1845, opened 2026-08-14
Preserve warnings from every Desktop import file

PaulHax:
# Preserve warnings from every Desktop import file

### Failure case

A Desktop import can contain more than one primary annotation file. Before this change, each file
replaced the warnings from the preceding file. The import report showed warnings only from the last
file.

Supported annotation data was still imported. However, the missing warnings could hide skipped or
changed data from earlier files.

### Steps to reproduce

Use `dive-classification-test-data/coco/rle-mask-warning-aggregation/` from the test-data ZIP built
by [PaulHax/dive-devkit](https://github.com/PaulHax/dive-devkit).

1. Use `media/image-sequence/` as the image sequence.
2. Call the exported Desktop `ingestDataFiles` function with
   `rle-warning-fish.coco.json` and `rle-warning-shark.coco.json` in the primary input list.
3. Inspect the warnings after the import completes.
4. Expect one warning for each file. Before this fix, the result contains only the last file's
   warning list.

### Changes

- Append warnings from each successful primary file in input order.
- Keep duplicate warnings because each warning can describe a different input file.
- Keep primary-file warnings before multicamera warnings.
- Propagate a later parse error through the exported ingestion entry point.

### Scope

This PR changes `common.ts` and `common.spec.ts`. It does not change the warning text or the import
dialog. The tests use the exported Desktop ingestion entry point. They do not use a live Electron
import dialog.

### Hierarchy prerequisite

The hierarchy feature does not cause this bug, but its KWCOCO imports add warnings that the existing
bug could discard.

### Stack

This is PR 2 of 9. [Previous: TypeScript spec lint](https://github.com/Kitware/dive/pull/1844).
[Next: isolate soft-clone metadata](https://github.com/Kitware/dive/pull/1846).

1. [Allow devDependency imports in TypeScript spec files](https://github.com/Kitware/dive/pull/1844)
2. **Current — Preserve warnings from every Desktop import file**
3. [Copy source metadata when creating a single-camera soft clone](https://github.com/Kitware/dive/pull/1846)
4. [Add hierarchical track classification](https://github.com/Kitware/dive/pull/1847)
5. [Centralize hierarchical classification changes](https://github.com/Kitware/dive/pull/1848)
6. [Replace mutable merged tracks with read-only projections](https://github.com/Kitware/dive/pull/1849)
7. [Make track lifecycle operations classification-safe](https://github.com/Kitware/dive/pull/1850)
[...]

PR #1846, opened 2026-08-14
Copy source metadata when creating a single-camera soft clone

PaulHax:
# Copy source metadata when DIVE creates a single-camera soft clone

### Failure case

A single-camera soft clone used the source metadata object directly. The source and clone then
shared one mutable object. Clone markers and later nested changes could modify the source metadata.

The multicamera clone path already makes an independent copy.

### Steps to reproduce

1. Create a source folder object that has nested metadata.
2. Call the public `createSoftClone` server entry point.
3. Change a nested value in the source metadata object and inspect the clone metadata object.
4. Change a nested value in the clone metadata object and inspect the source metadata object.
5. Expect each object to keep its own value. Before this fix, each object receives the change from
   the other object.

This object-isolation defect has no supported user-interface reproduction. The automated
`createSoftClone` test is the authoritative reproduction.

### Changes

- Make a deep copy of source metadata before DIVE adds clone markers and default values.
- Test the public `createSoftClone` path.
- Verify that nested changes cannot pass from the source to the clone.
- Verify that nested changes cannot pass from the clone to the source.

### Scope

This PR changes `crud_dataset.py` and `test_multicam_export_clone.py`. It does not test a live
Girder clone, the Mongo save boundary, or a viewer reload.

### Hierarchy prerequisite

The hierarchy is stored in dataset metadata, so a soft clone must copy that metadata without
changing or sharing the source object.

### Stack

This is PR 3 of 9. [Previous: preserve Desktop import warnings](https://github.com/Kitware/dive/pull/1845).
[Next: add hierarchical track classification](https://github.com/Kitware/dive/pull/1847).

3. **Current — Copy source metadata when creating a single-camera soft clone**

This PR is stacked on `desktop-warning-preservation`.

Commit: `4ad85091 Copy source metadata when creating a single-camera soft clone`.

### DIFF
--- a/server/dive_server/crud_dataset.py
+++ b/server/dive_server/crud_dataset.py
@@ -180,7 +180,7 @@ def _create_single_camera_soft_clone(
         reuseExisting=False,
         creator=owner,
     )
-    cloned_folder['meta'] = source_folder['meta']
+    cloned_folder['meta'] = copy.deepcopy(source_folder['meta'])
     media_source_folder = crud.getCloneRoot(owner, source_folder)
     cloned_folder[constants.ForeignMediaIdMarker] = str(media_source_folder['_id'])
     cloned_folder['meta'][constants.PublishedMarker] = False
--- a/server/tests/test_multicam_export_clone.py
+++ b/server/tests/test_multicam_export_clone.py
@@ -58,6 +58,46 @@ def _descriptor(name: str):
     return {'itemId': f'{name}-id', 'name': name}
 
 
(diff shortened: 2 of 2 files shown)


BryonLewis (dismissed):
I can understand why this is needed.  I think before the metadata editing we assumed that the metadata was attached to the video and that only annotations would be changed from clones.  That is no longer true given the other things that are being stored in the metadata now.

BryonLewis (dismissed):

BryonLewis (approved):

PR #1848, opened 2026-08-14
Centralize hierarchical classification changes

PaulHax:
# Centralize hierarchical classification changes

PR 1 lets several user actions change classifications through existing model methods. In a
multicamera dataset, those methods can change only one camera or a temporary merged track.

This PR adds commands for one logical track. A command calculates one canonical confidence vector.
It then installs independent copies on all linked camera tracks.

### Changes

- Add pure operations for reassignment, acceptance, confidence changes, and pair removal.
- Add `CameraStore` commands for all linked camera tracks.
- Use the first configured camera as the canonical source.
- Notify each changed camera one time.
- Route the sidebar, bottom bar, Track Details, confidence editor, and rename actions through the
  new commands.
- Preserve the commands after save and reload.

### Behavior

- Reassignment replaces one hierarchy lineage and keeps unrelated alternatives.
- **Accept as correct** sets the selected pair to `1.0`. It keeps related pairs and removes
  unrelated pairs.
- A normal confidence edit to `1.0` does not accept the pair as correct.
- Pair removal removes only the selected pair.
- PR 4 handles deletion of a logical track when its last pair is removed.

### Manual tests

Test data: download and extract
[classification.zip](https://github.com/user-attachments/files/31053807/classification.zip). Paths
below are relative to `dive-classification-test-data/`.

#### Single-camera classification commands

1. Create the dataset from
   `dive-classification-test-data/classification/single-camera-linked-types/`.
2. In separate save-and-reload cycles, reassign a leaf, accept a parent, change a confidence to
   `1.0` without accepting it, and remove one pair.

Verify: Reassignment changes one lineage, acceptance removes only unrelated pairs, a normal `1.0`
edit keeps unrelated pairs, and pair removal removes only the selected pair.

#### Multicamera classification synchronization

1. Create the dataset from
   `dive-classification-test-data/classification/multicamera-linked-types/`.
2. Reassign, accept, change confidence, and remove a pair; save and switch cameras after each
   action.

Verify: Both camera tracks contain equal confidence vectors after each action.

### Stack

This is PR 5 of 9. [Previous: add hierarchical track classification](https://github.com/Kitware/dive/pull/1847).
[Next: add read-only track projections](https://github.com/Kitware/dive/pull/1849).

1. [Allow devDependency imports in TypeScript spec files](https://github.com/Kitware/dive/pull/1844)
2. [Preserve warnings from every Desktop import file](https://github.com/Kitware/dive/pull/1845)
[...]

PR #1847, opened 2026-08-14
Add hierarchical track classification

PaulHax:
# Add hierarchical track classification

This PR adds an optional type hierarchy to each dataset. For example, a dataset can define `fish`
as the parent of `shark`. It can then define `shark` as the parent of `great white shark`.

DIVE keeps the existing `confidencePairs` on each track. The hierarchy is a display and editing
rule. It does not add hierarchy data to tracks or change the annotation schema.

### Selection

The viewer selects the deepest checked type that meets its confidence threshold. If `great white
shark` does not meet its threshold, the viewer can select `shark`.

- The selected pair controls the displayed type, confidence, and color.
- Single-camera and multicamera views use the same selection rule.
- Each camera resolves its own camera track.
- Lists, counts, and charts use the merged logical-track view.
- A dataset without a hierarchy keeps the existing flat selection behavior.

### Configuration

DIVE stores the hierarchy as `typeHierarchy` in DIVE Configuration JSON. The value maps each child
type to its direct parent.

- A hierarchy is a forest in which each child has one parent.
- DIVE rejects empty names, self-edges, cycles, and conflicting parents.
- An invalid import does not change the stored configuration.
- An overwrite import replaces a nonempty hierarchy.
- `null` and `{}` delete the hierarchy during an overwrite import.
- An additive import adds valid edges. In this mode, `null` deletes and `{}` makes no change.
- Web, Desktop, ZIP export, clone, and multicamera paths carry the configuration.

DIVE Annotation JSON, VIAME CSV, KPF, and NIST files do not carry the hierarchy. PR 5 adds KWCOCO
transport.

### Type editing

- Hierarchy types remain ordinary rows in the Type List. This PR does not add a tree editor.
- Rename updates all valid hierarchy references and linked camera tracks.
- DIVE rejects a rename that creates a cycle, a parent conflict, or a duplicate track pair.
- Delete Type remains blocked while a camera still uses the type.
- Deleting an unused type removes its filter and style settings. It keeps hierarchy edges.
- **Prevent Cascade Types** is disabled while a hierarchy is active. DIVE keeps its saved value.

Counterpart tracks now copy the complete confidence vector from the source track. Delete Type now
checks the tracks in every camera. These two corrections also apply to flat multicamera datasets.

### Validation

The server and client use the same validation examples. Server code validates stored hierarchy
data. Client code handles traversal, rename, selection, and editing.

### Manual tests

Test data: download and extract
[...]

PR #1849, opened 2026-08-14
Replace mutable merged tracks with read-only projections

PaulHax:
# Replace mutable merged tracks with read-only projections

The last PR fixed type mutation for multicamera. This PR removes that mutable Track clone entirely, replacing it with a read-only TrackProjection (new file TrackProjection.ts), and routes mutations of notes, attributes, feature attributes, keyframes, interpolation, through new CameraStore commands the same way #1848 did for classification.

The multicamera track list returned a cloned `Track` model. The clone still had mutation methods
and notification state. A caller could change the clone without changing the stored camera tracks.

The single-camera path returned a live model. Thus, the two paths had different read and write
behavior.

### Changes

- Add a data-only `TrackProjection` for logical-track reads.
- Give single-camera and multicamera callers the same read interface.
- Remove mutation methods, notification state, and revision references from projections.
- Add pure helpers for display type, features, interpolation, split checks, and keyframe navigation.
- Cache each projection until a source camera track or camera set changes.
- Route notes, attributes, feature attributes, keyframes, and interpolation through `CameraStore`
  commands.
- Remove the obsolete notification bypass from `Track.merge`.

An edit in any camera invalidates the related projection. This includes an edit in a camera that is
not the canonical camera.

### Manual tests

Test data: download and extract [classification.zip](https://github.com/user-attachments/files/31053807/classification.zip). Paths below are relative to `dive-classification-test-data/`.

#### Multicamera row edits

1. Create the dataset from
   `dive-classification-test-data/classification/multicamera-linked-types/`.
2. Edit a note, track attribute, detection attribute, and keyframe; then save, reload, and switch
   cameras.

Verify: The logical-track view shows the saved note, attributes, and keyframe geometry after reload.

#### Projection refresh from either camera

1. Create the dataset from
   `dive-classification-test-data/multicamera/divergent-classification-replicas/`.
2. Edit track 2 from port, then edit it from starboard.

Verify: The list and Track Details refresh after either camera track changes.

### Stack

This is PR 6 of 9. [Previous: centralize classification changes](https://github.com/Kitware/dive/pull/1848).
[Next: make lifecycle operations classification-safe](https://github.com/Kitware/dive/pull/1850).

6. **Current — Replace mutable merged tracks with read-only projections**

This PR is stacked on `hierarchical-classification-pr2-mutations`.

Overall: 24 files, +808/-151. Commits: `523556b9` through `9bc865e1`.

### DIFF
--- a/client/dive-common/components/Attributes/AttributesSubsection.vue
+++ b/client/dive-common/components/Attributes/AttributesSubsection.vue
@@ -54,28 +54,22 @@ export default defineComponent({
     const sortingMethodIcons = ['mdi-sort-alphabetical-ascending', 'mdi-sort-numeric-ascending'];
     const sortingMode = ref(0);
 
-    const selectedTrack = computed(() => {
-      if (selectedTrackIdRef.value !== null) {
-        return cameraStore.getAnyTrack(selectedTrackIdRef.value);
-      }
-      return null;
-    });
-
-    // Using Revision to nudge the attributes after updating them
+    // Logical-track reads: same projection as the track row, so Detection
+    // Attributes stay visible when the selected camera has no replica.
     const selectedAttributes = computed(() => {
-      if (selectedTrack.value && selectedTrack.value.revision.value) {
-        const t = selectedTrack.value;
-        if (t !== undefined && t !== null) {
-          if (props.mode === 'Track') {
-            return t;
-          }
-          if (props.mode === 'Detection') {
-            const [real] = t.getFeature(frameRef.value);
-            return real;
-          }
-        }
+      if (selectedTrackIdRef.value === null) {
+        return null;
       }
-      return null;
+      try {
+        const projection = cameraStore.getTrackProjection(selectedTrackIdRef.value);
+        if (props.mode === 'Track') {
+          return projection;
+        }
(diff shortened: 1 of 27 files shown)


BryonLewis:
There is a specific error on this PR with the changes.

If you have two camera system "left" and "right".
1. You hit "n" on keyboard or start a new track while the "right" camera is selected
2. You click Escape to not draw the left camera but get out of creation mode
3. Click "n" again or new to start a new track while the right camera is selected
4. You are then in creation mode with the right camera selected, you then click into the left camera and draw the track/rectangle
5. In the detail view the track in the left camera has no track details attribtue available, it's because the TrackProjection system thinks the right camera owns it because newtrack was initialized when that camera was selected.  This hould be

BryonLewis (dismissed):
I added a commit to solve an issue where starting a track on one camera and actually drawing it on another would cause it to be virtually bound to the starting camera.  This binds it to where it is drawn instead of where it was started.

Also did another update where if a track is selected detection attributes show up no matter what camera is selected.  They only don't show up if there is no keyframe on any camera for that specific frame.

BryonLewis (approved):

PR #1852, opened 2026-08-14
Define raw and resolved classification boundaries

PaulHax:
# Define raw and resolved classification boundaries

A track stores raw confidence pairs. The viewer can resolve a different display type from those
pairs when a hierarchy and filters are active.

This PR defines which features use the resolved display type. It also defines which features must
use the raw stored pairs.

### Resolved display type

- Track and detection attribute filters use the same resolved pair as the viewer.
- Resolution uses hierarchy depth, checked types, and confidence thresholds.
- Suppression regions compare their type with the resolved display type.
- Lists, counts, and canvas code use the same suppression resolver.
- The suppression cache includes hierarchy and filter state.
- The cache also includes **Prevent Cascade Types** state.
- Attribute-name suppression keeps its existing attribute behavior.

### Raw stored pairs

- **Checked Types Only** export compares checked names with raw pair names.
- Filtered DIVE JSON removes unchecked pairs from exported vectors.
- Full-archive export uses the same raw-name rule.
- Export filtering does not change stored tracks.
- Server Library labels use the raw pair with the highest confidence.
- A tie keeps the first stored pair.

The server cannot use viewer resolution for Library labels. It does not have each viewer's checked
types or confidence thresholds.

### Related corrections

- Multicamera save keeps global attribute and filter definitions when the dataset uses named camera
  maps.
- The bulk type-update action appears only during multi-track editing.
- The bulk action cannot run with a hidden default during normal track selection.

### Manual tests

Test data: download and extract
[classification.zip](https://github.com/user-attachments/files/31053807/classification.zip). Paths
below are relative to `dive-classification-test-data/`.

#### Resolved attribute filters

1. Create the image-sequence dataset with the files in `display/resolved-attribute-filters/`, add
   the documented track and detection filters for `juvenile-red-snapper`, then change each filter
   type to `fish`.

Verify: Each track and detection filter uses the resolved type without a reload.

#### Resolved suppression regions

1. Create the image-sequence dataset with the files in `display/resolved-suppression-regions/` and
   set the suppression region type to `fish` at the default 99% overlap.
2. Uncheck `juvenile-red-snapper` and `red-snapper`, then recheck both.

Verify: Track 2 disappears and returns without a reload.

#### Raw confidence-pair export

1. Create the image-sequence dataset with the files in `export/raw-confidence-pair-filtering/`;
   leave `juvenile-red-snapper` and `fish` checked, and uncheck `red-snapper`.
2. Export with **Checked Types Only** and inspect track 1 in the exported DIVE JSON.

Verify: The export contains only the expected raw pair names, and the source dataset does not
change.

### Stack

This is PR 9 of 9. [Previous: add DIVE KWCOCO classification support](https://github.com/Kitware/dive/pull/1851).

9. **Current — Define raw and resolved classification boundaries**

This PR is stacked on `hierarchical-classification-pr5-kwcoco`.

Overall: 32 files, +723/-86. Commits: `90917889` through `25e427be`.

### DIFF
--- a/client/dive-common/components/TrackDetailsPanel.spec.ts
+++ b/client/dive-common/components/TrackDetailsPanel.spec.ts
@@ -10,6 +10,7 @@ const state = vi.hoisted(() => ({
   displayPairIndex: vi.fn(() => 1),
   track: null as Track | null,
   multiSelectList: [] as number[],
+  editingMultiTrack: false,
   acceptTrackType: vi.fn(),
   assignTrackType: vi.fn(),
 }));
@@ -40,7 +41,7 @@ vi.mock('vue-media-annotator/provides', () => ({
     typeStyling: ref({ color: (type: string) => `color:${type}` }),
   }),
   useEditingGroupId: () => ref(null),
-  useEditingMultiTrack: () => ref(false),
+  useEditingMultiTrack: () => ref(state.editingMultiTrack),
   useGroupFilterControls: () => ({ allTypes: ref([]) }),
   useCameraStore: () => ({
     camMap: ref(new Map([['singleCam', { groupStore: undefined }]])),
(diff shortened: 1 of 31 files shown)


BryonLewis (dismissed):
I added one fix that was basically making sure that the display and filtered exports respect the same 0 values.  Before they didn't so you would have a mis-alignment between exporting and what you saw.

BryonLewis (approved):

PR #1854, opened 2026-08-17
Auto Camera Registration Pipeline

romleiaj:
Add the ability to automatically register cameras with MINIMA-LoFTR. Changes aggregation to be over multiple frames, rather than a single frame as in the keypoint gui. Updates registration json to hold these per-frame points.


romleiaj:
# Auto-register cameras from many frames; registration format v2

Lining up two cameras today means clicking matching points by hand on a single
frame. It's slow on EO/IR, and one frame isn't enough for a good fit.

This adds an **Auto Register** button that does it for you. It runs a VIAME
pipeline that finds matching points across many frames at once and folds the
results into the dataset.

## What's new

**Auto Register.** Open Camera Registration, click *Auto Register Frames…*,
pick how many frames to use, and run. DIVE spreads the frames over the video,
sends them to VIAME, and merges back whatever matched. You can also hand-pick
specific frames instead. Frames the matcher couldn't use come back switched off
with a reason. Same on desktop and web.

**Registration Frames list.** A new panel section listing every frame the fit
uses, split into Auto and Manual. Each row shows its point count and how well it
agrees with the rest; you can switch it off, jump to it, or delete it. Frames
are marked on the timeline, and 3-camera rigs get a loop-closure check.
"Pick points" is now "Edit points".

**Format v2.** Points now attach to the specific pair of images they were picked
on, not just to a pair of cameras. Re-running the matcher replaces its own
frames and leaves hand-picked points alone. No migration needed — v1 never
shipped.

**Frame alignment.** Rigs whose cameras drop frames independently were landing a
frame or two off when jumping between cameras. Fixed.

## Testing

Needs VIAME with the ALIGN-CAMERAS add-on and a 2- or 3-camera image dataset.

- **Auto register** — run with defaults. ~10 frames under Auto, status line
  shows the fit, Align View warps, saved file says `"version": 2`.
- **Queue** — queue 3 frames with `+` and run. Only those 3 appear.
- **Review** — uncheck or delete a frame. RMS updates, its timeline marker dims
  or vanishes.
- **Replace** — add 2 manual points, re-run with Replace on. Manual points
  survive, old auto frames are gone, no "discard edits?" prompt.
- **Triplet** — 3-cam run with All pairs. Loop-closure line appears.
- **Web** — same run on web-girder; result shows up in the panel.

## Notes

Standalone on `main`, rebased onto be70fba7. Needs VIAME's `align_cameras`
process. `sealtk/registration-json` (#1780) should land around the same time so
both sides of the v2 format ship together. The camera display-order commit that
used to live here is already on `main` as #1857, so it dropped out in the
rebase.

52 files, +5087/-416. Commits 8be88524 through 4a9faef0.

BryonLewis (approved):
Made some minor changes

- fixed image path mix-up, the system was using transcoded extensions (tif to png) on the source data folder.  Fixed it so it would use the proper folder for the image-paths when doing auto-registration
- use `runtimeParams.imagePairs` in `writeMultiCamStereoPipelineArgs`.  I believe the server was doing this but desktop wasn't
- Properly fail on post-run failures for desktop jobs
- Updated to ensure desktop/web are using the same camera ordering mechanism when passing in camera names to pipelines.

PR #1855, opened 2026-08-17
Desktop pipefile job fixes

romleiaj:
Path sanitization and standardization across jobs, remove spaces from any names.


BryonLewis (approved):
I think think this is fine to merge in.

PR #1859, opened 2026-08-19
Point GPU worker python symlink at 3.12

mattdawkins:
- kitware/viame:gpu-algorithms-web base moved to Ubuntu 24.04 / Python 3.12; the `ln -fs` to python3.10 now creates a dangling symlink that only fails at runtime
- No other DIVE code depends on the base Python version (girder/worker images and the uv-managed 3.11 venv are independent)

### DIFF
--- a/docker/girder_worker_gpu.Dockerfile
+++ b/docker/girder_worker_gpu.Dockerfile
@@ -30,8 +30,8 @@ RUN chmod +x /tini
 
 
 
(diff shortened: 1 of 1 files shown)


BryonLewis (approved):
Known about change to the base image.

This symlink is to make it so viame doesn't try to use the uv system version of python and will use the proper version of viame tasks.

PR #1861, opened 2026-08-21
Import the frame rate a COCO video records

mattdawkins:
COCO was the only import path that dropped the dataset frame rate. CSV reads it from the `fps:` header (`viame.py` / `viame.ts:35`) and NIST sets it directly (`common.ts:1425`), but the COCO branch never populated `meta.fps`.

- `kwcoco.py`: `frame_rate_from_coco()` reads `videos[].annotation_fps`; the import route folds it into meta the same way the CSV branch does
- `coco.ts`: `frameRateFromDocument()` does the same for desktop; `parseFile` returns it in meta, which `common.ts` already merges
- Image sequences emit no `videos` table and so carry no rate — not an error
- Rejects unusable values (0, negative, string, bool, inf/nan) so nothing downstream sees `fps: 0`

The field is `annotation_fps` because it is the rate detections were produced at — the downsampled rate, not the video's native one. VIAME writes it there (VIAME/VIAME@9b1227cc6). Neither MS-COCO nor kwcoco define a frame-rate field, so the spelling is a convention shared between the two projects rather than a standard; DIVE's own dataset metadata key stays `fps`.

Verified against real `process_video.py` COCO output: desktop `parseFile` returns `meta.fps = 5` for a video and `undefined` for an image list; the server helper returns `5.0` / `None` on the same two files, plus the guard cases.

Not run: vitest, eslint and tsc — no `node_modules` in this checkout. The new server tests were executed directly and pass; the desktop spec is unrun.

Out of scope: neither exporter emits a `videos` table, so DIVE -> COCO -> DIVE still loses the rate. Writing it on export means introducing that table first.

### DIFF
--- a/client/platform/desktop/backend/serializers/coco.spec.ts
+++ b/client/platform/desktop/backend/serializers/coco.spec.ts
@@ -316,6 +316,40 @@ describe('COCO serializer', () => {
     expect(parsedMeta).not.toHaveProperty('datasetInfo');
   });
 
+  // --- annotation fps on videos[] ---
+
+  it('writes videos[].fps for video datasets and restores it on re-import', async () => {
+    const videoMeta = { ...imageMeta, type: 'video' as const, fps: 5, name: 'clip' };
+    await serializeFile('/output/video.coco.json', annotationSchema, videoMeta);
+    const out = await fs.readJSON('/output/video.coco.json');
+    expect(out.videos).toEqual([{ id: 1, name: 'clip', fps: 5 }]);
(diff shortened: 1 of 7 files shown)


BryonLewis (dismissed):
Added some documentation about the FPS and added some code to export the FPS when writing a COCO output from DIVE.

I think I will update the DIVE Annotation JSON in a new PR so it will import/export the video FPS as well if it exists.  Seems like something that should be unified across COCO/VIAMECSV/DIVE Annotation JSON.

PR #1863, opened 2026-08-21
Fix/dive params bugs

Louis-Pagnier-KW:
Small bug fixes on dive params.
I also renamed output of disparity pipeline from `depth_map%06d.png` to `map%06d.png` as it can output disparity map and depth maps.


BryonLewis (approved):
Thanks for doing this and testing the fixes.

PR #1864, opened 2026-08-21
swap from video.fps to info.video_annotation_fps

BryonLewis:
This swaps the video.fps to an info.video_annotation_fps in COCO for importing/exporting.
This also has future compatibility with multiple videos in the future if needed.


BryonLewis:
close sticking with the other format

PR #1867, opened 2026-08-21
Render hierarchical types as a tree

PaulHax:
# Render hierarchical types as a tree

[Bryon called out three follow-ups during review of the original hierarchy PR](https://github.com/Kitware/dive/pull/1847#pullrequestreview-4938834044):
render the Type List as a tree, let a parent toggle its children, and roll child counts into
parents. For datasets without a type hierarchy, the Type List retains its existing flat behavior.

Screenshot 2026-08-21 160322

https://github.com/user-attachments/assets/c8f20ad2-cd02-44e9-a969-f2b2b8529d30

### Changes

- Indent descendants and add branch disclosure controls.
- Give parent checkboxes checked, unchecked, and indeterminate states; toggling a parent applies to
  its complete subtree.
- Keep ancestor paths visible during search and sort siblings within each branch.
- Compact a shared unused leading lineage behind **Expand Parents** and **Compact Parents**.
- Roll descendant totals and current-frame counts into their parents.
- Make parent maximum-count navigation include descendants and use the same selected-camera and
  suppression rules as the displayed counts.
- Keep structural parents visible with **Show Empty** off, without showing disclosure controls for
  children that are filtered out.

### Manual tests

Test data: upload and extract
[hierarchical-type-list.zip](https://github.com/user-attachments/files/31320128/hierarchical-type-list.zip)
The ZIP opens to `hierarchical-type-list/` with self-contained `synthetic/` and `sefsc/` datasets.

#### Type List controls and counts

1. Create a 1 FPS image-sequence dataset from `synthetic/images/`, then import
   `synthetic/annotations.json` and `synthetic/config.json`.
2. Exercise branch collapse, parent checkboxes, search, **Show Empty**, all three sort modes, and
   **Expand Parents** / **Compact Parents**.
3. Compare counts at frames 0, 2, and 4, then use the `fish` maximum-count button.
4. Set the suppression region type to `fish`; uncheck and recheck `juvenile-red-snapper` and
   `red-snapper`.

Verify: The tree and tri-state checkboxes follow the hierarchy, search keeps the ancestor path, and
the compact breadcrumb is `eukaryota › animalia › chordata`. Parent counts include descendants;
`fish` totals four tracks and seeks to frame 2. Suppressed detections do not reappear in parent
counts or peak navigation.

#### Real SEFSC-SEAMAP hierarchy

1. Create a 5 FPS video dataset from the MP4 in `sefsc/`, then import its VIAME CSV and
   configuration from the same directory.
2. Search, collapse branches, and sort the Type List.

Verify: The 24 real tracks render in the correct hierarchy from `eukaryota` through the eight
observed species.

This is a follow-up to
[Define raw and resolved classification boundaries](https://github.com/Kitware/dive/pull/1852).

### DIFF
--- a/client/dive-common/components/GroupSidebar.spec.ts
+++ b/client/dive-common/components/GroupSidebar.spec.ts
@@ -1,0 +1,86 @@
+// @vitest-environment jsdom
+/* eslint-disable vue/one-component-per-file -- harness components for shallow mounting */
+/// <reference types="vitest" />
+import {
+  defineComponent, h, reactive, ref,
+} from 'vue';
+import { shallowMount } from '@vue/test-utils';
+import { clientSettings } from 'dive-common/store/settings';
+import FilterList from 'vue-media-annotator/components/FilterList.vue';
+import GroupSidebar from './GroupSidebar.vue';
+
+const provideMocks = vi.hoisted(() => ({
+  groupFilterControls: undefined as unknown,
+  groupStyleManager: undefined as unknown,
+}));
+
+vi.mock('dive-common/vue-utilities/prompt-service', () => ({
+  usePrompt: () => ({ prompt: vi.fn(), visible: () => false }),
+}));
+
+vi.mock('vue-media-annotator/provides', () => ({
+  useCameraStore: () => ({ camMap: ref(new Map()) }),
+  useGroupFilterControls: () => provideMocks.groupFilterControls,
+  useGroupStyleManager: () => provideMocks.groupStyleManager,
+  useHandler: () => ({ seekFrame: vi.fn() }),
+  usePendingSaveCount: () => ref(0),
+  useReadOnlyMode: () => ref(false),
+  useSelectedCamera: () => ref(''),
+  useTime: () => ({ frame: ref(0) }),
+}));
+
+describe('GroupSidebar type filter', () => {
+  it('wires the group FilterList to flat show-empty behavior', () => {
+    clientSettings.typeSettings.trackSortDir = 'a-z';
+    clientSettings.typeSettings.filterTypesByFrame = false;
+    clientSettings.typeSettings.suppressionType = '';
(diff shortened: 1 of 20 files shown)


BryonLewis (approved):
I think this is good as is, below is a list of other things I found while reviewing this PR.  They can be broken out into future PR updates:

- Afterwards lets update the behavior on filter-by-frame and the header checkbox toggling on/off all types.  It should probably only toggle the current types that are visible on the page.  This was an existing issue I just noticed while testing this PR.  When filter-by-frame is on it shows only the types for that frame, but the header checkbox toggle will toggle on/off all frames.  This is unexpected because if you use the search modifier it only toggles the types that the search displays in the list.
- I think we need to do some CSS modifications or maybe some options to reduce the amount of space taken up in the type list.  Currently between the tree structure, the total:frame counts and filter/ other buttons it is getting a little crowded.  Below are some options
	- reduce the offset for the tree indentation if possible
	- The space between the chevron and the checkbox can be reduced a bit
	- Make it so Total Count: Frame Count are toggle-able individually in the settings.  I.E you can have both on, you can one or the other
- Type Settings update
	- Deletion of hierarchy/parent types, I don't think they can be removed right now
	- Ability to add your own hierarchy from the interface.  We add the lowest level types but it may be nice to edit the types in a hierarchical way at some point and create the hierarchies.
- Option for cascading up types.  Somewhere in the settings I think if you turn off a leaf on the tree it should automatically move the type up the branch to the one that is visible.  This can be a setting that can be set by the user depending on what they want to see.

mattdawkins:
If it's easy, it might be nice to have a left oriented bar to show the hierarchy better, something like: 

image

BryonLewis:
It isn't, we are using Vuetify 2 and their system, We can save that sort of thing for a future update if we want to write a bunch of custom code to support this.

PR #1869, opened 2026-08-28
Run the server lint gate in CI and format what it finds

PaulHax:
CI runs `tox -e testunit` for the server but never `tox -e lint`, so the server side has had no formatting gate since the black pin arrived with #1658. The client is linted in CI; the server was not. Five spots had drifted from the pinned `black == 26.5.1` in the meantime.

Add the lint step to the server job, next to the client's, and apply `tox -e format`. Two of black's rewrites left an implicit string concatenation on one line, which parses fine and reads badly; those two messages are single strings now, which black leaves alone.

Formatting only, no behavior change sides the CI blank.yml change

### DIFF
--- a/.github/workflows/blank.yml
+++ b/.github/workflows/blank.yml
@@ -77,6 +77,9 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install tox
       run: pip install tox
+    - name: Lint
+      run: tox -e lint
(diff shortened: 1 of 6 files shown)


BryonLewis (approved):
This may result in a significant change if it requires that much refomatting but I guess it is the best for the code eventually.

PR #1870, opened 2026-08-28
Reclaim Type List row width and read counts as a fraction

PaulHax:
# Reclaim Type List row width and read counts as a fraction

Follow-up to [Bryon's #1867 review](https://github.com/Kitware/dive/pull/1867#pullrequestreview-5030473848).

Deep hierarchy rows run out of room for the type name.

### Changes

- Tighten tree indentation, disclosure width, and checkbox spacing; flat rows are unchanged.
- Fill the Type List's available height instead of leaving space below its rows.
- Render counts as `3 / 12  red-snapper` instead of `12 : 3  red-snapper`.
- Add **Show Total Count** and **Show Frame Count** settings, both on by default.
- Let **Compact Parents** absorb the deepest unused shared ancestor.
- Truncate breadcrumbs at the front so the most specific parent remains visible, while preserving
  character order for names containing digits or punctuation.
- Skip per-frame count work when it is not needed for display, sorting, or filtering.

image

### DIFF
--- a/client/dive-common/store/settings.spec.ts
+++ b/client/dive-common/store/settings.spec.ts
@@ -1,0 +1,30 @@
+// @vitest-environment jsdom
+/// <reference types="vitest" />
+
+describe('clientSettings hydration', () => {
+  it('keeps every type-list switch a reactive key when nothing is stored', async () => {
+    localStorage.removeItem('Settings');
+    vi.resetModules();
+
+    const { clientSettings } = await import('./settings');
+
+    expect(Object.keys(clientSettings.typeSettings)).toEqual(expect.arrayContaining([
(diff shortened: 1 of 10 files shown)


client/dive-common/store/settings.spec.ts
```diff
@@ -0,0 +1,30 @@
+// @vitest-environment jsdom
+/// <reference types="vitest" />
```
BryonLewis, line 2: I think after this merges they may need to be updated to be more in line with #1872 
I may just doa  secondary PR.

BryonLewis (approved):
I'll handle in another PR the one comment I have about the changes from the PR with testing.

PR #1871, opened 2026-08-28
Scope the Type List header checkbox to the displayed rows

PaulHax:
# Scope the Type List header checkbox to the displayed rows

Follow-up to [Bryon's #1867 review](https://github.com/Kitware/dive/pull/1867#pullrequestreview-5030473848).

With **Filter Types by Frame** on, the list showed only the current frame's types but the header
checkbox still toggled hidden types. Those selections also feed the delete button.

### Changes

- Scope the header checkbox to the rows actually displayed — after search, frame filtering,
  collapsed branches, and **Compact Parents**.
- Scope the delete button the same way. It read every checked type, so it deleted annotations of
  types the list was not showing, and stayed enabled when the only checked types were hidden.
- Keep parent-row checkboxes subtree-wide: parent means "this branch"; header means "what I can
  see."
- Show an empty list as unchecked and disable the header checkbox, and say whether the list is
  empty because nothing is defined or because the filters excluded everything.
- Add the missing default for **Filter Types by Frame**, making the setting reactive on fresh
  profiles.
- Hide hierarchy branches with no annotations while **Show Empty** is off. The candidate set took
  every type with a child, so an unannotated branch held a row whose own leaves stayed hidden.
- Cover flat and hierarchical lists with regression tests.

Frame-filtered rows disappear when unchecked; turning the frame filter off brings them back. This
matches existing per-row behavior.

A parent's tri-state still rolls up its complete subtree, so an ancestor of a used type can read
indeterminate against a hidden checked descendant. That is the row's documented meaning.

### DIFF
--- a/client/dive-common/store/settings.spec.ts
+++ b/client/dive-common/store/settings.spec.ts
@@ -1,0 +1,16 @@
+// @vitest-environment jsdom
+/// <reference types="vitest" />
+
+describe('clientSettings hydration', () => {
+  it('keeps every type-list switch a reactive key when nothing is stored', async () => {
+    localStorage.removeItem('Settings');
+    vi.resetModules();
+
+    const { clientSettings } = await import('./settings');
+
+    expect(Object.keys(clientSettings.typeSettings)).toEqual(expect.arrayContaining([
+      'filterTypesByFrame',
+    ]));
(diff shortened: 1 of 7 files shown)


BryonLewis (approved):
Just pulled and tested, made a small PR to align with the other vitest PR about the jsdom #1872.  Thanks for doing this.

PR #1875, opened 2026-08-31
Add a foundation-stereo option for warping detections between stereo cameras

mattdawkins:
## What

Adds a second stereo correspondence method next to the NCC template matcher: the Fast-FoundationStereo ONNX export VIAME ships in its `FAST-FDN-STEREO` add-on. **Track Settings → Stereo Settings → Point matching** picks between *Lower Accuracy, Higher Speed* (NCC, default) and *Higher Accuracy, Lower Speed* (foundation). Web only; desktop's method comes from the VIAME interactive stereo config.

## How

- **Model delivery**: `GET dive_configuration/stereo_foundation_model[/spec]`. Girder reads the `FAST-FDN-STEREO-WEB` row (a bare `.onnx`, platform `WEB-ONLY`) of VIAME's `download_viame_addons.csv`, downloads it once into `DIVE_MODEL_CACHE_DIR` (compose volume `girder_model_cache`), verifies the md5 and streams it; the matcher reads the input size from the graph. The client caches the bytes in the browser Cache API keyed by md5, so a re-published model is picked up with no DIVE change. `WEB-ONLY` rows are hidden from the add-on installer.
- **Preprocessing**: rectify + resize fused (OpenCV-style focal/principal scaling), RGB, ImageNet-normalised — the contract of `plugins/onnx/fast_foundation_stereo.py`. Both matchers take RGBA frames.
- **Precompute + cache**: the viewer's frame is watched and both warp directions prepared whenever the foundation method and auto-compute are on. The matcher keeps an LRU of 8 disparity maps, runs one inference at a time, dedupes concurrent requests, drops stale prefetches unless a warp is waiting on them, and remembers an inference failure instead of retrying every frame.
- **Runtime**: onnxruntime-web's *native* WebGPU provider (`onnxruntime-web/webgpu`, lazily imported) at `graphOptimizationLevel: 'basic'`. Verified probe-by-probe against CPU onnxruntime: the JSEP provider returns an all-zero cost volume for this graph on 1.27/1.29/1.31-dev, and the extended fusions corrupt the GRU gates by ~2 px; this configuration matches CPU to 5e-5 px. wasm is refused (the model needs GBs of activations).

## Model

The published file must be VIAME's web build of the export (`plugins/onnx/export_fast_foundation_stereo_web.py`): NVIDIA's stock single-file export materialises 1.5 GB correlation tensors and uses 3-D `ConvTranspose` and asymmetric 3-D conv padding that no browser provider runs. The web build is numerically identical (~1e-4 px) and expresses every 3-D conv as 2-D convs (the providers' 3-D conv kernel is naive; this took the browser time from 6.8 s to 0.69 s per pair). The desktop add-on zip (`FAST-FDN-STEREO`) is separate and carries the onnxruntime-CUDA export plus a TensorRT engine.

## Status

- Unit tests (rectification incl. resize, preprocessing, caching/prefetch/LRU, failure latching; server CSV/zip/md5) + lint clean.
- Browser output verified against CPU on the fixture pair on an RTX 4090 (windowed Chrome, native WebGPU provider, `basic`): mean |diff| 0.000 px, **0.69 s per pair at 576×960**, 0.26 s at 320×736. JSEP and level `all` confirmed wrong on real hardware. **Not yet exercised**: the dropdown and frame watcher in a running viewer.

https://claude.ai/code/session_013oFD6dP2EQY3R9UdXzx7dL

### DIFF
--- a/client/dive-common/components/TrackSettingsPanel.vue
+++ b/client/dive-common/components/TrackSettingsPanel.vue
@@ -7,6 +7,7 @@ import {
   computed,
 } from 'vue';
 import { clientSettings } from 'dive-common/store/settings';
+import { STEREO_MATCH_METHODS } from 'dive-common/use/stereo/stereoMatcher';
 import isDesktopRuntime from 'dive-common/isDesktopRuntime';
 
 export default defineComponent({
@@ -40,6 +41,7 @@ export default defineComponent({
       showMultiCamToolbar: 'Show multi-camera tools in the top toolbar when a track is selected',
       stereoUpdateLengths: 'When a line annotation is modified on a detection that is linked across both cameras, recompute its stereo measurement (length, midpoint, range, RMS) automatically.',
       stereoAutoCompute: 'When an annotation is drawn on one camera and the other camera has no detection for it yet, automatically warp it to the other camera using stereo disparity.',
+      stereoMatchMethod: 'How a point is matched on the other camera. Template matching correlates the source patch along the epipolar line. Foundation stereo runs a dense disparity network over the pair once, which is steadier where the patch is hard to match but needs its model to be configured.',
     });
     const modes = ref(['Track', 'Detection']);
     // Add unknown as the default type to the typeList
@@ -52,6 +54,7 @@ export default defineComponent({
       help,
       modes,
       typeList,
+      stereoMatchMethods: STEREO_MATCH_METHODS,
     };
   },
 });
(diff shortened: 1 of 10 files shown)


BryonLewis:
A 100MB stereo file shouldn't be bundled in the app or in the repo.  I think for the Web version it should be in a viame.kitware.com location and there should be a configuration option for the server that points at either the direct URL or girder Item.  Maybe it is a place where we can configure multiple models in the future if we start using moe onnx models.

For the desktop version I would probably have that onnx model embedded inside of the VIAME install and the express server in electron would serve that to the client when requested.  It could fall back to using the web version if it has an internet connection

Then the client when requested the first time would download and cache it for use.

I think this makes it a bit easier in the future for deployment of other models as well if we are going to rely on onnx web runtimes for other functionality.

PR #1879, opened 2026-09-02
Auto-register video multicam rigs on web; drop the stale video reader

romleiaj:
Running Auto Register on a rig whose cameras are videos failed on the girder worker before VIAME was ever invoked:


romleiaj:
# fix/camera-registration — summary

Four commits on top of `efe061e2` (Auto Camera Registration Pipeline, #1854).

---

## 1. `530e58b3` — Auto Register works on video rigs (web)

**Was:** running Auto Register on a rig whose cameras are videos failed
immediately with `Image-pair subsets are only supported for image-sequence
cameras (camera "XXX")`.

**Why:** DIVE Desktop extracts the chosen frames from each video into stills
before handing them to VIAME. The web/girder worker never got that half of the
feature — it just raised an error instead.

**Now:** the server extracts the frames with ffmpeg, exactly like desktop does,
and the results get mapped back to the right frame numbers when they're saved.
Also stopped the run from telling VIAME to expect a video once every input has
become a list of stills.

---

## 2. `13adf773` — Two Auto Register annoyances

**Was:**
- The panel says "see the Jobs tab for progress", but leaving the page popped an
  unsaved-changes warning that only offered *Discard* — and the dataset is
  read-only while a job runs, so there was no way to save.
- On web, if the job failed, the progress bar just kept spinning (for 30
  minutes).

**Now:**
- Registration edits are saved automatically right before the job starts, so
  there's nothing to warn about. As a bonus this makes "Replace existing" work
  properly — previously its changes were silently undone.
- Web now watches the actual job status like desktop does, so a failed or
  canceled job is reported straight away.

---

## 3. `67336ea8` — Pipelines run on TIFF (large-image) cameras

**Was:** a 3-cam run on a rig with a TIFF camera died with
`unexpected metadata {... 'type': 'large-image' ...}`.

**Why:** DIVE files `.tif` under "large-image" rather than "image sequence", and
the pipeline runner had no support for that type at all — it could only fetch
image sequences and videos.

**Now:** large-image cameras are downloaded and fed to VIAME the same way an
image sequence is. This covers single-camera and multi-camera runs, and training
gets it for free.

---

## 4. `167e3f03` — Tell the user why a run produced nothing

**Was:** a 3-cam run over flat sea ice rejected all 14 candidate frames and
fitted nothing, but the panel still said *"Auto Register complete: review the
registration frames below."* The real reason was only in the job log. Re-running
it said *"the results matched the registration already stored"*, which sounds
like everything is fine.

**Now:** the panel reports what the run actually achieved. If nothing fitted, it
shows a red line naming the reason and count, e.g.:

> Auto Register fitted no camera pairs: every candidate frame was rejected
> (14 low_texture). Try frames with more visible structure.

A partly successful run says how many pairs fitted and how many frames were
rejected. A clean run reads exactly as before.

---

## Testing

All four are covered by unit tests. Full suites pass: 1406 client tests, 421
server tests, plus lint and typecheck.

BryonLewis (approved):
Gave it a quick look and I think it is okay.

PR #1880, opened 2026-09-02
Stop passing --no-embedded-pipe to viame train

mattdawkins:
- VIAME's training tool now produces runnable pipelines by default, with `--embedded-pipe` as the opt-in for fragments; the old flag is a deprecated no-op
- Drops it from the server and desktop training invocations and the arg-sanitizer test
- Pair with a VIAME build containing the inverted default; on older builds the flag is still required


BryonLewis (approved):
Simple enough, looks good.

PR #1885, opened 2026-09-03
Desktop: list every trained pipe, not just the first

mattdawkins:
- VIAME training can emit both `category_models/detector.pipe` and `tracker.pipe`. `getPipelineList` used `find` to take the first `.pipe` in `readdir` order and dropped the rest, so the tracker was never runnable from desktop (which one won was FS-dependent: alphabetical on NTFS, hash order on ext4).
- Now emits one `trained` entry per `.pipe`, sorted for deterministic order, suffixed ` detector` / ` tracker` to match web (`crud_rpc.py:_load_dynamic_pipelines`).
- Single-pipe models still produce exactly one entry; their display name gains the ` detector` suffix, same as web.
- Deleting either entry still removes the whole model folder, unchanged and identical to web.
- Spec: new `detectorAndTrackerTrainingJob` fixture with both pipes + test asserting two entries; existing trained test now pins the name. 118 passed, eslint clean.

https://claude.ai/code/session_01Nzq6XQBXKUb1NRdaVpPYQZ

### DIFF
--- a/client/platform/desktop/backend/native/common.spec.ts
+++ b/client/platform/desktop/backend/native/common.spec.ts
@@ -497,6 +497,13 @@ beforeEach(() => {
             'trained_detector.zip': '',
           },
         },
+        detectorAndTrackerTrainingJob: {
+          category_models: {
+            'detector.pipe': '',
+            'tracker.pipe': '',
+            'trained_detector.zip': '',
+          },
+        },
       },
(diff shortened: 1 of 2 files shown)


BryonLewis (approved):
This looks proper, before we were just grabbing the detectors because I believe we didn't have trackers initially.  This looks like it

PR #1889, opened 2026-09-06
Add a scoring page that wraps viame score on desktop and web

mattdawkins:
- New top-level **Scoring** tab (desktop between Pipeline and Settings, web after Models) and a `Score` bulk action in the library.
- Score any number of sequences in one run: per row pick the computed and truth annotations (current, annotation set / revision on web, earlier annotation file or browsed file on desktop, or another dataset with the same footage). Pairs are exported as `computed/seq_N.csv` + `truth/seq_N.csv` and scored by one `viame score` folder-mode job.
- Results stored as `auxiliary/scoring_<ts>.json` on the first computed dataset; "Previous runs" lists every run the user can read; "Reuse setup" restores a run's sequences and parameters.
- Tabs: summary metrics (detection, localization, segmentation, keypoint, length, MOT/HOTA/KWANT), per-class table, confusion matrix (click a cell to filter errors), interactive PR/ROC curves (hover for threshold, consolidated vs per class, click a point then "Use as confidence filter"), threshold sweep with recommended per-class filters, per-object error browser with per-frame chart.
- All `viame score` options exposed in a parameters dialog (IoU, confidence, box/polygon matching, per-class, top-class, aux confidence, tracking, keypoint tolerance, sweep steps, filter estimator, default label, class synonyms).
- Web: `POST dive_rpc/score`, `GET dive_scoring`, `GET/DELETE dive_dataset/:id/scoring[/:resultId]`, `GET dive_dataset/:id/scoring_sources`; CSV export gains a `set` param. Desktop: `RunScoring` job on the CPU queue.
- Requires the matching VIAME change (VIAME/VIAME `tool-applets` 11f12f500) for `--match-mode`, `--output-matches` and inlined sweep curves.
- The shared viewer is untouched. Verified end to end on desktop; the web worker path is unit-tested only.

https://claude.ai/code/session_01YEefheVfnAJsDhdisaBmis


BryonLewis (approved):
Most changes are kept relatively separate from affecting other code.

Below are some changes I've made to this branch I think to imporve user experiance and during my testing.

- Source linking for scoring runs: Scoring UI updates so pairs can point at annotation sets, revisions, or on-disk sources, with viewer navigation into those annotations.
- Web dataset picker: Dialog-based dataset picker on web for choosing comparison datasets in scoring.
- PDF export cancel: Cancel support in the PDF preview flow for scoring exports.
- Sample scoring dataset script: samples/scripts/scoring/generate_dataset.py and README to generate test data for the scoring workflow.
- Annotation JSON import fixes: Better handling of .dive.json files (import path fix in the VIAME serializer and improved JSON annotation detection on import).
- Picker remembers last folder: Add-sequence / dataset picker opens to the last folder used.
- VIAME virtualenv fix: Clears VIRTUAL_ENV and adjusts PATH in viame_config.py so VIAME's embedded Python is not pointed at the DIVE venv.

PR #1891, opened 2026-09-08
Species List imports

romleiaj:
# Species list branch (`dive` @ `dev/species-list`)

## Features

- A KWCOCO file holding only a `categories` block is a **species list**: each `name` becomes a declared type, each `supercategory` a type hierarchy edge. It never touches annotations.
- Import it from the viewer's Import dialog (Overwrite replaces the declared list, additive adds to it), from the new **Species List** field at upload (Web page and Desktop import dialog, under advanced options), or by naming it to end in `species.json` beside the media.
- Desktop multicam imports now discover a `species.json` in the folder the cameras share, then beside a camera; differing lists across cameras warn instead of guessing.
- Repeated names, cycles, and conflicting parents fail the import with "No configuration was changed."
- On multicam datasets the declared types and hierarchy live on the parent, so any camera's import updates the whole dataset.
- Fixed: after an Overwrite import the viewer kept listing the dropped types and the next save restored them. Fixed: styleless species declarations leaked into the shared style store.

## Desktop testing done (2026-09-08)

- Multicam image-sequence import with `rockfish.species.json` in the advanced Species List field: four types declared, two under Sebastes.
- Viewer import of `extra.species.json` on that dataset exposed the stale-type bug (flavidus and melanops stayed listed at top level). Root cause and fix are in commit `ab1be4ff`.
- Manual walkthrough passed after the fix, on a multicam dataset:
  1. Import shows only the declared types, with the hierarchy intact.
  2. Overwrite import of example `rockfish.species.json` file resets to four types, two under Sebastes.
  3. Additive import of example `extra.species.json` file grows to five types, three under Sebastes.
  4. Overwrite import of `extra.species.json` shrinks to two types.
  5. A custom color survives a re-import that names the type and falls back to the palette when omitted
  6. Example `bad-duplicate.species.json` and `bad-cycle.species.json` files fail with "No configuration was changed." and leave the dataset untouched.
  7. Lock Types limits annotation to the declared names.
  8. With shared colors on, no species names leak into another dataset's Type List or Saved Styles.
  9. Multicam auto-discovery pre-fills the Species List field from the shared parent folder.
- Unit suites: server species-list tests (82 passed), Desktop native suites (208 passed), filter-control and shared-style tests (64 passed). Eight pre-existing DOM spec failures are Node 26 `localStorage` issues, unrelated.

## Still to do

Web checks need a Docker host: the separate Species List upload slot, automatic slotting of a file ending in `species.json`, and importing against one camera of a multicam dataset.


BryonLewis (approved):
Went through most of the changes (not really the tests) then tested the web version.  Seems to work and documentation updates are there.

PR #1900, opened 2026-09-09
Let the training dialogs email progress reports through viame monitor

mattdawkins:
Adds an optional **Email progress reports to** field to the desktop and web training dialogs. The address is passed to `viame train` as `--monitor-email`, which starts VIAME's training monitor next to the run (VIAME/VIAME main 879c9a753 and later). The monitor emails a start message, validation statistics every few epochs, error/deadlock notices and a final finished/failed report, and keeps `train.log` plus a status trail in the run's output directory.

Mail delivery is configured on the training machine rather than in DIVE: `VIAME_SMTP_SERVER` (and `VIAME_SMTP_USER` / `VIAME_SMTP_PASSWORD`) in the environment of DIVE Desktop or of the training worker, or a local sendmail on Linux. Without either the reports still go to the status trail. Documented in the training section of the pipeline docs.

Changes:
- `Api.runTraining` takes a trailing optional `monitorEmail`; the desktop job args and the server's `RunTrainingArgs` / `TrainingJob` carry it through.
- Desktop launcher and the training worker append `--monitor-email` when set.
- Shared `isValidEmail` helper validates the field in both dialogs; the train button stays disabled while the address is malformed.

Verified with typecheck, lint and the client unit tests; the desktop path was exercised end to end against a local VIAME build with a debugging SMTP server.

https://claude.ai/code/session_01W1C4QY6hxjHaUPJfWPfQuu

### DIFF
--- a/client/dive-common/apispec.ts
+++ b/client/dive-common/apispec.ts
@@ -454,6 +454,8 @@ interface Api {
       path?: string;
       folderId?: string;
     },
+    /** Address that receives progress and completion reports from the run. */
+    monitorEmail?: string,
   ): Promise<unknown>;
 
   /**
--- a/client/dive-common/constants.ts
+++ b/client/dive-common/constants.ts
@@ -242,6 +242,11 @@ function simplifyTrainingName(item: string) {
   return item.replace('.conf', '');
 }
 
+/** Loose check for a single mailbox address, enough to catch typos in forms. */
+function isValidEmail(address: string) {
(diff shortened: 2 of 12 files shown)


BryonLewis:
For this I would rather have some standardized way that VIAME reports back progress (either throught a file, function or standarized stdout) that could then be used by desktop and web to report back progress.  For Web we would then use the girder way for sending the email and for desktop we could present popup dialogs or other options.

PR #1909, opened 2026-09-09
Let the collapsed Edit Types group expand while nothing is being edited

mattdawkins:
- Collapsed activator is no longer disabled outside edit mode, so its expand toggle works like the Visibility one
- Outside edit mode the icon is dimmed and the dropdown does not open

https://claude.ai/code/session_01R9QNsyzjQpPbUrWRtiwmrq


BryonLewis (approved):
Noticed this but never got around to it, was something that was needed

PR #1911, opened 2026-09-10
Add a shared DatasetPicker component

mattdawkins:
Review, Query, Training, Scoring and Pipelines each choose datasets differently today: an autocomplete (Review, Scoring on desktop), a browse button (Review, Scoring on web), a searchable data table with a plus button (Training, Pipelines) and a checkbox table with select all (Query). This PR adds one component to replace all of them and wires it into the two pages already on main, Training and Pipelines. Review, Query and Scoring follow in their own PRs.

`dive-common/components/DatasetPicker.vue`

- Search field that filters by every listed column (name and type by default).
- Table of the datasets on offer with an add button per row; already selected rows are greyed with a check.
- "Select all (n)" adds everything the search currently lists that is not selected yet.
- Optional browse button (`pickerLabel` / `pick` event) for the web, where datasets come from a Girder dialog instead of a listing.
- `headers` prop for extra columns (Training and Pipelines add fps), `row-actions` slot for per-row extras (Training's view button), `compact` to hide paging on short lists.
- Pages keep their own table of what they selected; the picker only offers.

`dive-common/datasetPicker.ts` holds the row type and the filtering / select-all logic, with a spec.

Training and Pipelines: their available-dataset tables are replaced by the picker; Pipelines drops its own search and select-all code. Verified in the desktop app: search, add one, select all of a search, clear the search, staged counts and the run/train buttons update.

https://claude.ai/code/session_01W1C4QY6hxjHaUPJfWPfQuu

### DIFF
--- a/client/dive-common/components/DatasetPicker.vue
+++ b/client/dive-common/components/DatasetPicker.vue
@@ -66,10 +66,11 @@ export default defineComponent({
     },
   },
   setup(props, { emit }) {
-    const search = ref('');
+    /** Null when the field's clear button is used. */
+    const search = ref<string | null>('');
 
     const searchFields = computed(() => props.headers.map((h) => h.value));
-    const listed = computed(() => filterDatasetRows(props.items, search.value, searchFields.value));
+    const listed = computed(() => filterDatasetRows(props.items, search.value ?? '', searchFields.value));
     const addable = computed(() => selectableIds(listed.value, props.selectedIds));
     const selected = computed(() => new Set(props.selectedIds));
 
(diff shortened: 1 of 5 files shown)


BryonLewis (approved):
The previous search would search the type and the name of datasets.  That mean that searching for 'mult' would bring up all multi datasets or searching 'image' would bring up all image-sequence datasets.  I've modified it so there is a type filter for the type and search only searches the dataset name.  I think that is more how User's would expect it to work.

Also did some minor changes to handle stagedIds that change after selecting a multicamera/stereo pipeline.  Another change reduces the logic of trying to load all calibration files when not needed.

PR #1912, opened 2026-09-10
Follow VIAME stereo pipeline names and category

mattdawkins:
VIAME pipelines now use the `stereo_*` prefix, with measurement-producing pipelines named `stereo_measure_*`. Update desktop and server discovery, stereo category filtering and display, annotation input handling, and the rectified disparity dataset special case to follow these names. Remove the obsolete hidden stereo category rule.

Validation: 27 focused client tests passed; ESLint passed for all changed client files; rename/reference checks passed in VIAME. Updated server tests could not run locally because Girder and girder_worker are not installed.

### DIFF
--- a/client/dive-common/constants.ts
+++ b/client/dive-common/constants.ts
@@ -222,15 +222,13 @@ const zipFileTypes = [
   'zip',
 ];
 
-const stereoPipelineMarker = 'measurement';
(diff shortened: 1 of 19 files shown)


BryonLewis (approved):
only updated some documentation to align with the new names

PR #1913, opened 2026-09-11
Add Resume button to desktop navigation

mattdawkins:
Add Resume to the left of Library to reopen the last successfully loaded annotation sequence after navigating to Training, Review, or another menu page. The button is hidden until a sequence is available. The target is remembered for the app session, excludes scoring previews, and clears when the dataset is removed.

Validation: targeted ESLint passed; TypeScript checks passed for the navigation implementation before the label and visibility adjustment.


BryonLewis (approved):
Seems to work as intended, wondering if we should add something similar for web in the future

PR #1914, opened 2026-09-11
Fix stale annotation indices and redraw watchers after reload

mattdawkins:
Reloading pipeline annotations could leave stale detections in the frame index and redraw watchers attached to a destroyed viewer. Clear the interval tree directly when replacing annotations, and mount the tooltip after LayerManager setup so its redraw watchers are disposed with the component.

Regression tests cover clearing and repopulating the frame index, stopping redraws after the old viewer unmounts, and updating the replacement viewer once.

Validation:
- 58 tests passed across LayerManager, TrackStore, CameraStore, and desktop VIAME CSV serialization.
- Type checking passed.
- ESLint: no errors; two component-count warnings in the test file.
- `git diff --check` passed.

### DIFF
--- a/client/src/BaseAnnotationStore.ts
+++ b/client/src/BaseAnnotationStore.ts
@@ -206,9 +206,7 @@ export default abstract class BaseAnnotationStore<T extends Track | Group> {
 
   clearAll() {
     this.annotationMap.clear();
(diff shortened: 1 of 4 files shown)


BryonLewis (approved):
Small decent change to fix the statle issues.  I'm good with this.

PR #1924, opened 2026-09-12
Make sequence-wide Type Filters a saved user preference

mattdawkins:
Type Settings now exposes **Show types from entire sequence**, enabled by default. It keeps types with zero current-frame annotations in the list; disabling it limits rows to current-frame types. The choice is saved across sequences using the existing frame-filter setting.

Validation: 35 settings, settings-panel, and type-list tests passed; changed-file lint passed.


BryonLewis:
This is just renaming and forcing the switch to the opposite default for an existing feature.  We have the "Filter by Frame" option in the UI for type settings that will filter the annotations by the current frame.  It is off by default but can be toggled on and saved by the user.  It also matches our Typer counter systme where we show Total:frame counts.  This change renames it and makes it more like a negative opt-in.  I.E.  Before you "Filter by Frame" to make it show the current frame's types.  The update has you turn off "Show types from entire sequence" to simulate the same behavior.  I think this is a bit more confusing and not as direct as the existing option.

BryonLewis:
closed in favor of #1926

PR #1927, opened 2026-09-15
feat: Override default DIVE_PARAM values declared in includes

Louis-Pagnier-KW:
In the case of a DIVE_PARAM specified in an included .pipe file, if the same key is overridden in the main file, this value should be used by default instead of the one in the included file.

Example

my_included_file.pipe
```
process measurer
  block stereo_disparity:ocv_stereo_disparity
    num_disparities                            = 256  # DIVE_PARAM["Num Disparities", int]
```

my_main_pipe_file.pipe
```
include stereo_measure_current_annots_template.pipe

config measurer
  block stereo_disparity:ocv_stereo_disparity
    :num_disparities                           240
```

Here, we expect `num_disparities` to be set to 240 by default in the UI (instead of 256).


BryonLewis (approved):
I added some commits to do the following:

- Matched the DIVE Params override for Web/Girder version that as implemented in the Desktop version inside of common.ts
- Updated the comment to clarify that that bare assignments override inherited defaults
- Added some tests for the Web/Girder version
- Added some tests for the Desktop version as well

PR #1928, opened 2026-09-16
Foundation stereo point matching on web and desktop

mattdawkins:
- Web labels: "Higher Quality, Slower" (Fast Foundation Stereo) / "Lower Quality, Faster" (template matching). Default Higher Quality on both platforms.
- Desktop: same setting plus "Medium Quality, Medium Speed" (NCC + DINO); each maps to a VIAME interactive stereo config sent on `enable`; refuses to enable when the FAST-FDN-STEREO or DINO add-on is missing; changing the setting reloads the service.
- Web server reads VIAME's `cmake/download_viame_onnx.csv` and serves the export that fits the dataset's frame size (smallest covering, else largest; `DIVE_STEREO_WEB_MODEL=HxW` forces one). Client sends the frame size and caches one copy per size.
- Requires VIAME `viame.core.interactive_service` with per-enable `config` support and the DINO add-on's `interactive_stereo_ncc_dino.conf`.

https://claude.ai/code/session_013oFD6dP2EQY3R9UdXzx7dL


BryonLewis (approved):
- made an update to prevent all of the requests for each annotation and new track.  This will cache the calbiration file and use it instead of making a request for every change
- Added download progress indicator for the 100MB version of the model
- add padding to the stereo option (it was too close to the switch above it)
- added a way to invalidate the calibration cache
- will fall back on desktop to the previous quick modle if the addon isn't installed
- Fallback that works for importing calibrations

PR #1932, opened 2026-09-17
Keep a detection's box corners editable while editing its line

mattdawkins:
- In line editing mode, when the detection also has a box and boxes are visible, the box corners get drag handles too, so the auto-generated box can be adjusted without leaving line mode.
- Implemented as a second `EditAnnotationLayer` (`companion: true`, corner handles only) that is live whenever the selected track has both a saved line at the selected key and bounds; it stays off while the line is still being drawn.
- GeoJS applies a handle drag to every annotation layer in edit mode, so each layer records at `actiondown` whether its own handle was grabbed and only that one moves and commits. A box commit does not rebuild the line layer, and a hovered handle's drag action is restored after the peer layer changes mode.
- Covered by a real-GeoJS jsdom spec that drags a box corner and a line vertex with the other layer's handle previously hovered.


BryonLewis (dismissed):
just fixed some minor issue that could clear out creation actions.

PR #1935, opened 2026-09-18
Let a lone head or tail point be dragged in line mode instead of only extended

mattdawkins:
Stacked on #1932 (its commits are included here; this shrinks to the last commit once #1932 merges).

- A detection with only a head or only a tail (e.g. a computed keypoint) could only be extended into a line in head/tail mode; the existing point could not be moved. It now appears as a draggable handle while the line tool waits for the missing point, and clicking elsewhere still places that point.
- Reuses #1932's companion edit layer: when the line tool is active and the track has exactly one of head/tail and no line, the companion becomes a `Point` editor on that key instead of the box-corner editor. Dragging commits through the head/tail recipe as a single-point edit (bounds enclose the new position), so stereo auto-compute and length updates react as they do to a placed point.
- Two GeoJS interactions handled: a mousedown on the point's handle is not a vertex of the line being drawn, and leaving the handle gives the line tool back the click-to-place actions GeoJS strips on handle hover.
- Real-GeoJS spec drags a lone head with the line tool in creation mode: the point moves, no line is started, and creation actions are restored.


BryonLewis:
The workflow on this is broken pretty badly:

- If I place only a single point down and click escape to end the point creation I still have the cursor icon for point editing/creation
- If I then edit the point and drag it around it creates a box based on where I drag it even multiple times it will create a large box
- If I drag it then click elsewhere to deselect it creates a tail point automatically in the new location and gets into a state where it will actually create 3 points.

I think something else needs to be done if we want this behavior, this PR isn't satisfying it properly.

What we really want is a way to have single points, the head/tails, or the spline for multiple points.  This may require more UI with line editing than what we have currently.  I.E maybe there are buttons after the first click that allow you end or create a tail point.  Maybe during editing after selecting there is a way to click to add a point instead of automatically creating the tail point.

PR #1939, opened 2026-09-18
Navigate a detection's polygons from either stereo camera

mattdawkins:
Edit handles, the mousedown camera switch and the commit routing from #1737 are already geometry-agnostic, so dragging a polygon vertex or a head/tail endpoint on the non-selected camera already commits to that camera, like boxes. What was still gated to the selected camera was polygon navigation:

- right-click inside the editor on the other camera (`polygon-edit-right-click`) now selects that camera and switches to the clicked polygon there, instead of being ignored
- the poly layer's right-click / right-click-outside handlers defer to the edit layer on any camera whose editor holds the detection, so a right-click no longer changes the selected key without switching camera
- specs for both; the blocked-switch case (linking mode) stays ignored

Point-click segmentation keeps its own cross-camera path and is unchanged.

To try: select a detection present on both cameras, enter polygon editing, then on the other camera drag a vertex, right-click a second polygon of the same detection, and drag a head/tail point in line mode.

- Follow-up: entering polygon editing now lands on a mask the detection already has (e.g. keyed `SegmentationPolygon`) instead of the empty key, so its vertices are editable at once rather than after a right-click; the stereo-segmented copy shares the source polygon's key so this holds on both cameras. Specs added.


BryonLewis (approved):
This seems to just match the behavior of other geometry

PR #1942, opened 2026-09-18
Ask which side of the confidence threshold a delete of every listed type reaches

mattdawkins:
- When every type shown in the type list is selected and the delete button is pressed, a dialog now asks whether to delete tracks **above** the current confidence threshold (what the list shows, the previous behaviour), **below** it (the tracks the list hides), or **all** of them regardless of threshold. Partial selections and the group list keep the plain confirmation.
- `BaseFilterControls.removeTypeAnnotationsByThreshold(types, scope)`: `above` delegates to the existing visible-list removal; `below`/`all` walk every annotation and remove the type only where its confidence is under that type's threshold, or everywhere, deleting the track when no type remains.
- Unit tests for the three scopes and for the dialog gating; docs updated.

- Follow-up: the track list's delete button (sidebar and bottom bar) asks the same question when every listed track is checked — above = the listed tracks, below = tracks of the enabled types hidden by the thresholds, all = both. The dialog moved into a shared `DeleteAllScopeDialog`; spec added.


mattdawkins:
Superseded by #1944: the question belongs on the track list, not the type list.

PR #1948, opened 2026-09-21
Keep Multi-Cam tools stationary and simplify detection deletion

mattdawkins:
Switching cameras currently moves the Multi-Cam toolbar between opposite sides of the top bar. Keep it in one stationary slot directly to the left of the visibility controls, after the flexible spacer, preserving the toolbar instance across camera switches.

For a track with a single state in the selected camera, show one delete action labeled “Delete detection from current camera.” Multi-state tracks retain separate frame and whole-track deletion actions. Apply this in both expanded and collapsed modes, and update the actions when states are added.

Validation: four component regression tests pass, covering camera switching, camera-specific deletion, and adding a second state in both toolbar modes. ESLint passes for EditorMenu.vue. The positioning update moves only the existing slot past the spacer.


BryonLewis (approved):
LGTM
In the future we may revisit and change a bit more about how this works

PR #1954, opened 2026-09-21
Add model ZIP import/export and desktop Models page

mattdawkins:
DIVE can now import and export complete model packs from the Models page. Export to ZIP preserves every pipeline, weight file, and supporting folder; the existing conversion action is labeled Convert to ONNX.

ZIP import accepts VIAME add-on layouts (`configs/pipelines/` or `pipelines/`) and flat packs, optionally inside one enclosing folder. Imports preserve relative paths, use a new name on collisions, reject unsafe archive entries, and clean up failed imports.

Desktop model management moves from Training to a Models tab immediately after Training, with import, ZIP export, ONNX conversion, and deletion. Imported pipelines can use arbitrary filenames and nested model weights. The web Models description no longer forces a line break.

Validation: 166 client tests and 22 server tests passed; TypeScript, ESLint, Python formatting checks, and web and desktop production builds passed. Tests cover archive layouts, round trips, failed-import cleanup, duplicate names, discovery, and UI actions.


BryonLewis:
- fixed an issue where the deletion on web didn't show up in the list until after a refresh
- Updated the import on Web so it immediately asks for a file and while uploading shows progress (because it could take longer on a remote server compared to local testing)
- Added a quick sanity check for the export to ONNX functionality that will disable and provide a tooltip if the basic requirements aren't met.  This doesn't mean it will succeed but simple checks for weights or checkpoints are made and will disable the button to reduce user confusion.  This happens on both desktop and web.

BryonLewis (approved):
Made some usability changes

